PR #143, opened 2018-05-09
Feature/apix and batchimport fixes

jannistsiroyannis:


batchimport/integtest/test.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
mxtthias: Bonus points for adding a short entry about this to the README.

PR #152, opened 2018-05-15
Update descriptionCreator

klngwll:
Please check that the change of descriptionCreator form doesn't break anything elsewhere in backend?

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -3865,7 +3865,7 @@
 
     "040": {
       "aboutEntity": "?record",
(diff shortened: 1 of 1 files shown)


klngwll:
@jannistsiroyannis or @mxtthias can you check these changes with backend handling of these properties?

niklasl:
Given the reasoning that desciptionUpgrader is reasonably always a Library (or "something") with a sigel, but that it is recorded manually now, this is good enough.

The theoretical case that we might create library URI:s for this but still not be able to look those up during embellish, is a risk, but probably a negligible one. *If* that happens, we "should" notice problems during revert (if bib 040 d is important enough for systems) and add a uriTemplate at that point. If we do it now, the only actual effect is that existing descriptionUpgrader data will appear as linked in the cataloguing client, but with no possible means of manually linking new ones.

PR #155, opened 2018-05-16
Make the download marc button correctly use MARCXML where chosen in p…

jannistsiroyannis:
…rofile. Also, make proper use of bash subfields, and use cached embellished data.


rest/src/main/groovy/whelk/rest/api/LegacyMarcAPI.groovy
```diff
@@ -124,19 +126,21 @@ class LegacyMarcAPI extends HttpServlet {
                 profileString = defaultProfileString
             }
 
+            //System.out.println("Using profile: " + profileString)
```
mxtthias: Remove or add a comment explaining why it should be kept.

PR #160, opened 2018-05-17
Feature/index cards v2

mxtthias:


importers/src/main/groovy/whelk/reindexer/ElasticReindexer.groovy
```diff
@@ -9,7 +9,7 @@ import whelk.util.Tools
 @Log
 class ElasticReindexer {
 
-    static final int BATCH_SIZE = 500
+    static final int BATCH_SIZE = 1000
```
mxtthias: We could probably increase this even more, since we only index cards now. (b02610cbbd2c4cd1317a331551e78744fc6b300e mentions 1.5Mb per batch with 500 complete documents.)

PR #161, opened 2018-05-17
Feature/missing indicators

klngwll:
There are alot of missing indicators. I would prefer to merge and make new feature branches for chunks of those. Not entirely happy with how these turn out but I guess not loosing them is no.1 priority.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -4155,17 +4155,19 @@
       "addLink": "marc:hasGeographicClassification",
       "resourceType": "marc:GeographicClassification",
       "i1": {
-        "TODO:link": null,
+        "property": "marc:i1codeSource",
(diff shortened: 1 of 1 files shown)


niklasl:
I'd like to add them all tomorrow, if we can make a joint sweep then? Let's keep this branch until then to see if we can batch them all in.

niklasl:
I think this is good to go now...

I would prefer to make the i1:s of bib 520 and 521 into "plan B" as well though (i.e. prefix both property and tokenMap values with `marc:`). We can't really use `typeNote` for those, as these values would have to remain strings if so (as opposed to using "micro-enum" lists (dropdowns) which is our plan for "required" plan B terms).

niklasl:
I went ahead and made those final adjustments. @klngwll would you like to chime in, or should we call this done?

klngwll:
Looks good!

PR #171, opened 2018-05-21
Surface auth-references with data in the about part of responses alon…

jannistsiroyannis:
…g side

holds. The point of this is to enable OAIPMH clients to have access to raw
auth records to perform their own mangling of records when needed.


jannistsiroyannis:
Yes, for the known cases where _we_ use this we do (or will). For example this will also enable a speedup of batch export. The alternative would be to an additional _includeauth, which would be a nightmare, because we'd then have to cover every possible permutation of format postfixes (and 2 are bad enough).

PR #181, opened 2018-05-25
When using the download compiled marc button, at least rudimentary ho…

jannistsiroyannis:
…lding

info should be included, regardless of what the export profile says.


rest/src/main/groovy/whelk/rest/api/LegacyMarcAPI.groovy
```diff
@@ -55,7 +55,7 @@ class LegacyMarcAPI extends HttpServlet {
             "format=ISO2709\n" +
             "status=on\n" +
             "longname=LIBRIS testprofil\n" +
-            "extrafields=Mtm\\:852,856\n" +
+            //"extrafields=Mtm\\:852,856\n" +
```
mxtthias: I'd like a comment here saying why we keep this line, since I'll forget otherwise.. :/

PR #187, opened 2018-05-26
Strip mod_deflate ETag suffix

mxtthias:
mod_deflate appends a suffix to the ETag by default, which means that we
have to clean that up to be able to match it against the modified
timestamp.


mxtthias:
@niklasl, @jannistsiroyannis: I'm not sure if this is the best way to solve it. Another way of doing it is rewriting the incoming headers (`If-Match`/`If-None-Match`) in the Apache conf, which might make more sense from a caching perspective? However, since we perform the modified check in the code, I'm leaning towards cleaning up the incoming values there as well. What do you think?

PR #196, opened 2018-05-31
Feature/importfields

klngwll:
Add bibliographic fields 029, 385, 386 to marcframe because imports.


niklasl:
This looks good!

If we have a bit of time we/I might factor out the $x-to-ISSN mapping to an included pattern (as it looks identical across a bunch of fields).

Also, perhaps 029 could use the idbase-pattern as we've done with a bunch of other identifiers in the bib 02x/03x fields? I.e. just add a marc:OCLCNumber?

klngwll:
Needs spec for 029

niklasl:
As discussed today, these will likely become suppressed in imports anyway, so do we close without merge?

PR #203, opened 2018-05-31
Preserve type details in bib 006

niklasl:
Use new baseTypeMap to map the specific types to the defined column sets divided by base type.

*NOTE* that we may want to change from `hasPart` to `marc:additionalWork`, given the odd nature of this piece of description...


niklasl:
We'll keep `hasPart` as it is (groupId and all).

PR #205, opened 2018-06-01
Add indicators to 505

klngwll:


klngwll:
Leave this for now. LC defines these as unneccessary.

PR #218, opened 2018-08-20
Feature/changedby in oaipmh

jannistsiroyannis:
Denna behöver mergas _och_ deployas i produktion innan motsvarande ändring (feature/respect-biboperators-setting) mergas in på master i "export"

### DIFF
--- a/oaipmh/src/main/java/whelk/export/servlet/Helpers.java
+++ b/oaipmh/src/main/java/whelk/export/servlet/Helpers.java
@@ -77,7 +77,7 @@ public class Helpers
     {
         // Construct the query
         String selectSQL = "WITH bib_with_heldby AS (" +
-                " SELECT lddb.id, lddb.data, lddb.collection, lddb.modified, lddb.deleted, lddb.data#>>'{@graph,1,heldBy,@id}' AS sigel, lddb.data#>>'{@graph,1,itemOf,@id}' AS itemOf, lddb_attached_holdings.data#>>'{@graph,1,heldBy,@id}' as heldBy" +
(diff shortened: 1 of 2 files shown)


jannistsiroyannis:
The point of this is to support the previously existing biboperators-functionality.
Essentially, it is necessary for the export program (which consumes OAIPMH exclusively) to be able to filter the list of records it produces based on "who" made the (last) change. The format we're sending is the old fashioned sigel-code, which (while perhaps regrettable) is used in several places in OAIPMH already.

There is an unfortunate inherent race condition in this, which is that if a record is changed more than once inbetween export-program runs, then only the "who" of the last change will be accounted for.

Wether or not this is an acceptable tradeoff until the export-program is moved away from OAIPMH (which is coming for other reasons), will have to be a decision for the projekt lead.

niklasl:
I see. So the export program is the only client?

With format I mean what vocabulary the XML is representing. It's not RDF, and it's not in the OAIPMH namespace, right?

As long as we're going to remove the deviating inconsistencies I'm fine with tradeoffs to make the already running stuff do what has to be done. I suppose our OAIPMH solution is deviating as it is, so as long as we don't make up structures ad-hoc which others will depend on, I won't object.

jannistsiroyannis:
No, the export program is not the only client (although the only known one), but this stuff is being added inside the <about> tag, which is a free-form escape hatch in the OAIPMH protocol, ment for "Individual implementation communities" to put whatever they want.
It's true that whatever is in there should have a schema definition in order to be fully compliant, but I think we're past that at this point.
(Though for the record, I don't want this "feature" either).

niklasl:
Ah, right, the about tag is "anyXml". At least we're not signalling any semantics and it's outside of the self-describing record. I think I'm up to speed: and we're on the same page. No objections, just notes for the road ahead.

PR #220, opened 2018-08-27
Feature/lddb to graphstore

niklasl:
Add scripts and instructions for creating load files, importing into, and running the BlazeGraph SPARQL graph store.

### DIFF
--- a/GRAPHSTORE.md
+++ b/GRAPHSTORE.md
@@ -1,6 +1,54 @@
 # Using a Graph Store
 
+NOTE: this is a work-in-progress. We have not yet set up a final, continuously fed Graph Store upon XL.
+
 In principle, any Graph Store supporting the SPARQL 1.1 Graph Store HTTP Protocol will work.
+
+## Using BlazeGraph
+
+Download the latest blazegraph jar (from: https://sourceforge.net/projects/bigdata/files/bigdata/; see: https://wiki.blazegraph.com/wiki/ for details).
+
+1. Put required configuration and tools to the loading machine and/or server:
+
+    ```
+    scp -r ./librisxl-tools/blazegraph $TARGET:
+    ```
+
+2. Go to the target machine (`ssh $TARGET`).
+
+3. Setup variables for your specific environment:
+
+    ```
+    SOURCE_HOST=pgsql-SOMEDOMAIN
+    DATADIR=/tmp/lddb-loadfiles
+    XL_JSONLD_CONTEXT=https://id.kb.se/context.jsonld
+    ```
+
+4. Create load files:
+
+    ```
+    for coll in definitions auth bib ; do # hold ...
+      echo "Dumping $coll from LDDB"
+      psql -h $SOURCE_HOST -Uwhelk -tc "SELECT data FROM lddb WHERE collection = '$coll' AND deleted = false;" |
+      python ./lddb-to-import.py $DATADIR/$coll
+    done
+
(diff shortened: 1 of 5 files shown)


GRAPHSTORE.md
```diff
@@ -1,7 +1,66 @@
...
+
+    ```
+    curl https://id-dev.kb.se/context.jsonld -o $DATADIR/context.jsonld
+    curl https://raw.githubusercontent.com/libris/definitions/develop/source/vocab-overlay.jsonld -o $DATADIR/base-context.jsonld
+    # TODO:
+    # Edit to remove @type: xsd:dateTime (fails at times in blazegraph).
+    # Also, decide whether you want context or just base-context below.
+
+    psql -h $SOURCE_HOST -Uwhelk -tc "SELECT data FROM lddb WHERE collection = 'definitions' AND deleted = false;" |
+    python ./lddb-to-import.py $DATADIR/parts/definitions 1000 ../context.jsonld
```
mxtthias: Consider indenting the python call to make it more obvious that it's part of a pipe. Same goes for the for loop below.

librisxl-tools/blazegraph/lddb-to-import.py
```diff
@@ -0,0 +1,78 @@
...
+        return
+
+    # add record id to top graph to name it
+    line = b'{{"@id": "{0}", {1}'.format(rec_id, line[1:])
+
+    # add context reference
+    line = b'{{"@context": "{0}", {1}'.format(context_path, line[1:])
+
+    # Remove Unicode control characters (mostly harmful in terms and ids)
+    line = re.sub(r'[\x00-\x1F\x7F]', b'', line)
```
mxtthias: Is this something that we should try to fix at the source?

librisxl-tools/scripts/crunch-lddb-lines.py
```diff
@@ -0,0 +1,120 @@
+"""
+This began as just an example script for running against a stream of JSON lines
+(specifically a PGSQL data dump with double escapes.) Then the Selector utility grew out ouf it...
+
+A*void developing this further*. Mostly, you will get by by using something
```
mxtthias: s/A*void/*Avoid/

librisxl-tools/scripts/crunch-lddb-lines.py
```diff
@@ -0,0 +1,120 @@
+"""
```
mxtthias: I'm divided when it comes to adding this script. On the one hand, it might be good as a reference at some point. On the other, if this isn't how we want to do things in the future, it might be better to skip this for now and move on to developing the One True Way(tm).

What are your thoughts?

niklasl: I was equally divided. I think removing it is good since the good parts are easy to redo, the other parts shouldn't be redone. But I'd like to make a note about "lddb-processing" out of a bunch of the "one-off" scripts in here. We could remove this and assemble a separate PR for that note.

librisxl-tools/blazegraph/lddb-to-import.py
niklasl: We *should*...

niklasl: https://jira.kb.se/browse/LXL-1790

mxtthias: :+1:

PR #225, opened 2018-09-06
Feature/fix condition error

sarwik:


niklasl:
This looks great! I just noticed one more error which permeates all these mappings though...

We've (I've) used `http://id.kb.se/` and not `https://id.kb.se/`! :open_mouth:

Can you adjust that too? We're gonna need to globally change these anyhow...

sarwik:
I'll fix! :)

PR #226, opened 2018-09-10
Feature/fix faulty mapping

sarwik:


sarwik:
@niklasl: There is one TODO left (on line 1524) which concerns revert. It looks like something 
you have started (with isActualValue). Do you remember what your thoughts were?

PR #229, opened 2018-09-20
Feature/music medium

klngwll:
Update to make musicMedium an objectProperty instead of the faulty mapping as a datatype.

DO NOT MERGE UNTIL VERIFYING WE CAN DO A GLOBAL CHANGE AND POSSIBLY A REINDEXING.


whelktool/scripts/2018/11/fix-musicMedium.groovy
```diff
@@ -0,0 +1,61 @@
...
+    def remodelledTermList = []
+
+    //Create musicMedium object and add to list
+    term[MUSICMEDIUM_PROPERTY].each {
+        def map = [:]
+        map << ['@type':MUSICMEDIUM_TYPE]
+        map << ['label':it]
+        remodelledTermList << map
+    }
+    term[MUSICMEDIUM_PROPERTY] = remodelledTermList
```
niklasl: This could be shortened to:
```groovy
    term[MUSICMEDIUM_PROPERTY] = term[MUSICMEDIUM_PROPERTY].collect {
        ['@type': MUSICMEDIUM_TYPE, 'label': it]
    }
```

whelktool/scripts/2018/11/fix-musicMedium.groovy
```diff
@@ -0,0 +1,61 @@
...
+        obj.each {
+            findAndFixValuesInData(data, it)
+        }
+    }
+    else if (obj instanceof Map) {
+        obj.each { key, value ->
+            checkValueInData(data, obj, key, value)
+        }
+    }
+    else {
```
niklasl: Redundant block?

whelktool/scripts/2018/11/fix-musicMedium.groovy
sarwik: Great. I update the code with your suggestion.

whelktool/scripts/2018/11/fix-musicMedium.groovy
niklasl: I did, and (as we agreed) removing it is good (that or leaving a comment about "skipping any other case").

sarwik:
I have updated the code with what we agreed on and tested it locally and it works fine. 
However, I have a question about generationProcess. Although we do not run the script as loud, do we not still want to update this property?
Or will it cause problems when running the export?

niklasl:
When we do dry-runs, doModified isn't called, and it is that which sets generationProcess (and generationDate). You could try running a non-dry (step-wise) on your local DB to see that it's updated.

PR #232, opened 2018-10-03
When the batchimport is set to replace a record, any existing SystemN…

jannistsiroyannis:
…umbers

(035$A) is kept, alongside various other IDs.


kaipoykio:
Looks good!

PR #234, opened 2018-10-04
Change mapping of termComponentList

sarwik:
Make mapping of $v, $x, $y, $z in termComponentList to 
subclasses of Subdivision. Only mapping of $a shall be treated
as a base class, such as Topic, Work, GenreForm, Person, Geographic, 
Temoral etc.

termComponentList should consist of one main term ($a) and at least
one subdivision term ($v, $x, $y or/and $z).

This improves the model and makes the export to work
as expected, for new records. To make existing data (data already
imported with previous mapping) able to export properly, we
need to make global changes to fit the new mapping.


sarwik:
This feature may not be merged until we have made the global changes to existing data.

PR #237, opened 2018-10-10
Add a reindexFrom (time) importers-command, and handle the fallout.

jannistsiroyannis:


importers/src/main/groovy/whelk/reindexer/ElasticReindexer.groovy
```diff
@@ -33,7 +33,36 @@ class ElasticReindexer {
...
+    /**
+     * Reindex all changes since a point in time
+     */
+    void reindexFrom(long fromUnixTime) {
+        try {
+            int counter = 0
+            for (collection in whelk.storage.loadCollections()) {
+                boolean includeDeleted = true
+                List<Document> documents = []
+                for (document in whelk.storage.loadAll(collection, includeDeleted, new Date(fromUnixTime))) {
```
niklasl: Can the body of this for loop be replaced by a parameterized call to `reindex(collection, includeDeleted, since)`?

jannistsiroyannis: It could, but that would result in combinatorial ugliness (and potential errors) with the parameters. The two look almost identical, but imo they're actually not since a reindex with deleted makes no sense at all, nor does a reindexFrom without it.

niklasl: Or that `includeDeleted = since != null`?

jannistsiroyannis: I feel that a little duplication of code is sometimes a lesser evil (less cryptic) than this type of logic (I did consider what you're suggesting while writing this too). If you feel strongly enough about this, then go ahead and change it, otherwise I'll leave it.

PR #246, opened 2018-10-24
Optimize whelktool scripts and make them independent

niklasl:
The scripts fix-unbalanced-brackets and remodel-termComponentList now use selectBySqlWhere. The mainjob script (for this "month") is removed. This optimizes the selection for remodel-termComponentList and sets up the two jobs to run independently.

I believe this is what we want to do on STG and in PROD. If approved before Friday, we can rebase the release to pin the jobs to the release, although we currently run the jobs from the develop branch (pending a possibly stricter process).


whelktool/scripts/2018/10/remodel-termComponentList.groovy
```diff
@@ -19,7 +19,13 @@ boolean changeToSubdivision(it, subdivisionTypes) {
     return changed
 }
 
-process { data ->
+selectBySqlWhere('''
+        data::text LIKE '%"termComponentList"%'
+        ''') { data ->
+    // guard against missing entity
+    if (data.graph.size() < 2) {
```
mxtthias: Just for my own understanding, when can this happen?

niklasl: AFAIK, only when the LIKE select finds the vocab itself (which describes "termComponentList"), which is its own mainEntity (sort of). I opted for a guard over tuning the select to be more robust.

PR #248, opened 2018-10-30
Export all affected bib records when a holding record is relinked mul…

jannistsiroyannis:
…tiple times.

When a holding record is relinked multiple times within an export-interval
we would previously only export the newly linked bib and the previously linked
bib. That is not enough, since a holding may have been relinked (to different
bibs) more than once during an interval.

The new approach is to export all bibs to which the holding _has_ been linked
within the update interval (including the bib it was linked to before the inteval).

As an implementation detail the initial scan for changes is now done on the
versions table instead of lddb, which means that exporting from time A to B
will always produce the same result, regardless of whether or not the records
that changed within that interval are changed again after the interval.

### DIFF
--- a/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
+++ b/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
@@ -138,6 +138,12 @@ public class ProfileExport
         {
             List<Document> versions = m_whelk.getStorage().loadAllVersions(id);
 
+            // The 'versions' list is sorted, with the most recent version first.
+            // 1. We iterate through the list, skipping (continue) until we've found a
+            // version inside the update interval.
+            // 2. We keep iterating over versions and check if we're still inside the
+            // interval _after_ each iteration, which means we will pass (export) all
+            // versions inside the interval and exactly one version "before" the interval.
             for (Document version : versions)
             {
                 String itemOf = version.getHoldingFor();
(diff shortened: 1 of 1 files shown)


marc_export/src/main/java/whelk/export/marc/ProfileExport.java
```diff
@@ -137,38 +138,26 @@ else if (collection.equals("hold") && updateShouldBeExported(id, collection, pro
...
+
+                String itemOfSystemId = m_whelk.getStorage().getSystemIdByIri(itemOf);
+                if (itemOfSystemId != null) // itemOfSystemId _can_ be null, if the bib linked record is deleted (no thing-uri left in the id table)
+                    exportDocument(
+                            m_whelk.getStorage().loadEmbellished(itemOfSystemId, m_whelk.getJsonld())
+                            , profile, output, exportedIDs, deleteMode, doVirtualDeletions);
+
+                boolean insideInterval = from.toInstant().isBefore(modified) && until.toInstant().isAfter(modified);
+                if ( !insideInterval )
+                    break; // Only one more once outside the interval, then stop.
```
mxtthias: These three lines and the `versions` loop are kinda difficult to grasp, unless you know the ordering of the versions. Can you clarify the comment and explain why the `insideInterval` makes sense here instead of e.g. at the top of the loop?

PR #253, opened 2018-11-12
Cleanup deprecated and unused code

niklasl:
I had some cleanup hanging around. We might want to keep this alive for a while, in case we stumble upon more corners we want to dust off.

(Currently, the one commit cropped up during a conversation with GUI, where it was unclear whether the resource was till used. It was not; but we want to detect such deviating use early on, hence the removal which then never landed.)


niklasl:
@oBlissing So we can remove this safely? Otherwise we can wait until the corresponding GUI code is cleaned up.

PR #260, opened 2018-11-19
Feature/merge publication on fuzzy dates

niklasl:
This compares date (as well as startYear plus optional endYear) with year for fuzzy equivalence when merging a PrimaryPublication. We do so by removing defined "fuzz" from the fuzzy value, and then remove the stricter dates as well. If the result is empty, the fuzzy date is considered equivalent enough.

Feature includes:
* Updated marcframe logic
* A whelktool script (to change approx. 2.1 M descriptions)

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
@@ -249,8 +249,7 @@ class RestructPropertyValuesAndFlagStep extends MarcFramePostProcStepBase {
                 if ((!flagRule.revertRequiresNo ||
                         !target.containsKey(flagRule.revertRequiresNo)) &&
                     (targetContainsSomeRemapped ||
-                     (!fuzzyMergeProperty || onlyExpectedValuesInFuzzyProperty)
-                    )
+                     onlyExpectedValuesInFuzzyProperty)
                    ) {
                     mergeOk = ld.softMerge(obj, target)
                 }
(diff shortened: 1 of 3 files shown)


whelktool/scripts/2018/11/cleanup-publication-duplicates.groovy
```diff
@@ -0,0 +1,67 @@
...
+''') { data ->
+    def (record, instance, work) = data.graph
+
+    if (!isInstanceOf(instance, 'Instance')) {
+        return
+    }
+
+    List publications = instance.publication
+
+    if (!(publications instanceof List) || publications.size() < 2) {
```
mxtthias: Doesn't this check need to be after the `removeAll` call below?

whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
```diff
@@ -249,6 +263,37 @@ class RestructPropertyValuesAndFlagStep extends MarcFramePostProcStepBase {
         }
     }
 
+    /**
+     * This compares a fuzzy value with stricter values for fuzzy equivalence
+     * when merging a node. We do so by removing defined "fuzz" from the fuzzy
+     * value, and then remove the stricter values as well. If the result is
+     * empty, the fuzzy value is considered equivalent enough.
+     */
+    boolean checkOnlyExpectedValuesInFuzzyProperty(Map target, Map obj,
```
mxtthias: This code is not very easy to follow. In particular, how and why the pattern is created and how we actually extract the value we want to check.

I'm (almost) always a fan of splitting things into smaller functions and giving them explanatory names. Would that be possible here?

whelktool/scripts/2018/11/cleanup-publication-duplicates.groovy
niklasl: Ah, no, it's just a guard to skip processing instances where there aren't multiple publications.

mxtthias: Discussed offline, no change necessary.

PR #261, opened 2018-11-20
Feature/update fix music medium

sarwik:


sarwik:
I have tested on QA data and the script only skips newly added records.

PR #267, opened 2018-12-07
Add the scheduleRevertTo call in whelk tool.

jannistsiroyannis:
The point of this is to restore a record to the state it had at a certain point in time.
The restoration itself can be both loud and silent.

### DIFF
--- a/whelktool/src/main/groovy/datatool/WhelkTool.groovy
+++ b/whelktool/src/main/groovy/datatool/WhelkTool.groovy
@@ -1,5 +1,6 @@
 package whelk.datatool
 
+import java.time.ZonedDateTime
 import java.util.concurrent.LinkedBlockingDeque
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


whelktool/src/main/groovy/datatool/WhelkTool.groovy
```diff
@@ -355,6 +360,35 @@ class WhelkTool {
...
+    private boolean doRevertToTime(DocumentItem item) {
+
+        // The 'versions' list is sorted, with the most recent version first.
+        List<Document> versions = whelk.storage.loadAllVersions(item.doc.shortId)
+
+        ZonedDateTime restoreTime = ZonedDateTime.parse(item.restoreToTime)
+        Document selectedVersion = versions.get(versions.size()-1)
+
+        // Go over the versions, oldest first (in reverse),
+        // until you've found the oldest version that is younger than the desired time target.
```
mxtthias: Should we silently ignore cases where `restoreTime` is older than any version timestamp or do we need to inform the user?

jannistsiroyannis: I've pushed a new version, logging such occurrences.

PR #269, opened 2018-12-17
Use Whelk.remove for deletions

mxtthias:
This way, we'll do deletions in ES as well (if it's configured).


mxtthias:
I'll let you know once I've tested it locally.

PR #270, opened 2018-12-20
Add script to delete holdings for Frescati

mxtthias:
(+ a file with specified IDs.)


mxtthias:
General question: How do we want to handle ID:s supplied by customers? Trust them implicitly? Check if the posts have the expected characteristics (in this case holdings for `H`)?

PR #271, opened 2018-12-21
Feature/upgrade java tooling

niklasl:
This contains preparatory steps to mitigate the upgrade to Java 11.

One remaining thing is to bump up Gradle to 5.0, which requires some (presumably beneficial) changes to our inclusion of whelk-core in the various applications within this repo.


olovy:
Had to upgrade jetty version used by gretty.
Had to add a `gradle.properties` specifying version to each subproject since `gradle.properties` in root dir doesn't get picked up when executing gradle wrapper from subdirectory(?)

PR #272, opened 2019-01-08
Feature/enable sort on main title

mxtthias:


mxtthias:
Tested with both ways of creating the ES index.

PR #273, opened 2019-01-10
Remove mapping of 352q, add TODO

klngwll:
digitalCharacteristic/EncodingFormat/label is now mapped only to 347#b in MARC21. Contains TODO about improved future handling.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -8174,11 +8174,8 @@
         "about": "_:cartographicobj",
         "property": "count"
       },
-      "$q": {
-        "addLink": "digitalCharacteristic",
-        "resourceType": "EncodingFormat",
-        "property": "label"
-      }
+      "$q": null,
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -8174,11 +8174,8 @@
         "about": "_:cartographicobj",
         "property": "count"
       },
-      "$q": {
-        "addLink": "digitalCharacteristic",
-        "resourceType": "EncodingFormat",
-        "property": "label"
-      }
+      "$q": null,
+      "NOTE:$q": {"NOTE:record-count": 0, "TODO": "To enable future imports to EncodingFormat, we need something like onRevertPrefer 347$b to avoid collisions in export."}
```
niklasl: Might want to leave the addLink, resourceType and property in here if we want to just uncomment when we have an onRevertPrefer for subfields?

klngwll: It is quite uncomplicated, and I think it might be better if it is not there so we actually check the latest update of the mapping from LC instead of assuming it is still the same.

niklasl: Fair enough; and the reference to 347$b is enough to know what terms are intended. Good stuff!

PR #276, opened 2019-01-15
Create nested sort for hasTitle.mainTitle

mxtthias:


mxtthias:
If possible, I would prefer the shorthand properties, since that would probably make the ES code simpler. :)

PR #278, opened 2019-01-22
Feature/map bib 33x 3 to type note

niklasl:


sarwik:
Good stuff! Looks good 👍

PR #279, opened 2019-01-28
Remove deprecated `setup-dev-whelk.sh`

mxtthias:


niklasl:
I'm in favour for simplification. I just want to point out that the xl_local target could do with a "use my local repos" instead of the isolated `repositories` next to the fabfile. Also, while mostly academical, this makes librisxl inacessible for the public, since the devops repo is (and will most likely continue to be) private.

niklasl:
The "for public use" part is certainly out of scope, and (e.g. by scrutinizing `manage-welk-storage.sh` and main parts of  the code).

Given that `xl_local` is mentioned in the README change,, and that developers are not supposed to develop *within* the devops `repositories` directory, I'd say it's within scope (at least to note in the README how to reconfigure this locally (which AFAIK isn't currently easy)).

PR #287, opened 2019-01-31
Clean up ES query creation

mxtthias:
All code for creating the query is now in one place.


rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
```diff
@@ -0,0 +1,460 @@
...
+     * Public for test only - don't call outside this class!
+     *
+     */
+    @CompileStatic(TypeCheckingMode.SKIP)
+    public List getFilters(Map<String, String[]> queryParameters) {
+        List filters = []
+
+        // TODO This is copied from the old code and should be rewritten.
+        // We don't have that many nested mappings, so this is way to general.
+        Map groups = queryParameters.groupBy { p -> getPrefixIfExists(p.key) }
```
niklasl: For simple lambdas, prefer either a name or `it` over a single letter. Both `it` and a single letter requires the reader to understand the thing being operated upon (which IMHO is good for comprehension), and (IMHO) `it` makes that (fairly) obvious.

mxtthias: I agree! As discussed afk, we'll keep it as is for now.

PR #296, opened 2019-02-11
Feature/characteristic source

klngwll:
Fixing source which ends up on Instance instead of digitalCharacteristic, videoCharacteristic, projectionCharacteristic or soundCharacteristic.

Open for discussion.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -6816,10 +6816,35 @@
     },
 
     "256": {
+      "NOTE:LC": "Prior to the definition of this field in 1987, file characteristics were recorded in subfield $a of field 300",
       "aboutEntity": "?thing",
-      "$a": {"link": "digitalCharacteristic", "resourceType": "FileType", "property": "label"}
+      "pendingResources": {
+        "_:digitalCharacteristic": {"addLink": "digitalCharacteristic", "resourceType": "DigitalCharacteristic"}
+      },
+      "$a": {"about": "_:digitalCharacteristic", "link": "hasNote", "resourceType": "Note", "property": "label"},
+      "$6": null,
+      "$8": null,
+      "_spec": [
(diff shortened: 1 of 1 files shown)


klngwll:
256 has been slightly remodeled. Some fields are no longer valid. This is only halfway to a real solution but still much better than the current situation. It could be a good thing to merge it and analyze it in a test environment while we still have some time to reverse or reshape until next release.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -6816,10 +6816,37 @@
     },
 
     "256": {
+      "NOTE:LC": "Prior to the definition of this field in 1987, file characteristics were recorded in subfield $a of field 300",
       "aboutEntity": "?thing",
-      "$a": {"link": "digitalCharacteristic", "resourceType": "FileType", "property": "label"}
+      "addLink": "digitalCharacteristic",
+      "resourceType": "DigitalCharacteristic",
+      "repeatable": false,
+      "$a": {"addLink": "hasNote", "resourceType": "Note", "property": "label", "NOTE:repeatable": false},
```
niklasl: I'd really prefer for this to either A) be reduced to just a `comment` property, or B) for this "sketchy" `digitalCharacteristic` to be changed to just a `hasNote` with type `marc:FileTypeNote`.

(Although the majority of data values seem to be a textual label for the `FileType` (somewhat in favour of the original mapping) I presume bib 256 is considered as legacy data which we might even drop the majority of?)

niklasl:
I like this compromise. A total cleanup/normalization isn't feasible, and this is a solid step towards a more managable structure. Let's just sort out some minutia and then merge.

whelk-core/src/main/resources/ext/marcframe.json
klngwll: It has now been updated per discussion.

PR #297, opened 2019-02-12
Set fixedDefault of Serial 008[18] and 006[1] and update matchUriToken:

sarwik:
* token 'u' will be used as default
* remove token '_' as a potential uri in favor of 'u'.


sarwik:
The behavior has been tested locally together with https://github.com/libris/definitions/pull/84 - It works as expected.

niklasl:
Note that the 'u' will be represented as the *absence* of a `frequency` , so we don't really need to match for it. (We do include the `fixedDefault` value in other plaes using `matchUriToken` though, so keeping it in the list of tokens should not cause practical problems.)

sarwik:
I have added spec for the fixedDefault case. 

NB. I aslo needed to update the mechanics of revert of specs to be able to allow all specs to use a full record source (in the same way as convert)

PR #304, opened 2019-02-19
Feature/set default of serial frequency

sarwik:


sarwik:
Tested locally together with https://github.com/libris/definitions/pull/90 and it works as expected: marc:SerialsRegularityType is no longer subClass of Frequency and therefore not possible to add to frequency (for serials). Default value of export is set to '|'.

PR #306, opened 2019-02-25
Hide keyword usage in external API

mxtthias:


rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
```diff
@@ -466,4 +466,42 @@ class ESQuery {
         }
         return result
     }
+
+    @CompileStatic(TypeCheckingMode.SKIP)
```
niklasl: Nitpick: Javadoc should appear before annotations.

mxtthias: D:

I fix!

PR #308, opened 2019-02-27
Feature/has variant topics

klngwll:
Upgrade auth 448 and 480 to hasVariant. Remove mapping of 485.

Update in descriptions will be done manually.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -17044,12 +17044,12 @@
           "absorbSingle": true
         }
       },
-      "subfieldOrder": "w i a x z y v",
+      "subfieldOrder": "w a v x y z",
       "$a": {"about": "_:topic", "property": "prefLabel"},
       "$w": {"about": "_:topic", "property": "marc:controlSubfield"},
       "_spec": [
         {
-          "name": "Repeated Temporal",
+          "name": "Repeated Temporal terms in hasVariant",
           "source": [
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17032,17 +17032,42 @@
...
+      "pendingResources": {
+        "_:topic": {
+          "addLink": "termComponentList",
+          "resourceType": "Temporal",
+          "required": true,
+          "itemPos": "first",
+          "absorbSingle": true
+        }
+      },
+      "subfieldOrder": "w i a x z y v",
```
sarwik: We haven't mapped 'i'. Should we? Otherwise remove it from subfieldOrder...

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17304,31 +17329,51 @@
...
+      "resourceType": "TopicSubdivision",
+      "pendingResources": {
+        "_:topicSubdivision": {
+          "addLink": "hasVariant",
+          "resourceType": "TopicSubdivision",
+          "required": true,
+          "itemPos": "first",
+          "absorbSingle": true
+        }
+      },
```
sarwik: From what I can see, pendingResource isn't needed in this case (since it's not a nested entity).

klngwll: Fixed it! :)

sarwik: \o/

PR #312, opened 2019-03-05
Add temporal terms from scheme barn, which also shall be deleted

sarwik:


whelktool/scripts/cleanups/2019/02/fix-complexsubject/delete-specific-temporal-subjects-and-update-refs.groovy
```diff
@@ -3,10 +3,14 @@ SAO_URI = 'https://id.kb.se/term/sao'
...
 TEMPSUB_TYPE = 'TemporalSubdivision'
 SUBJECTS_TO_DELETE = ['pm14bcq70h0qnnr': 'medeltiden',
+                      'dbqt2sfx579q4h9': 'medeltiden',
                       '53hlst5p58lc91k': 'forntiden',
+                      'hftx5wm1424mg1f': 'forntiden',
                       '64jmtv6q2v2t5b9': 'antiken',
+                      'fcrv3tgz3dr0z7r': 'renässansen',
                       '75knvw8r0qh4c88': 'renässansen']
 
+TERMS_TO_UPDATE = ['medeltiden', 'forntiden', 'antiken', 'renässansen']
```
niklasl: You could use `SUBJECTS_TO_DELETE.values() as Set` here.

sarwik: Updated :)

niklasl: Any particular reason for using unique() instead of Set? :) (It doesn't matter here, but sets are unique by definition, and lookups are quicker.)

sarwik: No, I just didn't know :) Then of course, Set is better. I have updated and tested and it works fine :)

PR #314, opened 2019-03-05
Change mapping of bib 653 ind2=5 to Geographic

sarwik:
This fixes the incorrect mapping of bib 653 ind2=5. By changing it to Geographic, it will align with bib 651 and Place will only be used together with place.

This also means that we need to update existing data globally in order to conform to the new mapping.


niklasl:
Looks good! Will you add the global change in the same feature-branch? (It might be good for book-keeping.)

klngwll:
Geographic is both a subClass and and a equivalentClass in vocab, maybe remove the equivalence?

PR #317, opened 2019-03-07
Use language container alias map and subproperties on revert

niklasl:
Control language selection using a 'locale' setting, and subproperty
usage using an 'infer' flag.


niklasl:
Added specs and a slight improvement.

PR #320, opened 2019-03-15
Fix revert bug of subject, type Family with no inScheme. Should revert to ind2=4

sarwik:


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -10333,7 +10334,33 @@
             }
           }}
         },
-
+        {
+          "name": "Set in2=4 as default value",
```
niklasl: s/in2/ind2/ ;)

sarwik: Uuups... :P

PR #325, opened 2019-03-29
Feature/fix cards for es

niklasl:


whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -622,41 +622,47 @@ class JsonLd {
...
-        Map lensGroups = displayData.get("lensGroups")
-        Map cardLensGroup = lensGroups.get("cards")
         Map result = [:]
 
-        Map card = removeProperties(thing, cardLensGroup)
-        if (chipsify) {
-            card.each {key, value ->
-                result[key] = toChip(value)
+        Map card = removeProperties(thing, 'cards')
+        // If result is too small, see if a chip would be bigger, then use
```
mxtthias: Comment is a bit out of sync with the code. Something like "If result is too small, use chip instead" would probably be nicer.

PR #330, opened 2019-04-04
Use value of mapping during its removal from page params

niklasl:
This fixes the "up" link in the facet/aggregate stats, which previously
removed all parameter-value pairs, not only the one matching the current
mapping value.


oBlissing:
🔥

PR #348, opened 2019-05-22
Added script for removing BOKR records without holdings

lrosenstrom:


whelktool/scripts/cleanups/2019/05/from-list-remove-BOKR-bibs-without-hold/from-list-remove-BOKR-bibs-without-hold.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+String BIB_ID_FILE = 'lxl-2419-ids.txt'
+
+PrintWriter notDeletedIds = getReportWriter("not-deleted-bibIds")
+PrintWriter failedBibIDs = getReportWriter("failed-to-delete-bibIDs")
+PrintWriter scheduledForDeletion = getReportWriter("scheduled-for-deletion")
+File bibIDsFile = new File(scriptDir, BIB_ID_FILE)
+
+bibIDsFile.eachLine { line ->
+    selectByIds([line]) { bib ->
```
niklasl: Any particular reason (e.g. sheer size) for not reading all id:s from the `bibIDsFile` and sending that to `selectByIds`?

lrosenstrom: No, that's a valid point. I was receiving an index out of bounds exception on the select from the beginning and this was a way to pinpoint the error relate it to a particular id. I can remove the eachLine and use only the select instead for readability and consistency.

PR #351, opened 2019-05-27
Feature/improve search

niklasl:


librisxl-tools/elasticsearch/libris_config.json
```diff
@@ -91,6 +130,22 @@
...
+                            "fields": {
+                                "keyword": {
+                                    "index": true,
+                                    "type": "keyword"
+                                }
+                            },
+                            "analyzer": "numeric_extractor",
+                            "index": true,
+                            "type": "text",
+                            "fielddata": true
```
mxtthias: Probably redundant, since we have a `keyword` field too.

rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
```diff
@@ -79,20 +89,62 @@ class ESQuery {
...
 
-        Map queryString = [
+        Map simpleQuery = [
             'simple_query_string': [
                 'query': q,
                 'default_operator':  'AND'
             ]
         ]
 
+        Map queryParam = simpleQuery
```
mxtthias: `queryParam` is easy to confuse with `queryParameters` and so on. Maybe `queryClauses` is a better name?

rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
```diff
@@ -125,55 +177,65 @@ class ESQuery {
...
+    List<String> getBoostFields(String[] types, String boostMode) {
+        if (boostMode?.indexOf('^') > -1) {
+            return boostMode.tokenize(',')
+        }
+
+        String typeKey = types != null ? types.toUnique().sort().join(',') : ''
+        typeKey += boostMode
+        List<String> boostFields = boostFieldsByType[typeKey]
+        if (boostFields == null) {
+            boostFields = boostFieldsByType[typeKey] =
```
mxtthias: I feel this would be more readable with an `if` rather than using the ternary operator.

whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -670,6 +704,40 @@ class JsonLd {
...
+            List propertiesToKeep = (List) lens.get("showProperties")
+            for (prop in propertiesToKeep) {
+                def values = object[prop]
+                if (isLangContainer(context[prop]) && values instanceof Map) {
+                    values = locales.findResult { values[it] }
+                }
+                if (!(values instanceof List)) {
+                    values = values ? [values] : []
+                }
+                // TODO: find recursively (up to a point)? For what? Only
```
mxtthias: Should this be fixed later? If so, consider creating a ticket and referencing it here so it isn't forgotten.

rest/src/main/groovy/whelk/rest/api/ESQueryLensBoost.groovy
```diff
@@ -0,0 +1,121 @@
...
+        boostFields += selectedChipsLenses.sum { lens ->
+            int boost = 200
+            collectBoostFields(lens, boost, seenKeys)
+        }
+
+        // Compute boost for "card" properties.
+        def selectedCardsLenses = selectLenses(
+                cardsLenses, types, baseTypes, seenKeys)
+
+        boostFields += selectedCardsLenses.sum { lens ->
```
mxtthias: This closure is a bit gnarly. Consider splitting it up/extracting it like with `collectBoostFields`.

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -224,13 +225,14 @@ class ElasticSearch {
                     log.warn("Could not get external doc ${id} for ${document.getShortId()}, skipping...")
                 }
             }
-            whelk.jsonld.embellish(document.data, referencedData, true)
+            boolean filterOutNonChipTerms = true // Consider using false here, since cards-in-cards work now.
```
mxtthias: Might be good to explain why cards-in-cards is a good/bad thing here.

rest/src/main/groovy/whelk/rest/api/ESQueryLensBoost.groovy
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    List<String> computeBoostFieldsFromLenses(String[] types) {
+        def boostFields = []
+        def seenKeys = [] as Set
+
+        def chipsLenses = jsonld.displayData.lensGroups?.chips
+        def cardsLenses = jsonld.displayData.lensGroups?.cards
+
+        boostFields += [ "$JsonLd.SEARCH_KEY^100" as String ]
```
mxtthias: Maybe extract all boost values to constants? They are only used once, but I prefer `CHIP_BOOST_VALUE` to `200`.

rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
niklasl: Much better; will change!

rest/src/main/groovy/whelk/rest/api/ESQuery.groovy
niklasl: Fair point; changing!

rest/src/main/groovy/whelk/rest/api/ESQueryLensBoost.groovy
niklasl: Certainly!

rest/src/main/groovy/whelk/rest/api/ESQueryLensBoost.groovy
niklasl: Turning boosts into constants!

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
sarwik: Agree. And should it be changed to false now?

niklasl: Not in this change, as we need to verify that it doesn't alter any behaviour unexpectedly more thoroughly.

PR #359, opened 2019-07-01
Add test and documentation to leader

klngwll:
* Document fixedDefault with notes explaining what it is.
* Add test and future nice-to-have todo for RecordStatus.
* More explicit stated positions in leader (bib, auth, hold)

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -2101,6 +2101,7 @@
     ],
 
     "000": {
+      "NOTE:[0:5]": "Record length (only applicable in ISO 2709 serialization)",
       "[5]": {
         "aboutEntity": "?record",
         "property": "recordStatus",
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -16065,22 +16065,37 @@
...
+        "TODO": "Filter all but c & n for auth, fallbackdefault c"
+      },
       "[6]": {
+        "NOTE": "Posttyp - z - Auktoritetspost",
         "fixedDefault": "z"
       },
       "[7]": null,
       "[8]": null,
+      "[9]": {
+        "NOTE": "Teckenuppsättning - a - UCS/Unicode",
```
niklasl: Better to put this note in the inherited column (as it applies there too).

niklasl: Or not inherit bib:000 at all...

klngwll: Removed inheritance and added more explicit notes with reference to bib

PR #360, opened 2019-07-04
Feature/bfconversion2xx

klngwll:
Housekeeping update 2xx fields with some changes (except for titles) from MARC2BF2 conversion 1.5.

* Add bib 251 - Version Information, as ignored.
* Update bib 254 to hasNote correctly map field with link property. (These fields are completely dropped from conversion until this fix, ~800 records before XL launch.)
* Update bib 255 with a note on BF2 difference and mapping of $6 field.
* Update field 256 with mapping of $6.
* Change bib 257 to production place and revert to 264, dropping incompatible subfields (36 records need to be manually updated).
* Update bib 258 - Philatelic Issue Date, as ignored with expanded note (0 records in db).
* Some indicators in spec for 260 had underscore, these are changed to space.
* Add bib 261 and 262 with mapping to provisionActivities and revert to 264 (3 records unhandled).
* Add bib 265 -  acquisitionSource, with revert to 037 (1 record unhandled).
* Add expanded note for 270.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -7414,7 +7414,8 @@
     },
     "258": {
       "ignored": true,
-      "NOTE:LC": "258 - PHILATELIC ISSUE DATE (R) nac"
+      "NOTE:LC": "258 - PHILATELIC ISSUE DATE (R) nac",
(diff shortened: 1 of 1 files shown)


klngwll:
Records for 261, 262, 265 are now manually corrected.

PR #361, opened 2019-07-05
Implement reverse parameter in find API

olovy:
- Followed the existing structure in SearchUtils.groovy when adding the new parameter/mechanism


olovy:
TODO: shrink huge vocab, display-data and context files used in unit tests.

niklasl:
@olovy This looks good! I'm afraid I don't have enough time to verify everything before vacation. Let's do a final call when we're both back.

It'd be great if you're able to reduce the included vocab and display data to what's used in the tests.

Furthermore, I'd really like to cleanup and remove the traces of at least "*ByQuotation*" methods in the PostgreSQLComponent class, and related usage from SearchUtils. That used to serve this function in a version of the API long gone. Looking at that, I recall we used to have an 'o' query parameter (from the RDF "spo" notion, i.e. "subject predicate object"), along with 'value' (for literals) and 'p' (which could be used, repeatedly, to limit which relations we look for).

Also, can you verify that the upgrade of the servlet API version plays well with our deployed tomcats? (Possibly related: I'd love to finalize and merge #271 now, since the older Gradle version gets in the way for us.)

olovy:
> TODO: shrink huge vocab, display-data and context files used in unit tests.

Fixed.

whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -506,6 +505,50 @@ class JsonLd {
...
+            } else if (obj instanceof List) {
+                descend(((List) obj).withIndex())
+            }
+        }
+
+        private void descend(List<Tuple2> nodes) {
+            for (n in nodes) {
+                path << n.second
+                node(n.first)
+                path.remove(path.size()-1)
```
olovy: List.pop() removes the first item in Groovy 2.5 (!) and removeLast() is not available before 2.5.

olovy:
> Furthermore, I'd really like to cleanup and remove the traces of at least "_ByQuotation_" methods in the PostgreSQLComponent class, and related usage from SearchUtils. That used to serve this function in a version of the API long gone. 

Done.

olovy:
> Also, can you verify that the upgrade of the servlet API version plays well with our deployed tomcats? 
Changed the provided version to 3.0. Tomcat on staging and prod is 7.0 which is compatible with Servlet API 3.0.

Bumped it because Crud.groovy was using HttpServletResponse.getStatus() which was added in 3.0.

olovy:
Fixed generation of `last` link in search result (used by all searches, not only reverse). 
Before, when `totalItems / itemsPerPage` was even the link would be wrong except on the last page.

olovy:
Tests: https://github.com/libris/lxl_api_tests/pull/2

rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
```diff
@@ -161,15 +173,60 @@ class SearchUtils {
                                      limit, offset, total)
     }
 
+    private Map findReverse(String id, String lens, int limit, int offset) {
```
lrosenstrom: :+1:

rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
```diff
@@ -399,7 +399,7 @@ class CrudSpec extends Specification {
         when:
         crud.doGet(request, response)
         then:
-        response.getStatus() == HttpServletResponse.SC_NOT_FOUND
+        response.getStatus() == HttpServletResponse.SC_NOT_ACCEPTABLE
     }
 
     def "GET /<id>/data.rdf should return 404 Not Found"() {
```
niklasl: Does this media type (application/rdf+xml) still resolve with 404?

olovy: I changed the test to match the behavior.
It seemed reasonable since the view /data.rdf does not exist
'Accept: application/rdf+xml' on the resource or data view returns 406 Not Acceptable. (this has later been changed in https://github.com/libris/librisxl/pull/364 where the Accept header is ignored if not acceptable and jsonld served)

The test for data.ttl is broken here. It was fixed on the lxl-2608 branch (https://github.com/libris/librisxl/pull/364)

niklasl:
This looks very good! A thing which vexes me now is whether we can replace this solution with the (hitherto unsued) `o` parameter, instead of introducing the new `_rev` parameter (I know I'm shilly-shallying here).

That `o` parameter, along with the optional `p` and alternative `value`, where introduced a *long* time ago (as with the now removed "quotation" mechanism). The intent then was to see if we could leverage the same API to serve [Triple Pattern Fragments](https://linkeddatafragments.org/concept/); something we never properly landed (but is still on the table for the long run). That may still warrant another data shape than the find API results though, but the parameters could be kept (we made them up anyway, as TPF defines indirected hypermedia controls).

Let's have a chat!

rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
niklasl: OK, sounds fair.

olovy:
> This looks very good! A thing which vexes me now is whether we can replace this solution with the (hitherto unsued) `o` parameter, instead of introducing the new `_rev` parameter (I know I'm shilly-shallying here).
> 
> That `o` parameter, along with the optional `p` and alternative `value`, where introduced a _long_ time ago (as with the now removed "quotation" mechanism). The intent then was to see if we could leverage the same API to serve [Triple Pattern Fragments](https://linkeddatafragments.org/concept/); something we never properly landed (but is still on the table for the long run). That may still warrant another data shape than the find API results though, but the parameters could be kept (we made them up anyway, as TPF defines indirected hypermedia controls).
> 
> Let's have a chat!

After chat:
* Renamed `_rev` to `o`
* Removed "search by relation"

PR #363, opened 2019-08-08
Feature/bfconversion3xx

klngwll:
Housekeeping update 3xx fields with some changes (except for 34X RDA-characteristics) from MARC2BF2 conversion 1.5.

* Add bib 306 - hasDuration.
* Update bib 310 frequency with subfield $b, better wrapping and spec. (3528 "310" in Uncompleted)
* Update bib 321 frequency on Serials with subfield $6, better wrapping and spec. (0 in Uncompleted)
* Add mapping of subfields $0 and $6 and $1 as null to bib 336, 337, 338.
* Update bib 340 with note and subfield $1 as null.
* Add placeholder for new field bib 341 contentAccessibility.
* Extend note on 342, 343, 355, 357
* add bib 350 acquisitionTerms. (59 "350" in Uncompleted)
* Update bib 352 to explicitly ignore subfields.
* Update bib 363-388 with either extended note or TODO
* Add supplementary to $6 in general.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -9541,8 +9541,13 @@
         "about": "_:cartographicobj",
         "property": "count"
       },
-      "$q": null,
-      "NOTE:$q": {"NOTE:record-count": 0, "TODO": "To enable future imports to EncodingFormat, we need something like onRevertPrefer 347$b to avoid collisions in export."}
+      "$d": null,
+      "$e": null,
+      "$f": null,
+      "$g": null,
+      "$i": null,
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -8311,8 +8445,11 @@
       },
       "$a": {"about": "_:contentType", "property": "label", "infer": true},
       "$b": {"about": "_:contentType", "property": "code"},
+      "$0": {"about": "_:contentType", "property": "marc:recordControlNumber"},
```
niklasl: Adding these ($0 and $6) to a potentially linked resource requires the presence of these on the linked resource. Is that the intent?

niklasl: Having thought and discussed it, this isn't worse than the already mapped $8. If these ever crop up in data, they're a signal to investigate and restructure.

PR #365, opened 2019-08-22
Map missing bib 024$d

klngwll:
Mapping of complementing code in bib 024$d. This deviates slightly from BF2 which had note as a literal type mapped which does not exist in their vocabulary. The original meaning in MARC21 is additional code.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -4404,6 +4404,7 @@
         "TODO:definedElsewhereToken": "7"
       },
       "i2": {"marcDefault": " ", "TODO": "ok to lock as undefined?"},
+      "$d": {"property": "code"},
       "$2": {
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -4404,6 +4404,7 @@
         "TODO:definedElsewhereToken": "7"
       },
       "i2": {"marcDefault": " ", "TODO": "ok to lock as undefined?"},
+      "$d": {"property": "code"},
```
niklasl: Since `code` is on par with `value` (akin to `label`), we need something "additional".

I suppose `note` made sense, apart from it not being a literal BF2 property, and apart from us attempting to avoid it (to not confuse it with `hasNote`).

klngwll: Since we have used comment for all literal notes we use this consequently here as well.

PR #367, opened 2019-08-26
Allow user with global_registrant permission to update any holding

olovy:


rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
```diff
@@ -1196,6 +1196,130 @@ class CrudSpec extends Specification {
         assert response.getStatus() == HttpServletResponse.SC_CREATED
     }
```
lrosenstrom: Awesome!

olovy:
Updated a bunch of access control tests in CrudSpec which were false positives (passing for the wrong reason). Most of them were expecting HTTP 403 and getting that because of missing `active_sigel` in the request instead of what they were trying to test.

Going forward we should try to write API tests in more of a black box manner to make them less brittle.

PR #370, opened 2019-08-27
Feature/bfconversion7xx

klngwll:
* Update conversion of 75x fields.
* Ignore 754 and 755.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -14956,8 +14956,11 @@
       "$f": {"addProperty": "marc:citySubsection", "TODO:ignored": "true?"},
       "$g": {"addProperty": "marc:otherNonjurisdictionalGeographicRegionAndFeature"},
       "$h": {"addProperty": "marc:extraterrestrialArea"},
+      "$0": {"addProperty": "marc:recordControlNumber"},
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -14956,14 +14956,21 @@
...
       "$g": {"addProperty": "marc:otherNonjurisdictionalGeographicRegionAndFeature"},
       "$h": {"addProperty": "marc:extraterrestrialArea"},
+      "$0": {"addProperty": "marc:recordControlNumber"},
+      "$1": null,
       "$2": {"property": "marc:sourceOfHeadingOrTerm"},
-      "$0": {"addProperty": "marc:uri"}
+      "$4": {"addProperty": "marc:relatorCode"},
+      "$6": {"property": "marc:fieldref"}
     },
     "753": {
```
niklasl: I'd like to see if we can possibly capture $0 or $1 here, unless more than one of $a, $b and $c is present at the same time (or just put the URI in one of them).

klngwll: I have left a note for $0 until we can take care of it properly, it is the only one which is in LC examples and has a link to a possible vocabulary.

PR #371, opened 2019-08-28
Update created, modified, recordStatus inside document when saving

olovy:


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -548,6 +554,7 @@ class PostgreSQLComponent {
                 Document dependerDoc = assembleDocument(resultSet)
                 if (linkFinder != null)
                     linkFinder.normalizeIdentifiers(dependerDoc, connection)
+                dependerDoc.setModified(modTime)
```
jannistsiroyannis: This gets a little backwards. If a record A has been merged with record B, it does not follow that record C (which happens to link to A) should have its modified-time updated.

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -875,14 +883,18 @@ class PostgreSQLComponent {
         }
     }
 
-    private PreparedStatement rigInsertStatement(PreparedStatement insert, Document doc, String changedIn, String changedBy, String collection, boolean deleted) {
+    private PreparedStatement rigInsertStatement(PreparedStatement insert, Document doc, Date timestamp, String changedIn, String changedBy, String collection, boolean deleted) {
```
jannistsiroyannis: By doing timestamp.getTime() a bunch of times instead of just once and then passing the timestamp as a parameter (as was done before), we will end up in situations where these timestamps are no longer identical (time will have moved/changed from one call to getTime() to the next).

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
olovy: OK.
But the old code updates `lddb.modified` for `dependerDoc`. Should we change that behavior?

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
olovy: No, `timestamp.getTime()` returns the same value every time.

jannistsiroyannis: Aah, I looked to quickly! You're right of course! :)

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
jannistsiroyannis: Ahh, you're right again of course! Well done :)

PR #372, opened 2019-08-28
Feature/bfconversion0xx

klngwll:
* Update BF2 conversion to 1.5 for 010-088 with notes for things which require global change.
* Add extended notes.
* Removal of ignored alphanumeric subfields.
* Add spec for 050 and 060.


klngwll:
Backed on a some hard drops, also some broken properties and resourceTypes in classifications.

PR #373, opened 2019-08-30
Fix percent-encoding of query parameters in search result navigation links

olovy:


niklasl:
@oBlissing Can you verify that this doesn't change any expectations in lxlviewer? (It *shouldn't* since it's correct; but we should similarly ensure that lxlviewer hasn't made any (incorrect) assumption about the (previously absent) encoding of parameter values in the API.)

(We'll reasonably see any such assumptions crop up during this sprint or the next, so consider this mostly as a "heads-up".)

PR #375, opened 2019-09-03
Access control: Global registrant can correct holding with missing sigel

olovy:


oBlissing:
Make sure to release together with this other PR libris/lxlviewer/pull/491

PR #376, opened 2019-09-03
Added script for removing properties from auth records

lrosenstrom:


whelktool/scripts/cleanups/2019/08/remove-properties-from-auth-records/remove-properties-from-auth-records.groovy
```diff
@@ -0,0 +1,31 @@
...
+    removeFieldFromRecord(record, "marc:subjectSubdivision", scheduledForChange)
+    removeFieldFromRecord(record, "marc:recordUpdate", scheduledForChange)
+    removeFieldFromRecord(record, "marc:modifiedRecord", scheduledForChange)
+
+    documentItem.scheduleSave(onError: { e ->
+        failedIDs.println("Failed to save ${record[ID]} due to: $e")
+    })
+}
+
+private void removeFieldFromRecord(record, String fieldName, scheduledForChange) {
```
niklasl: Just a nitpick: since the scheduledForChange object is global it doesn't have to be passed.

niklasl: Or actually, since it's typed, it might have had to be? (The behaviour of local methods differs on that.)

PR #378, opened 2019-09-04
Feature/lxl2292 cleanup sweb

jannistsiroyannis:


olovy:
Too late to the party but I think the script should be placed in 2019/09, not 2019/10?

jannistsiroyannis:
But I thought Bodil said we moved to a 6 week release cycle and the next release is in October!
But perhaps this shouldnt be run when releasing.

niklasl:
Right, good point. AFAIR, the year/month-dir have been creation dates, not running dates (as some are run when ready, some at release depending on the ready state of some mapping issue). And that's just namespacing to avoid to many/long job names.

So no worries with the `10` in practice.

niklasl:
(BTW, we regularly delete branches when merged to keep the branch list "tidy".)

PR #387, opened 2019-09-09
Handle component hold items in script for deleting Nv NC holdings

olovy:
Updated script which handles the (18) posts which have component hold posts or use `physicalLocation` instead of `shelfMark`


olovy:
Change: If there's only one component left it should be moved to top level

PR #389, opened 2019-09-10
Backend should set descriptionLastModifier

olovy:
A bit ugly with the hardcoded check of `'https://libris.kb.se/sys/globalchanges/` and sigel SEK.
But we want `lddb.changedby` to be the script URI (?)

It would cleaner to pass `changedBy` = SEK from WhelkTool and maybe also have the possibility to override Sigel in the script.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -342,6 +342,7 @@ class PostgreSQLComponent {
             if (linkFinder != null)
                 linkFinder.normalizeIdentifiers(doc, connection)
 
+            //FIXME: throw exception on null changedBy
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2207,7 +2208,20 @@ class PostgreSQLComponent {
     }
 
     private String getDescriptionChangerId(String changedBy) {
-        // FIXME: hardcoded
-        return "https://libris.kb.se/library/" + changedBy
+        //FIXME(?): hardcoded
+        // for historical reasons changedBy is the script URI for global changes
```
lrosenstrom: Would be nice to do this in another way in the future!

olovy:
Will do a nicer fix for this.

PR #392, opened 2019-09-11
Correct link to digitized version for 151 posts by Uppsala.

olovy:
Move Alvin URI from hold to bib.


olovy:
regexp is needed because of some URLs:
`"url": ["Ex. 1: http://urn.kb.se/..."]`

PR #393, opened 2019-09-12
Refactor changedBy

olovy:
For global changes:
* `descriptionLastModifier` should be `SEK`
* `generationProcess` should be `https://libris.kb.se/sys/globalchanges/...` (also when deleting)
* `lddb.changedBy` should be `https://libris.kb.se/sys/globalchanges/...`

For other updates:
* `descriptionLastModifier` should be `https://libris.kb.se/library/<Sigel>`
* `lddb.changedBy` should be `<Sigel>`

-----

Changed:
`400 BAD REQUEST` if header `XL-Active-Sigel` is missing

### DIFF
--- a/apix_export/src/main/java/whelk/apix/ExporterThread.java
+++ b/apix_export/src/main/java/whelk/apix/ExporterThread.java
@@ -3,6 +3,7 @@ package whelk.apix;
 import org.codehaus.jackson.map.ObjectMapper;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import whelk.Changer;
 import whelk.Document;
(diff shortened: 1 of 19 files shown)


olovy:
> Just to verify, does this automatically set descriptionLastModifier for WhelkTool script jobs? It shouldn't, at least not for silent jobs, and for loud, the id should be controlled by the script IMO.
> 
> (As this is a refactor, I'm assuming this was never the case and that this PR doesn't change behaviour.)

The refactoring was done after we decided and changed loud scripts to use 'SEK'.
I've added the option to override it in the scripts. And also to unset it.

Silent jobs never set descriptionLastModifier.

PR #395, opened 2019-09-12
Remove OAI-PMH harvester project

olovy:
It is obsolete.


olovy:
```
importers/src/main/resources/oaipmh.properties.in
4:# harvestClass defaults to whelk.harvester.OaiPmhHarvester if not specified
6:service.harvestClass = whelk.harvester.OaiPmhHarvester
```
?

niklasl:
I'm pretty sure we can remove the `oaipmh.properties(.in)` file. Neither that nor the `harvestClass` is used in the code anymore AFAICS. (These were part of an old homegrown IoC thingy which is long gone (after a short pico container interlude).)

olovy:
> I'm pretty sure we can remove the `oaipmh.properties(.in)` file. Neither that nor the `harvestClass` is used in the code anymore AFAICS. (These were part of an old homegrown IoC thingy which is long gone (after a short pico container interlude).)

Great. Consider it done.

PR #399, opened 2019-09-21
Feature/auth encoding level

klngwll:
Add fixed encodingLevel value n (full level) to auth Leader.
Add more specs Meeting and GenreForm to auth 008.


sarwik:
I have fixed the conversion bug that threw a NullPointerException, but in another PR: https://github.com/libris/librisxl/pull/406 (encodingLevel for hold records). **Therefor, this branch must be merged after above PR**. I have tested this branch with the bug fix together with the new spec and it works as expected.

PR #400, opened 2019-09-25
Feature/remove nonfiling chars with empty string value

lrosenstrom:
Created two scripts: a simple one for auth and a more complicated one for bib. There are only 30 holdings with nonfilingChars = " " so i skipped those.


lrosenstrom:
@jannistsiroyannis Yes, for auths all nonfilingChars = " " are located at the same place.

PR #404, opened 2019-10-02
Disable gretty hot deployment for rest

olovy:
Disable gretty hot deployment. It leaks database connections so that the local backend stops working if you touch the code too much. Annoying IMO.


olovy:
We could also change `interactiveMode` to `rebuildAndRestartOnKeyPress`.

PR #406, opened 2019-10-03
Feature/hold encoding level

sarwik:
- Drop mapping of encodingLevel of hold and revert to fixedDefault value
- Update spec with normalized to verify behavior of above. If normalized exists of leader spec, set marcSkeletons to normalized in favour of source 
- Add script that deletes all occurrences of encodingLevel in hold records


whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
```diff
@@ -51,7 +51,7 @@ class MarcFrameConverterSpec extends Specification {
                 }
 
                 if (tag == '000') {
-                    marcSkeletons[marcType] = dfn._spec[0].source
+                    marcSkeletons[marcType] = dfn._spec[0].normalized ?: dfn._spec[0].source
```
niklasl: Good catch!

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -144,13 +144,13 @@
...
-        "5": "marc:HoldingsLevel4WithPieceDesignation",
-        "m": "marc:MixedLevels",
-        "z": "marc:OtherLevel"
+        "1": null,
+        "2": null,
+        "3": null,
+        "4": null,
+        "5": null,
+        "m": null,
+        "z": null
```
niklasl: Did we need these, or could fixedDefault: " " do, so we can remove HoldingEncLevelType entirely from marcframe?

sarwik: I definitely agree that it would be best to remove HoldingEncLevelType completely. If fixedDefault was " " we could easily change to "[17]": null. But unfortunately, fixDefault is "u" which makes it more difficult.

PR #409, opened 2019-10-07
Add script to fix faulty values of bib encodingLevel

sarwik:


whelktool/scripts/cleanups/2019/10/lxl-2594-fix-faulty-encodingLevels/set-correct-bib-encodingLevel-value.groovy
```diff
@@ -0,0 +1,33 @@
...
+                             "marc:CoreLevel",
+                             "marc:FullLevel",
+                             "marc:FullLevelMaterialNotExamined",
+                             "marc:LessThanFullLevelMaterialNotExamined",
+                             "marc:MinimalLevel",
+                             "marc:PartialPreliminaryLevel",
+                             "marc:PrepublicationLevel"]
+
+selectBySqlWhere("""
+        collection = 'bib' AND data#>>'{@graph,0,encodingLevel}' IS NOT NULL
```
niklasl: If this is too slow on all bib:s you may gain speed by using a more specific select.

whelktool/scripts/cleanups/2019/10/lxl-2594-fix-faulty-encodingLevels/set-correct-bib-encodingLevel-value.groovy
```diff
@@ -0,0 +1,33 @@
...
+    """) { data ->
+
+    def (record, instance, work) = data.graph
+
+    if (!VALID_BIB_ENCODING_LEVELS.contains(record['encodingLevel'])) {
+        record.put('encodingLevel', VALID_BIB_ENCODING_LEVELS[0])
```
niklasl: Could extract first value as an `ABBR_LEVEL` constant, use that in the list and use `as Set` on that (for clarity and an admittedly negligible speedup).

klngwll:
> Has AbbreviatedLevel been agreed upon? (That get's overwritten IIRC? Some in the forum discussion thought otherwise and the issue is unclear on what the resolution has been. In the chat MinimalLevel was recommended unless I've mixed up issues?)

Moving from a numerical code to the EncLevelTypes has been a source of confusion from the beginning since they don't correspond to the swedish usage, hence when we speak of "miniminivå" it is actually AbbreviatedLevel:   

```
:AbbreviatedLevel a :EncLevelType ;
    skos:notation "3" ;
    skos:prefLabel "Abbreviated level"@en, "Miniminivå"@sv .

:MinimalLevel a :EncLevelType ;
    skos:notation "7" ;
    skos:prefLabel "Minimal level"@en, "Biblioteksnivå"@sv .
```

whelktool/scripts/cleanups/2019/10/lxl-2594-fix-faulty-encodingLevels/set-correct-bib-encodingLevel-value.groovy
sarwik: Good point. I have adjusted the sql query to only select records with encoding Levels set to hold and auth levels.

PR #410, opened 2019-10-08
Feature/merge clusters

olovy:
Merge work clusters using disjoint-set aka union–find aka merge–find data structure.

### DIFF
--- a/whelktool/src/main/java/datatool/util/DisjointSets.java
+++ b/whelktool/src/main/java/datatool/util/DisjointSets.java
@@ -1,4 +1,5 @@
 package datatool.util;
+
 import java.util.List;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


olovy:
This is a https://en.wikipedia.org/wiki/Disjoint-set_data_structure with the addition of a circular linked list for each set so that we can find all elements in a set (not only answer if two elements are members of the same set).

whelktool/src/main/java/datatool/util/DisjointSets.java
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+	
+	/**
+	 * Add a set, merging it with existing intersecting sets
+	 * 
+	 * @param set a set to be added
+	 */
+	public void addSet(Iterable<T> set) {
+		Iterator<T> i = set.iterator();
```
lrosenstrom: Streamify?

PR #414, opened 2019-10-10
Feature/declare storage and control indexing in whelktool

niklasl:


niklasl:
@olovy I see why, but Whelktool already does that in a number of places, so I ran with that.

PR #421, opened 2019-10-22
Remove blank languages

olovy:
Mechanism for removing/linking blank language nodes.

TODO:
- update definitions with language definitions from LOC (https://github.com/libris/definitions/pull/149/files)
- ~~update definitions with sv altLabel~~
- use in import


whelktool/src/main/groovy/datatool/util/DocumentSearch.groovy
```diff
@@ -0,0 +1,98 @@
+package datatool.util
+
```
lrosenstrom: Nice!

whelktool/src/test/groovy/datatool/util/DocumentSearchSpec.groovy
```diff
@@ -0,0 +1,58 @@
+package datatool.util
+
+import datatool.util.DocumentSearch.Remove
+import datatool.util.DocumentSearch.Replace
+import spock.lang.Specification
+
+import static DocumentSearch.NOP
+
```
lrosenstrom: :+1:

whelktool/scripts/cleanups/2019/10/lxl-2737-remove-redundant-blank-languages/script.groovy
```diff
@@ -0,0 +1,78 @@
...
+        'samiska (nordsamiska)'           : 'nordsamiska',
+        'samiska (lulesamiska)'           : 'lulesamiska'
+]
+
+linker = new LanguageLinker(buildLanguageMap())
+
+selectByCollection('bib') { bib ->
+    try {
+        def (record, thing, work) = bib.graph
+        if (!((String) work['@id']).endsWith('#work')) {
```
niklasl: If we need to semantically check if this is a work, either see if its type is a subclass of Work, or if thing links to it with instanceOf.

(Also, which case cropped up to require this check? It may be an indicator of something being wrong...)

niklasl:
I'd like to go through this together and understand what it entails. (I had a simpler `Map languageByLang` and a `work.language = work.language.findResults { ... }` manoeurvre in mind (with a possible generalized `toSet` utility), so I'd like to know how these extra tools help us.)

whelktool/scripts/cleanups/2019/10/lxl-2737-remove-redundant-blank-languages/script.groovy
olovy: It was an quick and dirty way to guard against (theoretical) instances without local works. Checking type would be more correct.

olovy:
> Should be `https://id.kb.se/language/gre` and `https://id.kb.se/language/grc`, right?

yes

niklasl:
Have we resolved the issue with "&" in languages, or at least deactivated that splitting until it has been? We need to find out whether these (the second and following languages) are to be mapped to e.g. `relatedTo.language` (see Jira issue for the discussion about practise). We should not run this script until these questions have been resolved.

PR #423, opened 2019-10-25
Try to avoid deadlocks when executing global changes

olovy:
It's hard to test the absence of deadlocks but I ran the following script successfully two times on QA ...
```
selectByCollection('auth') { auth ->
    try {
        auth.scheduleSave()
    }
    catch(Exception e) {
        println(e)
        e.printStackTrace()
    }
}

¯\_(ツ)_/¯
```


olovy:
This doesn't solve everything because documents may have been locked by the main update statement. e.g. auth --> auth dependencies.

olovy:
```
whelk=> select count(*) from lddb__dependencies d, lddb l, lddb l2 where l.id = d.id and d.dependsonid = l2.id and l.collection = 'auth' and l2.collection = 'auth';
 count 
-------
 63476
(1 rad)

```

PR #424, opened 2019-10-25
Avoid DB deadlocks

olovy:


whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -187,8 +188,8 @@ class Whelk implements Storage {
             }
         }
 
-        // If the number of dependers isn't too large. Update them synchronously
-        if (dependers.size() < 20) {
+        // If the number of dependers isn't too large or we are inside a batch job. Update them synchronously
+        if (dependers.size() < 20 || (batchJobThread() && indexers.activeCount() > 30) ) {
```
olovy: Using `indexers.activeCount()` is a bit sketchy...
The idea is to not let the number of reindexer threads grow unbounded if we are inside a slow batch job

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -633,13 +633,8 @@ class PostgreSQLComponent implements Storage {
             rigUpdateStatement(updateStatement, doc, modTime, changedIn, changedBy, collection, deleted)
             updateStatement.execute()
 
-            // The versions and identifiers tables are NOT under lock. Synchronization is only maintained on the main table.
             saveVersion(doc, connection, createdTime, modTime, changedIn, changedBy, collection, deleted)
             refreshDerivativeTables(doc, connection, deleted)
-            for (Tuple2<String, String> depender : getDependers(doc.getShortId())) {
-                updateMinMaxDepModified((String) depender.get(0), connection)
-                removeEmbellishedDocument((String) depender.get(0), connection)
-            }
```
olovy: Doing this inside the main transaction can lead to deadlocks even if we sort the dependencies since `doc` might also be part of the dependencies of another transaction.

whelktool/src/main/groovy/datatool/WhelkTool.groovy
```diff
@@ -642,8 +656,9 @@ class Counter {
     synchronized int getSaved() { modifiedCount.get() + deleteCount.get() }
 
     String getSummary() {
-        double docsPerSec = readCount.get() / getElapsedSeconds()
-        "read: ${readCount.get()}, processed: ${processedCount.get()}, modified: ${modifiedCount.get()}, deleted: ${deleteCount.get()} (at ${docsPerSec.round(3)} docs/s)"
+        lastSummary = System.currentTimeMillis()
+        double docsPerSec = processedCount.get() / getElapsedSeconds()
```
olovy: `processedCount` gives a truer number than readCount()``

PR #425, opened 2019-10-28
Postgres connection pool improvements

olovy:
* Use a reasonable connection pool size (~ num postgres CPUs * 2)
* Configurable pool size
* Use HikariCP instead of dbcp2
* Only create one Whelk instance in `rest` server
* Only create one MarcFrameConverter instance per Whelk
* Clean up PostgreSQLComponent


olovy:
Probably better to use a simple singleton for Whelk instead of jumping through JNDI hoops?

PR #435, opened 2019-11-08
Add findInverseBroaderRelations

olovy:
Cache refreshing is done on a separate thread so consumers only have to wait when new entries are added to the cache.

Cache size and time settings are plucked out of thin air...

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -425,24 +425,12 @@ class Whelk implements Storage {
                 .collect { it.first }
     }
 
-    List<String> findIdsLinkingTo(String id, String relation) {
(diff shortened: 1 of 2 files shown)


olovy:
Faster way to do this in one query
```
WITH RECURSIVE nested_relation AS (
    SELECT
        id
    FROM
        lddb__dependencies
    WHERE
        dependsonid = ?
        AND relation = 'broader'
    UNION
    SELECT
        d.id
    FROM
        lddb__dependencies d
        INNER JOIN nested_relation n ON d.dependsonid = n.id
        WHERE relation = 'broader'
) SELECT
    i.iri
FROM
    nested_relation n, lddb__identifiers i
WHERE
    n.id = i.id
    AND i.graphindex = 1
    AND i.mainid IS TRUE
```

PR #448, opened 2019-11-19
Add isImpliedBy(broaderId, narrowerId)

olovy:


olovy:
TODO: make whelk testable, add tests

niklasl:
Nice! Would you like to factor out the cache logic to possibly make it more testable? (And just expose that through Whelk.isImpliedBy by a fowarding method?)

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -242,29 +242,11 @@ class PostgreSQLComponent implements Storage {
...
-                "    FROM $dependenciesTableName d " +
-                "    INNER JOIN nested_relation n ON d.dependsonid = n.id " +
-                "    WHERE relation = ANY (?) " +
-                ") " +
-                "SELECT i.iri " +
-                "FROM nested_relation n, $idTableName i " +
-                "WHERE n.id = i.id " +
-                "AND i.graphindex = 1 " +
-                "AND i.mainid IS TRUE"
-     }
```
olovy: Unfortunately this can be really slow sometimes.
eg `IRI = 'https://id.kb.se/term/saogf/Sk%C3%B6nlitteratur'`
`relations = ['broader', 'broadMatch', 'exactMatch']` 

Probably because of cycles in the dependency graph that don't get culled.
Maybe it is possible to fix.

olovy:
Refactored.

olovy:
`findInverseBroaderRelations('https://id.kb.se/term/saogf/Sk%C3%B6nlitteratur')` takes ~1s without cache (whelk on my laptop, DB on QA). So caching is needed.

olovy:
> `findInverseBroaderRelations('https://id.kb.se/term/saogf/Sk%C3%B6nlitteratur')` takes ~1s without cache (whelk on my laptop, DB on QA). So the cache is needed.

Could of course be made more efficient by creating specialized getDependenciesByIri/getDependersByIri queries joining in `lddb__identifiers`

PR #451, opened 2019-11-20
Added script for mapping H-classified records to marc fiction 

lrosenstrom:


niklasl:
Has this been vetted and is ready to run, or is it to be reworked in accordance with normalization (and/or the match-broader mechanics)?

PR #453, opened 2019-11-20
Refactor "Relations"

olovy:


whelk-core/src/main/groovy/whelk/component/DependencyCache.groovy
```diff
@@ -26,7 +26,7 @@ class DependencyCache {
             .refreshAfterWrite(5, TimeUnit.MINUTES)
             .build(loader(storage.&getDependersOfType))
 
-    private LoadingCache<Tuple2<String,String>, Set<String>> dependencyCache = CacheBuilder.newBuilder()
+    private LoadingCache<Tuple2<String,String>, Set<String>> dependenciesCache = CacheBuilder.newBuilder()
             .maximumSize(1000)
```
niklasl: Factor out to constant? I wonder if 1000 is enough?

whelk-core/src/main/groovy/whelk/component/DependencyCache.groovy
```diff
@@ -49,7 +49,7 @@ class DependencyCache {
             String relation = it[0]
             String iri = it[1]
             dependersCache.invalidate(new Tuple2(iri, relation))
-            dependencyCache.invalidate(new Tuple2(docIri, relation))
+            dependenciesCache.invalidate(new Tuple2(docIri, relation))
```
niklasl: Could we have a key collision here with subject+relation vs. object+relation? (E.g. for `<X> :broader ?o` would we overwrite the cache for `?s :broader <Y>`?)

olovy: `dependersCache` is `object+relation -> [subject1, ..., subjectN]`
`dependencyCache` is `subject+relation -> [object1, ..., objectN]`

olovy: When a document is updated only its outgoing links might have changed so we clear out all cached relations where it is the subject.

niklasl: Now I see it! All good then!

whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -392,12 +392,8 @@ class Whelk implements Storage {
...
     }
 
-    boolean isImpliedBy(String broaderIri, String narrowerIri) {
-        return relations.isImpliedBy(broaderIri, narrowerIri)
-    }
-
-    Set<String> findInverseBroaderRelations(String iri) {
-        return relations.findInverseBroaderRelations(iri)
+    Relations relations() {
+        return relations
```
niklasl: getRelations() ?

olovy: For me `relations()` reads better, but I changed it to be consistent.

niklasl: I see your POV, I just think in "bean property" terms here (and will use it as `whelk.relations` in groovy...). We don't have much of a "fluent" style, so yes, consistent too.

PR #458, opened 2019-11-22
Feature/no hold on deleted

jannistsiroyannis:
…something deleted.

Also clear out unreferenced old exceptions.


jannistsiroyannis:
> Looks good! I might be missing something but don't you mean to write rs.getString(1) in the comment?
You mean this line?
if (rs.getBoolean(2)) // If deleted==true, then doc refers to a deleted document which is not ok.
resultset.get* is (wierdly enough) 1-indexed, and we want the second column of the result, so I believe this is correct. Or did I missunderstand you?

lrosenstrom:
Sorry, not in the the comment, I mean shouldn't ${rs.getBoolean(1)} --> ${rs.getString(1)} in the string parameter to the LinkValidationException?

jannistsiroyannis:
Aah! Good catch!

PR #459, opened 2019-11-29
Add support for explicitly ignored fields and subfields in marcframe

niklasl:
This silently drops these without adding them to unhandled in the
output. (Use with care!)

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -5099,7 +5099,7 @@
       "include": ["idbase"],
       "resourceType": "LCCN",
       "$a": {"property": "value"},
-      "$b": null,
+      "$b": {"ignored": true},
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -5117,14 +5117,38 @@
...
+        },
+        {
+          "name": "ingore $b",
+          "source": {
+            "010": {
+              "ind1": " ", "ind2": " ",
+              "subfields": [{"a": "a 54009564"}, {"b": "IGNORED"}]
+            }
+          },
+          "result": {
```
sarwik: Excellent! Also just needs a "normalized" for revert (for both specs) without the ignored subfield.

niklasl: Good call, will do!

PR #461, opened 2019-12-03
Updated getting-started instructions

lrosenstrom:
Is the  `delete-by-query` plugin part still needed?


niklasl:
I suspect delete-by-query is part of ES nowadays, but we must verify this by running through a clean install with integration tests.

PR #465, opened 2019-12-10
Added ISBN13 <-> ISBN10 conversion for indexing in Elastic

lrosenstrom:
```
"identifiedBy": [
                        {
                            "@type": "ISBN",
                            "value": "0070231826",
                        }
```
is on reindex transformed to

```
"identifiedBy": [
                        {
                            "@type": "ISBN",
                            "value": "0070231826",
                            "_str": "ISBN 0070231826"
                        },
                        {
                            "@type": "ISBN",
                            "value": "9780070231825",
                            "_str": "ISBN 9780070231825"
                        }
```
so the document is searchable using both ISBN10 and ISBN13 in ISBN search as well as fritext. 

- The method addTypedThingIdentifier on the Document class was previously unused. Any risks using it?
- One problem now is that both ISBN types are presented when viewing the doc from the search result list (elastic), and only the original one is shown in the full document view (jsonld).

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -19,6 +19,8 @@ import org.apache.http.params.HttpConnectionParams
 import org.apache.http.params.HttpParams
 import org.apache.http.util.EntityUtils
 import org.codehaus.jackson.map.ObjectMapper
+import se.kb.libris.utils.isbn.Isbn
+import se.kb.libris.utils.isbn.IsbnParser
 import whelk.Document
 import whelk.JsonLd
 import whelk.Whelk
@@ -236,6 +238,7 @@ class ElasticSearch {
         boolean addSearchKey = true
         copy.data['@graph'] = copy.data['@graph'].collect { whelk.jsonld.toCard(it, chipsify, addSearchKey) }
 
+        setComputedProperties(copy)
         copy.setThingMeta(document.getCompleteId())
         List<String> thingIds = document.getThingIdentifiers()
         if (thingIds.isEmpty()) {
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -250,6 +253,19 @@ class ElasticSearch {
         return framed
     }
 
+    private static void setComputedProperties(Document doc) {
+        List<String> isbnValues = doc.getIsbnValues()
+
+        if (isbnValues.size() == 1) {
```
olovy: Why limit it to documents with a single ISBN?
See e.g. https://libris.kb.se/r82z67c35265vx0/data.jsonld

olovy: What about ISBNs in `getIsbnHiddenValues()` (`["@graph", 1, "indirectlyIdentifiedBy"]`) ? Do we want to handle them too?

lrosenstrom: Now handling multple ISBN values and generating ISBN10/13 for indirectlyIdentifiedBy.

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -273,29 +291,20 @@ class ElasticSearch {
     }
 
     private static List<Isbn> getOtherIsbn(List<String> isbnValues) {
```
niklasl: Perhaps have this just take a String and use like `isbnValues.findResults { getOtherIsbn(it) }`?

lrosenstrom: Yes, that's a bit cleaner. :gem:

PR #470, opened 2019-12-12
Add a retry queue for failed ElasticSearch indexing attempts.

jannistsiroyannis:
The list is as of yet only memory resident, and retries are attempted
every 10 seconds.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -24,6 +24,7 @@ import whelk.JsonLd
 import whelk.Whelk
 import whelk.exception.WhelkRuntimeException
 
+import java.util.concurrent.LinkedBlockingDeque
 import java.util.concurrent.locks.ReentrantLock
 
 @Log
@@ -51,7 +52,7 @@ class ElasticSearch {
     }
 
     private final ReentrantLock retryLock = new ReentrantLock()
-    private final LinkedList<RetryEntry> indexingRetryQueue = new LinkedList()
+    private final LinkedBlockingDeque<RetryEntry> indexingRetryQueue = new LinkedBlockingDeque()
 
     ElasticSearch(Properties props) {
         this.elasticHosts = getElasticHosts(props.getProperty("elasticHost"))
@@ -238,13 +239,7 @@ class ElasticSearch {
             log.debug("Indexed the document ${doc.getShortId()} as ${indexName}/${collection}/${responseMap['_id']} as version ${responseMap['_version']}")
         } catch (Exception e) {
             log.error("Failed to index ${doc.getShortId()} in elastic, placing in retry queue.", e)
-            retryLock.lock()
-            try {
-                indexingRetryQueue.add(new RetryEntry(doc: doc, collection: collection, whelk: whelk))
-            } finally {
-                retryLock.unlock()
-            }
-
+            indexingRetryQueue.add(new RetryEntry(doc: doc, collection: collection, whelk: whelk))
         }
     }
 
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -90,7 +90,7 @@ class ElasticSearch {
         // httpParams.setParameter(ClientPNames.CONN_MANAGER_TIMEOUT, new Long(TIMEOUT_MS));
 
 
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -78,6 +89,23 @@ class ElasticSearch {
         // FIXME: upgrade httpClient (and use RequestConfig)- https://issues.apache.org/jira/browse/HTTPCLIENT-1418
         // httpParams.setParameter(ClientPNames.CONN_MANAGER_TIMEOUT, new Long(TIMEOUT_MS));
 
+
```
lrosenstrom: Probably better to use a scheduledExecutorService instead? https://stackoverflow.com/questions/409932/java-timer-vs-executorservice

olovy:
> Nice!
> 
> I would use e.g. a `java.util.concurrent.LinkedBlockingQueue` instead to avoid having to deal with synchronizing access to the queue.

```
private final Queue<RetryEntry> indexingRetryQueue = new LinkedBlockingQueue<>()

...
void run() {
    RetryEntry entry
    while ((entry = indexingRetryQueue.poll()) != null) {
        index(entry.doc, entry.collection, entry.whelk)
    }
}
...

```

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
jannistsiroyannis: Perhaps. But looking over that list, none of the issues are relevant here.
The one-thread thing is actually a benefit here (although the code intentionally guards against multiple retries at once too). Slight deviations in timing don't matter to us and everything that could potentially throw is already guarded in try/catch blocks.

Also, ExecutorServices are complicated and messy (imo). :(

jannistsiroyannis:
> > Nice!
> > I would use e.g. a `java.util.concurrent.LinkedBlockingQueue` instead to avoid having to deal with synchronizing access to the queue.
> 
> i.e.
> 
> ```
> private final Queue<RetryEntry> indexingRetryQueue = new LinkedBlockingQueue<>()
> 
> ...
> void run() {
>     RetryEntry entry
>     while ((entry = indexingRetryQueue.poll()) != null) {
>         index(entry.doc, entry.collection, entry.whelk)
>     }
> }
> ...
> ```

That would break down if a record could repeatedly not be indexed (because it contained something bad). In effect this would lock that thread up.

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
olovy: It would nice to set name for the timer and make it not daemon.
for example `new Timer("ElasticIndexingRetrier", false)`

jannistsiroyannis: Done!

olovy:
> > > Nice!
> > > I would use e.g. a `java.util.concurrent.LinkedBlockingQueue` instead to avoid having to deal with synchronizing access to the queue.
> > 
> > 
> > i.e.
> > ```
> > private final Queue<RetryEntry> indexingRetryQueue = new LinkedBlockingQueue<>()
> > 
> > ...
> > void run() {
> >     RetryEntry entry
> >     while ((entry = indexingRetryQueue.poll()) != null) {
> >         index(entry.doc, entry.collection, entry.whelk)
> >     }
> > }
> > ...
> > ```
> 
> That would break down if a record could repeatedly not be indexed (because it contained something bad). In effect this would lock that thread up.

Exactly, I mixed up two thoughts. Could still use the bounded loop but with a thread-safe queue

jannistsiroyannis:
> > > > Nice!
> > > > I would use e.g. a `java.util.concurrent.LinkedBlockingQueue` instead to avoid having to deal with synchronizing access to the queue.
> > > 
> > > 
> > > i.e.
> > > ```
> > > private final Queue<RetryEntry> indexingRetryQueue = new LinkedBlockingQueue<>()
> > > 
> > > ...
> > > void run() {
> > >     RetryEntry entry
> > >     while ((entry = indexingRetryQueue.poll()) != null) {
> > >         index(entry.doc, entry.collection, entry.whelk)
> > >     }
> > > }
> > > ...
> > > ```
> > 
> > 
> > That would break down if a record could repeatedly not be indexed (because it contained something bad). In effect this would lock that thread up.
> 
> Exactly, I mixed up two thoughts. Could still use the bounded loop but with a thread-safe queue

I changed it. Do you like it better this way? The other place for taking locks is still needed however.

PR #477, opened 2019-12-16
Handle exceptions thrown by Isbn utility class

lrosenstrom:
Catch exception from Isbn parse method breaking QA reindex.


lrosenstrom:
Reviewed and ok:ed by @olovy

PR #478, opened 2019-12-16
Feature/differentiated personal name

klngwll:
* Removal of marc:personalName from Person and Family, including automatic rendering of 008/32:a for Person in MARC export.
* Updated specs.


niklasl:
How come we added special provisioning for this in the first place? VIAF?

niklasl:
Never mind, I think I recall this. :)

PR #481, opened 2019-12-18
Feature/use meta on revert

niklasl:
This enables the production of $w in MARC link fields from the record controlNumber of linked things.

It is also generic, which means it will expose the record of embellished things in embellished data. We probably want to unify this with https://github.com/libris/librisxl/pull/479 (possibly by just reverting that).


sarwik:
Already merged, but LGTM as well 🎈

PR #486, opened 2019-12-20
Don't wait for MarcFrameConverter on startup

olovy:
This shaves ~10s from backend startup time by not waiting for MarcFrameConverter initialization.

* Lazy-load servlets depending on MarcFrameConverter
* Initialize MarcFrameConverter in a the background after application is loaded so that first request doesn't have to wait for it.
* Incoming requests between application started and MarcFrameConverter initialized are queued (web threads block on `Whelk.getMarcFrameConverter()`)


olovy:
This is of course mainly for faster development cycles

niklasl:
> This is of course mainly for faster development cycles

I tried to weigh that against potential runtime complexity, but it seems a fair trade-off. (That could be debated of course.)

PR #487, opened 2019-12-20
Handle framed top-level thing with meta record on revert to MARC

niklasl:
This supports e.g. the SwePub data shapes directly on revert.


niklasl:
> LGTM! Should we add testcases for this later on?

Good point! Added a simple spec to secure this.

PR #492, opened 2020-01-10
Refactor embellish

olovy:
Prepare for recursive embellish by using same embellish logic everywhere.


olovy:
:+1: 
Get/release connection overhead should be ~none with HikariCP.

PR #496, opened 2020-01-13
Keep information on incoming links in @graph,0,linksHereCount of each…

jannistsiroyannis:
… document in elastic.

DET ÄR INTE GIVET ATT DENNA SKA MERGAS!

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -192,7 +192,9 @@ class ElasticSearch {
         getOtherIsbns(doc.getIsbnHiddenValues())
                 .each { doc.addIndirectTypedThingIdentifier('ISBN', it) }
 
-        doc.data['@graph'][0]['linksHereCount'] = whelk.getStorage().getDependers(doc.getShortId()).size()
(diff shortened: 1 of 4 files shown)


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -185,12 +185,14 @@ class ElasticSearch {
...
 
-    private static void setComputedProperties(Document doc) {
+    private static void setComputedProperties(Document doc, Whelk whelk) {
         getOtherIsbns(doc.getIsbnValues())
                 .each { doc.addTypedThingIdentifier('ISBN', it) }
 
         getOtherIsbns(doc.getIsbnHiddenValues())
                 .each { doc.addIndirectTypedThingIdentifier('ISBN', it) }
+
+        doc.data['@graph'][0]['linksHereCount'] = whelk.getStorage().getDependers(doc.getShortId()).size()
```
niklasl: Looks promising! Let's just model this a bit before merging. (It could, in the output, be expressed as a link to an `?o`-search for the item, with a totalItems count.)

jannistsiroyannis:
> Nice! I think we need to URI-encode the id parameter (to e.g.. escape '#' as '%23').

Done!

PR #498, opened 2020-01-15
Feature/examples from prod

jannistsiroyannis:


importers/src/main/groovy/whelk/importer/WhelkCopier.groovy
```diff
@@ -67,18 +89,17 @@ class WhelkCopier {
         def newId = dest.baseUri.resolve(doc.shortId).toString()
         newDoc.id = newId
 
-        if (dest.storage.getMainId(newId)) {
-            dest.storeAtomicUpdate(newDoc.getShortId(), true, "xl", "WhelkCopier") {
-                it.data = newDoc.data
+        def collection = LegacyIntegrationTools.determineLegacyCollection(newDoc, dest.jsonld)
+        if (collection && collection != "definitions") {
```
lrosenstrom: Will LegacyIntegrationTools.determineLegacyCollection() ever return collection = "definitions"? When I was debugging this previously, it returned null for definitions-records.

jannistsiroyannis: You're probably right, but this doesn't hurt and (I think that) it more clearly communicates the intent to the reader this way.

importers/src/main/groovy/whelk/importer/WhelkCopier.groovy
```diff
@@ -24,20 +24,42 @@ class WhelkCopier {
...
+            else
+                doc = source.getDocument(id)
+            if (doc == null) {
+                System.err.println("Could not load document with ID: $id");
+                continue
+            }
+            doc.baseUri = source.baseUri
+
             // this links to:
             for (relDoc in selectBySqlWhere("""id in (select dependsonid from lddb__dependencies where id = '${id}')""")) {
```
olovy: Should we get more than one level of dependencies? (PostgreSQLComponent.followDependencies)

jannistsiroyannis: Perhaps, but that would then be a separate issue to this :)

PR #509, opened 2020-01-23
Feature/oaipmh without dep max modified

jannistsiroyannis:
This prepares for removing the depMaxModified column, as it removes its only use (which is in OAIPMH).
It does not (yet) remove the updating of depMaxModified (which is the real win). It only removes the need for it.


oaipmh/src/main/java/whelk/export/servlet/Helpers.java
```diff
@@ -1,21 +1,176 @@
...
 import java.time.ZonedDateTime;
-import java.util.Arrays;
-import java.util.Enumeration;
-import java.util.HashSet;
-import java.util.StringJoiner;
+import java.util.*;
 
 public class Helpers
 {
+    public static class ResultIterator implements Iterator<Document>, AutoCloseable {
```
olovy: :+1:

oaipmh/src/main/java/whelk/export/servlet/Helpers.java
```diff
@@ -1,21 +1,176 @@
...
+            {
+                if (setSpec.getRootSet().equals("auth"))
+                    resultingDocuments.push(updated);
+                if (includeDependenciesInTimeInterval && (setSpec.getRootSet().equals("bib") || setSpec.getRootSet().equals("hold")))
+                {
+                    List<Tuple2<String, String>> dependers = OaiPmh.s_whelk.getStorage().followDependers(updated.getShortId(), JsonLd.getNON_DEPENDANT_RELATIONS());
+                    for (Tuple2<String, String> depender : dependers)
+                    {
+                        String dependerId = depender.getFirst();
+                        Document dependerDocument = OaiPmh.s_whelk.loadEmbellished(dependerId);
```
olovy: Should this be `getDocument()`? (embellish is done in `ResponseCommon.emitRecord()`)

jannistsiroyannis: Good catch! Fixing it!

PR #510, opened 2020-01-23
Feature/refactoring of linkfields

sarwik:
* Remove all inherits from supplementTo (772) and replace with pattern
* Adjust incorrect mappings (e.g. otherPhysicalFormat and
  expressionOf). Existing data needs to be adjusted with global changes


whelktool/scripts/2019/11/remodel-link-entries/fix-faulty-mapping-otherPhysicalFormat.groovy
```diff
@@ -0,0 +1,59 @@
...
+scheduledForChange = getReportWriter("scheduled-for-change")
+
+PROPERTIES_TO_REMOVE = ['marc:toDisplayNote', 'marc:controlSubfield', 'partNumber', 'marc:fieldref']
+PROPERTIES_TO_MOVE = ['contribution', 'marc:displayText']
+
+selectBySqlWhere("""
+        collection = 'bib' AND data#>>'{@graph,1,otherPhysicalFormat}' IS NOT NULL
+    """) { data ->
+
+    def (record, thing, work) = data.graph
```
jannistsiroyannis: ", work" är död kod här va? (En petitess).

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -1391,8 +1391,111 @@
...
+        "_:describedBy": {
+          "NOTE": "If not possible to link, LinkFinder will change it to a Local identifiedBy property",
+          "about": "_:instance",
+          "addLink": "describedBy", "infer": true,
+          "resourceType": "Record"
+        }
+      },
+      "i1": {
+        "NOTE": "By moving 580 to 500 this indicator will not be usefull anymore",
+        "marcDefault": "0"
```
klngwll: Should we add ignored = true to indicators with default values to prevent any other values from showing up in unhandled?

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -16970,26 +17009,75 @@
...
-                      "@type": "Title",
-                      "mainTitle": "Briefec"
-                    }],
-                    "provisionActivityStatement": "provisionActivityStatement",
-                    "describedBy": [{
-                      "@type": "Record",
-                      "controlNumber": "000"
-                    }]
-                  },
-                  "marc:controlSubfield": "p1",
```
niklasl: I surmise that this was removed since it doesn't appear to have been used (in the shapes)? And/or did it cause trouble as well? (Not that I see any use for it; just to be sure.)

niklasl: It has been used though. The shapes file seems to be missing all 'marc:*' properties, for some hitherto unknown reason...

niklasl: (Correction: *my* files appear to miss them.)

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17372,38 +17561,108 @@
...
+                  "agent": {
+                    "@type": "Agent",
+                    "label": "Adam, von Bremen, 11th cent."
+                  }
+                }],
+                "hasTitle": [{
+                  "@type": "Title",
+                  "mainTitle": "Gesta Hammaburgensis ecclesiae pontificum. French."
+                }],
+                "marc:displayText": ["Online version:"]
```
niklasl: I see that we wrote that this should be put on the work in issue 2769, but why? AFAIK, marc:displayText and describedBy "go together"?

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -1391,8 +1391,111 @@
...
+      "i1": {
+        "NOTE": "By moving 580 to 500 this indicator will not be usefull anymore",
+        "marcDefault": "0"
+      },
+      "$a": {"about": "_:agent", "property": "label", "infer": true, "TODO:targetMatch": "1xx"},
+      "$b": {"about": "_:instance", "property": "editionStatement", "TODO:targetMatch": "250"},
+      "$d": {"about": "_:instance", "property": "provisionActivityStatement", "TODO:targetMatch": "260$c"},
+      "$g": {"about": "_:instance", "addProperty": "part"},
+      "$h": {"about": "_:instance", "addLink": "extent", "resourceType": "Extent", "property": "label", "NOTE:marc-repeatable": false},
+      "$i": {"about": "_:work", "addProperty": "marc:displayText"},
```
niklasl: AFAICS, since the extracted pattern is consequently wrong (using the instanceOf->LINKREF->hasInstance we need to get rid of), and scheduled to be fixed in the next round, marc:displayText as used in this pattern reflects that.

Unfortunately, this breaks the usage of marc:displayText in otherPhysicalFormat (in e.g. the digitization workflows and associated scripts), which has up until this PR been correctly located next to describedBy. And we cannot change that. (It would break practise *and* the scripts.)

So we cannot land this PR until marc:displayText is put on `_:instance` and, presumably, the change script adjusts that everywhere, ensuring that is left in the correct place for otherPhysicalFormat.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -1391,8 +1391,96 @@
...
+      "$s": {"about": "_:workTitle", "property": "mainTitle", "TODO:targetMatch": "240"},
+      "$t": {"about": "_:title", "property": "mainTitle", "TODO:targetMatch": "222, 245"},
+      "$x": {"about": "_:instance", "addLink": "identifiedBy", "resourceType": "ISSN", "property": "value", "TODO:targetMatch": "022"},
+      "$y": {"about": "_:instance", "addLink": "identifiedBy", "resourceType": "CODEN", "property": "value", "TODO:targetMatch": "030"},
+      "$w": {"aboutNew": "_:describedBy", "property": "controlNumber"},
+      "$6": {"ignored": true, "NOTE": "After inventory of usage, this has been set to ignored"},
+      "$7": {"ignored": true, "NOTE": "After inventory of usage, this has been set to ignored"},
+      "$8": {"ignored": true, "NOTE": "After inventory of usage, this has been set to ignored"},
+      "$9": {"ignored": true, "NOTE": "After inventory of usage, this has been set to ignored"},
+      "subfieldOrder": "6 7 i a s t b c d h g k ... x z w 4 8 9"
```
klngwll: In the subfieldorder there is also a subfield 4? Should this be ignored as well?

whelk-core/src/main/resources/ext/marcframe.json
niklasl: The simplest approach is probably to declare a sub-pattern for the general "displayText on works" and use that on all fields except 776, which keeps marc:displayText. Thus we don't need to rewrite much this round.

niklasl: The simplest approach is probably to move the "marc:displayText on works" to the `referenceToWork` sub-pattern, and define an "marc:displayText on instance" on referenceToInstance. Thus we don't need to rewrite much this round.

whelktool/scripts/2019/11/remodel-link-entries/fix-faulty-mapping-otherPhysicalFormat.groovy
sarwik: Stämmer. Helt onödig. Tänker bara att det blir tydligt att posten består av just de tre delarna.

whelk-core/src/main/resources/ext/marcframe.json
sarwik: I agree. The reason why I moved otherPhysicalFormat.marc:displayText to the work entity is to handle all the marc:displayText in the same manner. And after that, complete the plan to gradually resolve marc:displayText completely. But I agree, it's an unnecessary work to move otherPhysicalFormat, especially when marc:displayText should be together with describedBy and breaks existing scripts (which I was not aware of).

whelk-core/src/main/resources/ext/marcframe.json
sarwik: I haven't done anything yet with subfield 4 since it hasn't been mapped before this refactoring either. But I agree that we need to look into the use of this subfield. But maybe after a merge of this feature?

whelk-core/src/main/resources/ext/marcframe.json
sarwik: Good idea. I fix :)

whelk-core/src/main/resources/ext/marcframe.json
sarwik: Concerning the subfields that I have made ignored=true:

I have received ok to remove from the local libraries, we have been in contact with
- marc:controlSubfield
- partNumber

niklasl: Very good! Mostly needed to update my assumptions.

whelk-core/src/main/resources/ext/marcframe.json
niklasl: Great! I addressed this in ff275d5edabbf047b3aab06597e94c4859bc4502.

whelk-core/src/main/resources/ext/marcframe.json
klngwll: Yes, maybe write a TODO meanwhile so we don't forget 😄

whelk-core/src/main/resources/ext/marcframe.json
sarwik: Regarding marc:fieldref and marc:groupid I only have one note that says that we have decided to remove and ignore both of them. But I can't really remember how we came to this conclusion. Do you remember @niklasl @klngwll ?
I have checked the shapes files and they do occur, but are not very common; 338 occurrences of marc:fieldref in all link fields (of which hasPart, otherPhysicalFormat, continues and relatedTo have the most occurrences) and only one occurrence of marc:groupId.

PR #511, opened 2020-01-24
Fix handling of connections outside storage (OAI-PMH, marc_export)

olovy:
Avoid risk of connection related deadlocks in multithreaded situations where a request is holding a connection while calling a `PostgreSQLComponent` method that grabs another connection.

Do this by creating a second connection that is used for these situations, i.e code outside  `PostgreSQLComponent` that performs its own queries.
Not strictly needed everywhere, but changed for consistency.

Other more complex solutions could be to have some kind RequestContext ot ThreadLocal connections etc.


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -55,6 +59,9 @@ class PostgreSQLComponent implements Storage {
     private static final String driverClass = "org.postgresql.Driver"
 
     private HikariDataSource connectionPool
+    private HikariDataSource outerConnectionPool
+    private final Object outerConnectionPoolLock = new Object()
```
jannistsiroyannis: This is only used when creating a new outer pool, no? There's probably no need for a separate locking-object then, "this" could be enough.

olovy: True, YAGNI

PR #512, opened 2020-01-27
Feature/lxl 2726

jannistsiroyannis:


whelktool/scripts/cleanups/2020/01/lxl-2726/lxl-2726.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+        components.add(hold.graph[1]) // Shelf mark may also be on the outer Item (instead of component list)
+        for (Map component : components) {
+            if (component.shelfMark) {
+                def labels = component.shelfMark.label
+                if (labels instanceof List) {
+                    if (labels.size != 1 || !labels[0].equals(oldShelfMark))
+                        unexpectedRecordState.println("$iri had shelfMark: ${labels}, expected $oldShelfMark")
+                    labels.clear()
+                    labels.add(newShelfMark)
```
olovy: Is this what we want to do?
Is this an existing case in the actual data?

whelktool/scripts/cleanups/2020/01/lxl-2726/lxl-2726.groovy
```diff
@@ -0,0 +1,66 @@
...
+                } else {
+                    component.shelfMark.label = newShelfMark
```
olovy: Not checking the old value. Is this on purpose?
Not clear from the description of the issue I guess,

whelktool/scripts/cleanups/2020/01/lxl-2726/lxl-2726.groovy
jannistsiroyannis: Shelfmarks that no longer have the expected old value? Yes, there's a (small) bunch of those.

PR #517, opened 2020-01-29
Remove column lddb.depmaxmodified

olovy:
Obsoleted by #509

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -89,7 +89,6 @@ class PostgreSQLComponent implements Storage {
     protected String GET_THING_MAIN_IRI_BY_SYSTEMID
     protected String GET_DOCUMENT_BY_IRI
     protected String GET_MAX_MODIFIED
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -336,18 +335,6 @@ class PostgreSQLComponent implements Storage {
 
         GET_MAX_MODIFIED = "SELECT MAX(modified) from $mainTableName WHERE id IN (?)"
```
lrosenstrom: Just a thought popping up when reading all these min-max-dep-modified statements: wouldn't it be nice to call it latest modified instead? For some reason my brain always does: max modified = maximally modified, oh wait...

olovy: Ditto

jannistsiroyannis: In my brain, the column is a numeric value (sort of), and we constantly use < > (more than, less than) operators on them. And so, max/min makes the most sense to me. But, im not religious on the matter.

PR #520, opened 2020-01-31
Update reverse link counter inplace

olovy:


whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -409,6 +407,11 @@ class Whelk implements Storage {
         return systemId
     }
 
+    private String getLegacyCollection(Document doc) {
+        return LegacyIntegrationTools.determineLegacyCollection(doc, jsonld)
+                ?: storage.getCollectionBySystemID(doc.getShortId())
```
niklasl: Ah :)

PR #521, opened 2020-02-03
Feature/oaipmh without dep max modified (#509)

olovy:
* Make emitRecord take Documents instead of sql resultsets.

* Make both from and until optional again.

* Make the adhoc tests at least possibly reusable.


olovy:
This is the same changeset as #509

PR #525, opened 2020-02-05
Store document data in Unicode NFC form

olovy:
Convert to unicode NFC form before saving to DB. We probably want to do this earlier (e.g. `Document` constructor) so that the data is always in NFC form internally before doing any other operation or checks on it. But starting this way we can run a simple script on correct all existing data.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/util/Unicode.groovy
+++ b/whelk-core/src/main/groovy/whelk/util/Unicode.groovy
@@ -3,6 +3,12 @@ package whelk.util
 import java.text.Normalizer
 
 class Unicode {
+
+    /**
+     * Typographical ligatures from the "Alphabetic Presentation Forms" unicode block
+     * that are not covered by NFC but we want to normalize.
+     * https://www.unicode.org/charts/PDF/UFB00.pdf
+     */
     private static final List FORBIDDEN_UNICODE_CHARS = [
             'ﬀ', // 'LATIN SMALL LIGATURE FF'
             'ﬃ', // 'LATIN SMALL LIGATURE FFI'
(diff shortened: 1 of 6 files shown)


whelk-core/src/main/groovy/whelk/util/Unicode.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+class Unicode {
+    private static final List FORBIDDEN_UNICODE_CHARS = [
+            'ﬀ', // 'LATIN SMALL LIGATURE FF'
+            'ﬃ', // 'LATIN SMALL LIGATURE FFI'
+            'ﬄ', // 'LATIN SMALL LIGATURE FFL'
+            'ﬁ', // 'LATIN SMALL LIGATURE FI'
+            'ﬂ', // 'LATIN SMALL LIGATURE FL'
+            'ﬅ', // 'LATIN SMALL LIGATURE LONG S T'
+            'ﬆ', // 'LATIN SMALL LIGATURE ST'
```
niklasl: Are these cherry-picked or from a standardized set (e.g. of typographic ligatures)? Can we reference something, e.g. https://en.wikipedia.org/wiki/Orthographic_ligature or similar as a "source"?)

whelktool/scripts/cleanups/2020/01/normalize-unicode.groovy
```diff
@@ -1,10 +1,10 @@
...
+import whelk.util.Unicode
 
 PrintWriter failedIDs = getReportWriter("failed-IDs")
 PrintWriter scheduledForUpdate = getReportWriter("scheduled-for-update")
 
 selectBySqlWhere("deleted is false", silent: false, { docItem ->
-    if (!Normalizer.isNormalized(docItem.doc.getDataAsString(), Normalizer.Form.NFC)) {
+    if (!Unicode.isNormalized(docItem.doc.getDataAsString())) {
         scheduledForUpdate.println("${docItem.doc.getURI()}")
         docItem.scheduleSave()
```
niklasl: I *presume* we want to version this change (as is done here)? How many documents were affected?

olovy: Lost network connection during dry-run. Rerunning now.

PR #532, opened 2020-02-10
When copying data from prod XL, retain the hardcoded library URIs.

jannistsiroyannis:
Without this change, these URIs were being changed from
https://libris.kb.se/library/_
to
http://localhost:5000/library/_ (etc)

which caused all sorts of weird glitches _in test environments_.
OAIPMH and batch import were affected for example.

However, since the two URI forms coincide on PROD, no _real_ damage was done.

### DIFF
--- a/importers/src/main/groovy/whelk/importer/WhelkCopier.groovy
+++ b/importers/src/main/groovy/whelk/importer/WhelkCopier.groovy
@@ -84,14 +84,14 @@ class WhelkCopier {
 
         def libUriPlaceholder = "___TEMP_HARDCODED_LIB_BASEURI"
         def newDataRepr = doc.dataAsString.replaceAll( // Move all lib uris, to a temporary placeholder.
-                source.baseUri.resolve("library/").toString().replace(".", "\\."),
-                libUriPlaceholder)
+                '"\\Q' + source.baseUri.resolve("library/").toString() + '\\E',
+                '"' + libUriPlaceholder)
         newDataRepr = newDataRepr.replaceAll( // Replace all other baseURIs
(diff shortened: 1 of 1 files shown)


importers/src/main/groovy/whelk/importer/WhelkCopier.groovy
```diff
@@ -82,8 +82,17 @@ class WhelkCopier {
     void save(doc) {
         System.err.println "[$copied] Copying $doc.shortId from $source.baseUri to $dest.baseUri"
 
-        def newDataRepr = doc.dataAsString.replaceAll(/"${source.baseUri}/,
-                '"'+dest.baseUri.toString())
+        def libUriPlaceholder = "___TEMP_HARDCODED_LIB_BASEURI"
+        def newDataRepr = doc.dataAsString.replaceAll( // Move all lib uris, to a temporary placeholder.
+                source.baseUri.resolve("library/").toString().replace(".", "\\."),
```
niklasl: Did you intend to remove the leading `"` here? It was added to ensure that only values starting with the baseUri are affected.

It's good to escape the dot (not doing so was just sloppy, albeit practically OK). But be more safe, surround the baseUri with `\Q` and `\E` (e.g. `/"\Q${source.baseUri}\E/`, which is what `java.util.regex.Pattern.quote` does).

Also note that the replacement string is a regular string, so the only thing ever needing an escape there is `$` (which is what `java.util.regex.Matcher.quoteReplacement` does).

jannistsiroyannis: I pushed a new version, see if you like it better!

PR #537, opened 2020-02-13
Feature/faster recursive embellish

olovy:
* Replace CTE queries with regular queries and do recursion (actually iteration) in application.
* cache card dependencies
* always do `reindexAffected` on backgronud thread (except for whelktool) since calculating dependers set can be slow

makes reindex go from painfully slow to done in 1.8 hours on xlbuild->QA


olovy:
TODO
* limit number of reindexing threads by using fixed pool (+ "caller runs" when exhausted to get some backpressure in API)
* persistent reindexing queue?

PR #543, opened 2020-02-21
Ignore recordStatus in checksum calculation

olovy:
recordStatus is generated from created + modified + deleted
this means that right now, created and modified are implicitly part of checksum although they should be ignored.

Let's discuss pros/cons
e.g. requires recalculation of all checksums in `lddb`


jannistsiroyannis:
Vi borde merga detta, men kanske skulle den här PRn också innehålla en global körning (som bara sparar om samtliga poster)?

olovy:
> Vi borde merga detta, men kanske skulle den här PRn också innehålla en global körning (som bara sparar om samtliga poster)?

Ja!

PR #547, opened 2020-02-25
ES index: don't shrink embellishments to chips

olovy:
When embellishing doc for index, don't make chips out of everything that is included.
Because we want e.g.  Instance:instanceOf-> Work:contribution -> Agent to be indexed.

Also removed the `filterOutNonChipTerms` parameter to embellish since it was only used here.
Also removed `ImporterMain::reindexToStdout` since it was not working.

### DIFF
--- a/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
+++ b/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
@@ -56,42 +56,6 @@ class ImporterMain {
         new CardRefresher(whelk).refresh(collection)
     }
 
-    @Command(args='[COLLECTION] [no-embellish|no-cache]')
-    void reindexToStdout(String collection=null, String directive=null) {
-        if (directive == null && collection?.indexOf('-') > -1) {
-            directive = collection
-            collection = null
-        }
-        boolean useCache = directive != 'no-cache'
-        boolean doEmbellish = directive != 'no-embellish'
-
-        Whelk whelk = Whelk.createLoadedCoreWhelk(props, useCache)
-
-        println "Creating whelk with dummy ElasticSearch"
-        println "- collection: $collection"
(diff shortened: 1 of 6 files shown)


importers/src/main/groovy/whelk/importer/ImporterMain.groovy
```diff
@@ -56,42 +56,6 @@ class ImporterMain {
...
-        Whelk whelk = Whelk.createLoadedCoreWhelk(props, useCache)
-
-        println "Creating whelk with dummy ElasticSearch"
-        println "- collection: $collection"
-        println "- directive: $directive"
-        println "- useCache: $useCache"
-        println "- doEmbellish: $doEmbellish"
-
-        whelk.elastic = new ElasticSearch("", "", "") {
-            Tuple2<Integer, String> performRequest(String method,
```
olovy: This method doesn't exist...

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -219,11 +219,6 @@ class ElasticSearch {
         return framed
     }
 
-    void embellish(Whelk whelk, Document document, Document copy) {
-        boolean filterOutNonChipTerms = true // Consider using false here, since cards-in-cards work now.
```
olovy: This is the actual fix

importers/src/main/groovy/whelk/importer/ImporterMain.groovy
niklasl: We've either deleted that in a cleanup, or I threw this together with a local change. :stuck_out_tongue_closed_eyes: It was for performance testing gone by, so no need to keep it around.

niklasl:
Good! Did the ES index size increase significantly by this?

PR #549, opened 2020-02-25
Feature/improve soft search

olovy:
See LXL-2954


olovy:
TODO: softmatcher only works on top-level `_str` field

ebengtsson:
ok

PR #551, opened 2020-02-26
Feature/move linkfield to instance

sarwik:
* Move isPartOf, otherEdition, issuedWith and dataSource from Work to Instance and reference to an Instance object
* Remove marc:displayText from isPartOf
* Move isPartOf.part to a part property directly on Instance (Align with PR https://github.com/libris/definitions/pull/181)
* Move hasNote a marc:LinkingEntryComplexityNote to Instance

## Checklist:
- [x] I have run the unit tests. `gradlew integTest`
- [x] I have run script as dry-run on QA
- [x] Export issue: How to handle the potential collision between repeated bib 250 $a and bib 775 $b. Both maps to otherEdition.editionStatement. Fixed with script remove-repeated-editionstatements.groovy and https://jira.kb.se/browse/LXL-3022
- [x] Move hasSeries.instanceOf.marc:displayText and hasSubSeries.instanceOf.marc:displayText to Move hasSeries/hasSubSeries.marc:displayText. Will be fixed with https://jira.kb.se/browse/LXL-3025

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -17548,6 +17548,7 @@
       ]
     },
     "773": {
+      "NOTE": "We deviate from https://www.loc.gov/bibframe/mtbf/ as we do not link from the Work with hasInstance.",
       "include": ["referenceToInstanceSansDisplayText"],
       "aboutEntity": "?thing",
       "aboutAlias": "_:instance",
@@ -17903,6 +17904,7 @@
       ]
     },
     "775": {
+      "NOTE": "We deviate from https://www.loc.gov/bibframe/mtbf/ as we do not link from the Work with hasInstance.",
       "include": ["referenceToInstance"],
       "aboutEntity": "?thing",
       "aboutAlias": "_:instance",
@@ -18027,14 +18029,12 @@
       ]
     },
     "776": {
-      "NOTE": "We deviate from BF2 as we do not link from the Work with hasInstance.",
       "include": ["referenceToInstance"],
       "aboutEntity": "?thing",
       "aboutAlias": "_:instance",
       "addLink": "otherPhysicalFormat",
       "resourceType": "Instance",
       "i2": {"ignored": true, "marcDefault": "8"},
-      "NOTE:$i": "Not using same pattern as all other linkfields since only 776 is correctly mapped to Instance. Not creating new pattern because aiming for get rid of displayText",
       "_spec": [
         {
           "source": {"776": {"ind1": "0", "ind2": "8", "subfields": [
@@ -18705,6 +18705,7 @@
       ]
     },
     "786": {
+      "NOTE": "We deviate from https://www.loc.gov/bibframe/mtbf/ as we do not link from the Work with hasInstance.",
       "include": ["referenceToInstance"],
       "aboutEntity": "?thing",
       "aboutAlias": "_:instance",
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -17130,6 +17130,7 @@
             {"i": "displayText"},
             {"t": "Publikation"},
             {"c": "Särskiljande tillägg"},
+            {"g": "s. 144-145"},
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17618,37 +17647,127 @@
...
+                    "label": "Robson, James,"
+                  }
+                }]
+              },
+              "marc:toDisplayNote": true
+            }]
+          }}
+        },
+        {
+          "name": "Example from https://www.kb.se/katalogisering/Katalogisering/artikelindexering/exempel/#14.",
```
niklasl: Good example. It's not obvious whether this should be part of a spec for 787 instead though, given that it spans multiple fields.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17508,12 +17548,13 @@
...
     "773": {
-      "include": ["referenceToWork"],
-      "aboutEntity": "?work",
-      "aboutAlias": "_:work",
+      "include": ["referenceToInstanceSansDisplayText"],
+      "aboutEntity": "?thing",
+      "aboutAlias": "_:instance",
       "addLink": "isPartOf",
-      "resourceType": "Aggregate",
+      "resourceType": "Instance",
```
niklasl: The global change needs to take this into account as well. I.e. change all uses of `"@type": "Aggregate"` to `"Instance"`.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -17508,12 +17548,13 @@
       ]
     },
     "773": {
-      "include": ["referenceToWork"],
-      "aboutEntity": "?work",
-      "aboutAlias": "_:work",
+      "include": ["referenceToInstanceSansDisplayText"],
+      "aboutEntity": "?thing",
```
niklasl: It could be prudent to add a note like: `"NOTE": "We deviate from https://www.loc.gov/bibframe/mtbf/ as we do not link from the Work with hasInstance.",` to all fields where we do so. (And update the note in 776 which just says "BF2" (which is incorrect as BF2 itself allows this).)

whelk-core/src/main/resources/ext/marcframe.json
niklasl: Hm, maybe it (implicitly) already do so with the newInstanceObject?

whelktool/scripts/2020/01/remodel-link-fields/move-linkfield-to-instance.groovy
```diff
@@ -0,0 +1,145 @@
+/*
```
niklasl: Nice script! Minor thoughts:
* The script does take legacy properties into *some* account. Might update the NOTE to explain that they're temporarily kept to be cleaned up in the following steps.
* `docid` is passed at different positions throughout the functions. Might be prudent to be consequent here (or, *perhaps*, pass a logging Closure if it makes things clearer).

whelk-core/src/main/resources/ext/marcframe.json
sarwik: Yes it does. Before adding to newInstanceObject, the script sets type to 'Instance' (line 125). The same applies to dataSources that used to have type 'Dataset '

whelktool/scripts/2020/01/remodel-link-fields/move-linkfield-to-instance.groovy
sarwik: Fix it 👍

whelk-core/src/main/resources/ext/marcframe.json
sarwik: Fix it 👍

whelk-core/src/main/resources/ext/marcframe.json
sarwik: I'm not sure it fits best with 773 or 787. That's an example of 'Bidragskatalogisering'. Normally (in marc21) this is described with 773 in combination with 787. I'm leaving this for now. We can move it when we gets to reviewOf examples.

PR #552, opened 2020-02-27
fromJsonMap

kaipoykio:
Since marcFrameConverter.runRevert() returns marcJsonData as Map we would like to convert to MarcRecord directly without doing extra roundtrips converting to jsonstring and back.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/JSONMarcConverter.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/JSONMarcConverter.groovy
@@ -21,12 +21,19 @@ import org.codehaus.jackson.map.ObjectMapper
 @Log
 class JSONMarcConverter {
     protected final static ObjectMapper mapper = new ObjectMapper()
-    static MarcRecord fromJson(String json) {
+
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/converter/JSONMarcConverter.groovy
```diff
@@ -21,12 +21,19 @@ import org.codehaus.jackson.map.ObjectMapper
 @Log
 class JSONMarcConverter {
     protected final static ObjectMapper mapper = new ObjectMapper()
-    static MarcRecord fromJson(String json) {
+
+    static MarcRecord fromJson(String marcJson) {
+        Map resultJson = mapper.readValue(marcJson, Map)
+
+	return fromJsonMap(resultJson)
```
olovy: Indentation is off

niklasl:
@kaipoykio We follow [git flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) practises, so ideally these branches should be named e.g. `feature/fromJsonMap`.

kaipoykio:
Ok ;)

PR #557, opened 2020-02-28
URL-encode the sigel-part of library-uris, to match converter output.

jannistsiroyannis:


olovy:
LGTM!

Fixade PostgreSQLComponent också

niklasl:
Good fix!

PR #558, opened 2020-02-28
Feature/lxl 1737

jannistsiroyannis:


marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
```diff
@@ -22,16 +25,16 @@
 import java.sql.*;
 import java.util.*;
 
-public class TotalExport
+public class MarcCliExport
```
olovy: :+1:

marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
```diff
@@ -237,6 +281,16 @@ private PreparedStatement getAllHeldURIsStatement(ExportProfile profile, Connect
         return preparedStatement;
     }
 
+    private PreparedStatement getAllSaoStatement(Connection connection)
+            throws SQLException
+    {
+        String sql = "SELECT id FROM lddb WHERE data#>>'{@graph,1,inScheme,@id}' = 'https://id.kb.se/term/saogf' AND deleted = false";
```
niklasl: That is SAOGF for GenreForm, is that what's intended here? SAO are the Subject headings.

marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
```diff
@@ -56,6 +59,13 @@ public static void main(String[] args)
         if (args.length != 2 && args.length != 1)
             printUsageAndExit();
 
+        if (args[0].equals("--sao"))
```
niklasl: We should probably provide a filter on inScheme instead of hardcoding this.

marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
jannistsiroyannis: No, thats my mistake. I discussed it with Sara-Lean, but probably i misunderstood her. Whats the correct way to find the SAO subject headings?

marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
jannistsiroyannis: I disagree. This is for MARC export for a very specific legacy-purpose, and we should not be building new complex generic structures for other "future" sorts of MARC exports.

niklasl: I figured that was the reason. If we can guarantee that noone wants terms from any other scheme, that is fair. (We already have complex, hardcoded specific constructs for all sorts of MARC exports. Just note that `sao` is way more domain specific than `bib`, `auth` or `hold`.)

marc_export/src/main/java/whelk/export/marc/MarcCliExport.java
jannistsiroyannis: Fixed!

niklasl: Great!

PR #567, opened 2020-03-09
Minor fix: Set instance @type when only work properties exists

sarwik:


sarwik:
> LGTM, though I would prefer curly braces around each if-statement :)

Good to know! :)

PR #571, opened 2020-03-18
Feature/lxl 1455 allow isbn in remote search

lrosenstrom:
This pull request depends on additions in the metaproxy xml-file: 

`http://mproxy.libris.kb.se/db_Metaproxy.xml`

Correct behaviour has been verified with a local copy of the file where each library element has an additional child element:

```
<libraryCode id="NLH">
   <name>National Széchényi Library</name>
   <alternativeName>National Library of Hungary</alternativeName>
   <country>Ungern</country>
   <comment/>
   <address>http://regi.oszk.hu/index_en.htm</address>
   <requiresDcIdentifier>true</requiresDcIdentifier>
   <port>2222</port>
</libraryCode>
```

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/RemoteSearchAPI.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/RemoteSearchAPI.groovy
@@ -248,30 +248,7 @@ class RemoteSearchAPI extends HttpServlet {
                     queue.shutdown()
                 }
                 // Merge results
-                def results = ['totalResults':[:],'items':[]]
-                int biggestList = 0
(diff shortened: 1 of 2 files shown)


rest/src/main/groovy/whelk/rest/api/RemoteSearchAPI.groovy
```diff
@@ -286,6 +246,65 @@ class RemoteSearchAPI extends HttpServlet {
         }
     }
 
+    String mergeResults(List<MetaproxySearchResult> resultsList) {
```
olovy: 👍

PR #574, opened 2020-03-20
Add note cleanup script.

jannistsiroyannis:


whelktool/scripts/cleanups/2020/03/lxl-3047.groovy
```diff
@@ -0,0 +1,17 @@
...
+String where = "collection = 'bib' and data#>>'{@graph,2}' is not null"
+selectBySqlWhere(where) { data ->
+
+    def (record, instance, work) = data.graph
+
+    if (work.note != null) {
+        if (work.hasNote == null) {
+            work["hasNote"] = []
+        }
+        work.hasNote.add(work.note)
```
sarwik: This just adds work.note directly to hasNote, right?
hasNote should consist of a map:  [ @type: Note, label: work.note ]

And a thought: If hasNote already exists, can we be sure that existing hasNote always is a list?

PR #588, opened 2020-04-03
Script for removing enumerations from work

lrosenstrom:


whelktool/scripts/cleanups/2020/03/lxl-3070/remove-enumerations-from-work
```diff
@@ -0,0 +1,50 @@
+scheduledForChange = getReportWriter("scheduledForChange")
+musicFormatType = "https://id.kb.se/marc/MusicFormatType-"
+musicPartsType = "https://id.kb.se/marc/MusicPartsType-"
+audienceType = "https://id.kb.se/marc/AudienceType-"
+marcParts = "marc:parts"
+musicFormat = "musicFormat"
+intendedAudience = "intendedAudience"
+
+selectByCollection('bib') { bib ->
+    work = bib.doc.data['@graph'][2]
```
sarwik, line 10: Is there any risk that any of these enumerations may occur in a record that has no graph[2]? Or do you plan to run this script after we have ensured that all work entities will be placed in graph[2] (another issue)?

Otherwise, LGTM!

lrosenstrom: Added check for work as instanceOf in thing!

PR #591, opened 2020-04-06
Add document normalization step when saving

olovy:
Normalize documents before saving and in remote search results.

Problems with this implementation:
* Adds several seconds to whelk startup time (maybe because I tested on a remote Postgres instance)
* Adds some KBV specific stuff to whelk-core
* Need to rethink how to handle DB connections if we add e.g. a normalizer using Whelk.relations

Add index on `data #> '{@graph,1,@type}` in `lddb` so that we don't have to use elastic to find definitions?

### DIFF
--- a/whelk-core/src/main/groovy/se/kb/libris/BlankNodeNormalizers.groovy
+++ b/whelk-core/src/main/groovy/se/kb/libris/BlankNodeNormalizers.groovy
@@ -1,0 +1,70 @@
+package se.kb.libris
+
+import groovy.util.logging.Log4j2 as Log
+import whelk.Document
+import whelk.JsonLd
+import whelk.Whelk
+import whelk.component.DocumentNormalizer
+import whelk.filter.BlankNodeLinker
+import whelk.filter.LanguageLinker
+
+import java.sql.Connection
+
+@Log
+class BlankNodeNormalizers {
+
+    static DocumentNormalizer language(Whelk whelk) {
+        LanguageLinker linker = new LanguageLinker()
+        loadDefinitions(linker, whelk)
+
+        return { Document doc, Connection connection ->
+            linker.linkAll(doc.data, 'associatedLanguage')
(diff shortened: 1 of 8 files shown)


whelk-core/src/main/groovy/se/kb/libris/BlankNodeNormalizers.groovy
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    static DocumentNormalizer contributionRole(Whelk whelk) {
+        BlankNodeLinker linker = new BlankNodeLinker(
+                'Role', ['code', 'label', 'prefLabelByLang', 'altLabelByLang', 'hiddenLabel'])
+        loadDefinitions(linker, whelk)
+
+        return { Document doc, Connection connection ->
+            Map work = getWork(whelk.jsonld, doc)
```
olovy, line 32: search whole document instead?
So that we don't need work handling code here.

niklasl:
To get rid of the hardcoded KBV domain logic, how about annotating the involved classes (Role, Language) with something like `:category :heuristicIdentity`? In theory, we could use `owl:hasKey` to specify exactly what goes into the identity (`code`, `label`...), but that's too strict and wouldn't work in this more heuristic situation. Categorizing the classes whose members have "sparse" uniqueness and only hardcoding the rather "core" properties of codes and labels used to identify them might be a fair tradeoff...

olovy:
> To get rid of the hardcoded KBV domain logic, how about annotating the involved classes (Role, Language) with something like `:category :heuristicIdentity`? In theory, we could use `owl:hasKey` to specify exactly what goes into the identity (`code`, `label`...), but that's too strict and wouldn't work in this more heuristic situation. Categorizing the classes whose members have "sparse" uniqueness and only hardcoding the rather "core" properties of codes and labels used to identify them might be a fair tradeoff...

Sounds like a good idea!

PR #597, opened 2020-04-08
Feature/shrink index

olovy:
See LXL-3081

Remove a lot of stuff from the ES index.

Summary of changes:
* Only embellish one level with chips for ES index. (IS IT ENOUGH?)
  * (because of LXL-3094, `instanceOf` is included "for free")
* Only include Swedish and English xByLang
* No _str in `meta` of embellishments

**Example**
Before:
http://kblxes01.kb.local:9200/libris_qa/bib/j2vzn03v08lfhrm
~972 fields (`curl ... | jq ._source | grep ':' | wc -l`)
27647 bytes

After:
http://es01-dev.libris.kb.se:9200/libris_dev/bib/j2vzn03v08lfhrm
~388 fields
11830 bytes

### DIFF
--- a/whelktool/scripts/examples/updatebyid.groovy
+++ b/whelktool/scripts/examples/updatebyid.groovy
@@ -1,0 +1,5 @@
+selectByIds([
+    'https://libris-dev.kb.se/j2vzn03v08lfhrm',
+]) {
+    it.scheduleSave()
+}
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -355,9 +355,15 @@ class Whelk implements Storage {
         }
     }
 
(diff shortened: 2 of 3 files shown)


whelktool/scripts/examples/updatebyid.groovy
```diff
@@ -0,0 +1,5 @@
+selectByIds([
```
jannistsiroyannis, line 1: Cute :)

olovy:
> Så bara chip-saker blir sökbara?

I de länkade sakerna, ja.

whelktool/scripts/examples/updatebyid.groovy
olovy: Hade problem med python-snurran i lxlviewer lokalt och behövde ett sätt att uppdatera och det råkade slinka med i PR :)

PR #603, opened 2020-04-16
Add mapping and spec for auth 430

klngwll:
### Checklist
- [x] Run integTest

### Description

Mapping of auth 430 (hasVariant / Work)

Two things: 
1. $o is now mapped to `marc:arrangedStatementForMusic` and $s being mapped to `version` which differ from previous mappings which need to be revised. In BF2 they are both mapped to version. Not sure if we should aim for that or just add TODO for now.
2. Interpunction is weird as always.

[LXL-2753](https://jira.kb.se/browse/LXL-2753), [LXL-1625](https://jira.kb.se/browse/LXL-1625)
Script to fetch the unhandled data will be in a different PR/issue (link when available).

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -330,10 +330,10 @@
 
       "$l": {"about": "_:work", "link": "language", "resourceType": "Language", "property": "label", "infer": true},
       "$m": {"about": "_:work", "addLink": "musicMedium", "resourceType": "MusicMedium", "property": "label", "leadingPunctuation": ","},
-      "$o": {"about": "_:work", "property": "version"},
+      "$o": {"about": "_:work", "property": "version", "TODO": "musicArrangement?"},
 
       "$r": {"about": "_:work", "property": "musicKey", "leadingPunctuation": ","},
-      "$s": {"about": "_:work", "property": "marc:version"}
+      "$s": {"about": "_:work", "property": "marc:version", "TODO": "version"}
     },
     "i1-nonfiling": {
       "i1": {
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -22170,9 +22170,119 @@
...
+          "addLink": "hasTitle", "resourceType": "Title",
+          "embedded": true
+        }
+      },
+      "$a": {"about": "_:title", "property": "mainTitle"},
+      "$d": {"addProperty": "legalDate"},
+      "$f": {"property": "originDate"},
+      "$g": {"link": "genreForm", "resourceType": "GenreForm", "property": "label"},
+      "$h": {"ignored": true},
+      "$k": {"addProperty": "natureOfContent"},
```
sarwik, line 22189: $k differs also from the mapping of pattern titledetails, used by all other X30 fields. It would be a good idea to create an issue to change mapping of $k for all other X30 fields. Also includes global changes of existing data.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -22170,9 +22170,119 @@
         }
       ]
     },
-    "430": {
-      "ignored": true,
-      "TODO": "map, alternativeForm a Work (like a structured altLabel)"
+    "430" : {
+      "aboutEntity": "?thing",
+      "addLink": "hasVariant",
+      "include": ["i2-nonfiling"],
```
sarwik, line 22176: For all other X30 we are using the pattern titledetails but obviously we can not use it for 430 as mapping of several properties differs. I think it would be a good idea to update titledetails to be able to be used for all X30 fields by grouping links/properties that still are the same, OR just add an TODO/NOTE that we are aware that 430 are different for $o, $s and $k and all X30 will be aligned.

klngwll: Absolutely, I will create new issue for refactoring as well.

PR #616, opened 2020-04-27
Add illustrativeContent cleanup script.

jannistsiroyannis:


whelktool/scripts/cleanups/2020/04/lxl-3048.groovy
```diff
@@ -0,0 +1,32 @@
...
+    if (work.illustrativeContent instanceof List) {
+        Iterator it = work.illustrativeContent.iterator()
+        while (it.hasNext()) {
+            Object entity = it.next()
+            if (!entity.containsKey("@id")) {
+                it.remove()
+                changed = true
+            }
+        }
+    } else if ( (!work.illustrativeContent instanceof Map) || !work.illustrativeContent.containsKey("@id")) {
```
niklasl, line 21:
```suggestion
    } else if ( !(work.illustrativeContent instanceof Map) || !work.illustrativeContent.containsKey("@id")) {
```
because `!someMap == false`. ;) (This works anyway since if `work.illustrativeContent` ain't a List it's either a Map or *very* wrongly used (as in never) in practice.)

jannistsiroyannis: Nice catch! Damn parentheses! :)

jannistsiroyannis:
> Just a quick change to make it easier to identify in a sea of scripts to maybe name it "lxl-3048-illustrativecontent-cleanup" :)

Sure :)

PR #617, opened 2020-04-27
Add publictation country se -> sw clean up script.

jannistsiroyannis:


jannistsiroyannis:
> LGTM! Just a question: Would it be easier to check controlNumber on each record or is that not deemed as "safe"? :)

I see no safety problem with that but we don't have an index for the controlnumbers, so you'd need a sequential scan for each record (would take weeks to run). Or you could of course do where=true (everything) and check controlnumbers once per record in the script, but that would require every single libris record (every bib at least) being passed back and forth across the network to the script machine (also fairly slow).

PR #622, opened 2020-04-28
Feature/revert linkfield title

sarwik:
## Checklist
- [x] I have run integTest

## Description
For link fields (bib 760-78X): 
- Only one title should be exported to subfield t
- Only certain types are allowed to be exported to subfield t

The following should be exported (in order):
1. `hasTitle "@type": "KeyTitle"`
2. If no "KeyTitle", `hasTitle "@type": "Title"` should be exported
3. Other types of title should be ignored.

## Issue
[LXL-2308](https://jira.kb.se/browse/LXL-2308)

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -1446,7 +1446,7 @@
       "$t": {
         "about": "_:title",
         "property": "mainTitle",
-        "allowedRevertTypes": ["KeyTitle", "Title"],
+        "allowedTypesOnRevert": ["KeyTitle", "Title"],
         "TODO:targetMatch": "222, 245"
       },
       "$x": {"about": "_:instance", "addLink": "identifiedBy", "resourceType": "ISSN", "property": "value", "TODO:targetMatch": "022"},
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -16702,35 +16702,6 @@
               ]
             }
           }}
-        },
-        {
-          "source": {
-            "710": {"ind1": "2", "ind2": " ", "subfields": [
-              {"a": "Teologiska högskolan Stockholm."},
-              {"b": "Avdelningen för teologi."},
-              {"4": "org"}
-            ]}},
-          "result": {"mainEntity": {
-            "instanceOf": {
-              "@type": "Text",
-              "contribution" : [ {
-                "@type" : "Contribution",
-                "agent" : {
-                  "@type" : "Organization",
-                  "isPartOf" : {
-                    "@type" : "Organization",
-                    "name" : "Teologiska högskolan Stockholm."
-                  },
-                  "marc:subordinateUnit" : [ "Avdelningen för teologi." ]
-                },
-                "role" : [ {
-                  "@id" : "https://id.kb.se/relator/org",
-                  "@type" : "Role",
-                  "code" : "org"
-                } ]
-              } ]
-            }
-          }}
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


niklasl:
I'm curious whether the `revertForLink` and corresponding code (including `getWrappingEntity`) is actually necessary? I had imagined this to be solvable by extending the check controlled by `checkResourceType` around line 2836, to include the logic you've defined using `allowedRevertTypes`. The link (`hasTitle` in the existing cases) *should* already be specified by the mapping (unless they differ for some reason?), and the type checking logic could then be combined with the regular type check in the above mentioned place.

Other than that (due to possible redundant code for picking out the pieces and checking their types), I like this solution! (And I may have missed something about why a special `getWrappingEntity` is actually needed here.)

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -1443,7 +1443,13 @@
       "$n": {"about": "_:instance", "addLink": "hasNote", "resourceType": "Note", "property": "label",  "ignoreOnRevert": true},
       "$o": {"NOTE:subfield-count (Libris)": 0, "NOTE": "Cannot ignore this field since we import Mimer data with $o (We have lost $o data between 1.14.0 and 1.15.0). See LXL-1517"},
       "$s": {"about": "_:workTitle", "property": "mainTitle", "TODO:targetMatch": "240"},
-      "$t": {"about": "_:title", "property": "mainTitle", "TODO:targetMatch": "222, 245"},
+      "$t": {
+        "about": "_:title",
+        "property": "mainTitle",
+        "allowedRevertTypes": ["KeyTitle", "Title"],
```
niklasl, line 1449: For the line of thought in my lengthier comment, this would have to be moved to the declaration of `"_:title"` (at line 1415), where the link and type are specified.

niklasl: `allowedRevertTypes` might logically be better as `allowedSubTypesOnRevert` with just `KeyTitle` here, to omit the redeclaration of `Title` (which is also its base type). But that might be nitpicking over something we may not see used elsewhere  (or theoretically overly restricting if other cases do crop up -- although I suspect that logical restriction ought to apply in such cases too).

sarwik:
> I'm curious whether the `revertForLink` and corresponding code (including `getWrappingEntity`) is actually necessary? I had imagined this to be solvable by extending the check controlled by `checkResourceType` around line 2836, to include the logic you've defined using `allowedRevertTypes`. The link (`hasTitle` in the existing cases) _should_ already be specified by the mapping (unless they differ for some reason?), and the type checking logic could then be combined with the regular type check in the above mentioned place.
> 
> Other than that (due to possible redundant code for picking out the pieces and checking their types), I like this solution! (And I may have missed something about why a special `getWrappingEntity` is actually needed here.)

@niklasl

niklasl:
@sarwik I understand. Your description of `allowedTypesOnRevert` makes sense (about using "regular" Title over VariantTitle); let's go with that as you've delineated it. (It is a bit problematic logically, since a VariantTitle *is* a Title, but lets be pragmatic until we might manage to define a `RegularTitle` or `BaseTitle` and use that consistently everywhere (which we may never get to 😜 ).)

I think my suggestion was a bit too vague (and made by only skimming code without any hacking). I think there is a way to get at the candidate entities by measuring in the loop that goes down to each subfield, *but* hacking at the problem I see that the more important part is about putting the `allowedTypesOnRevert` on the pending `_:title` entity (per the specific code comment I made).

Having tested that, I'm gonna go ahead here and commit that to this feature branch. It seems to fix the remaining failing bib 776. Feel free to revert that in case I'm missing something crucial. (Note that this is specifically limited to use within a `pendingResource`, but I think it can be easily generalized to be allowed most everywhere we use `resourceType` today).

(It seems a spec for bib 773 is failing in this feature branch; that seems quite unrelated?)

sarwik:
@niklasl 
Ah smart to update `aboutMap` in `buildAboutMap`. I definitely didn't have that in mind. I merge this solution (approved by me ;) ).
(I have run integTest and no 773 spec fails for me)

PR #623, opened 2020-04-29
Add script for changing from @type NotatedMusicInstance.

jannistsiroyannis:


whelktool/scripts/cleanups/2020/04/lxl-1625-notatedMusicType.groovy
```diff
@@ -0,0 +1,49 @@
...
+            changed = true
+        }
+    }
+
+    if (!changed) {
+        instance["@type"] = "Instance"
+        changed = true
+    }
+
+    if (instance["hasType"] != null) {
```
sarwik, line 36: Two questions:
- Should `hasType` instead be `hasPart`? :)
- Do we know that `'{@graph,1,hasPart,@type}'` always exists in the sql query selection above? `(String where = "data#>>'{@graph,1,@type}' = 'NotatedMusicInstance' and data#>>'{@graph,2,@type}' = 'NotatedMusic'")`. We might need to adjust the sql query?

klngwll: 1. Yes it should be, sorry typo in my spec. 🙈 
2. Indeed you are right again, we can not presume to know in what order they exist.

jannistsiroyannis: I'm embarrased :) well spotted!

PR #625, opened 2020-05-05
Only demand all SAB classifications to imply fiction

lrosenstrom:


whelktool/scripts/cleanups/2019/11/marc-fiction-cleanup/fiction-cleanup.groovy
```diff
@@ -72,26 +71,27 @@ selectBySqlWhere(query, silent: false) { data ->
...
+boolean isSAB(classification) {
     def type = classification.'@type' as String
-    def code = classification?.code instanceof String ? [classification.code] : classification.code
 
     if (type == "Classification") {
         def inSchemeCode = classification.inScheme?.code as String
-        if (inSchemeCode && inSchemeCode == "kssb" && classification.inScheme?.'@type' == "ConceptScheme") {
-            return code && code.every { c -> c.startsWith("H") || c.startsWith("uH") }
-        }
+        return inSchemeCode && inSchemeCode == "kssb" && classification.inScheme?.'@type' == "ConceptScheme"
```
niklasl, line 85: It can be better to use `isInstanceOf(object, typeName)` instead of direct type value equality comparison (since that checks for subclasses as well).

lrosenstrom: Cool!

PR #627, opened 2020-05-06
Feature/ignore on revert fixed fields

sarwik:
## Checklist
- [x] I have run integTest
- [x] I have added/updated specs

## Description
1. Add functionality to be able to set ignoreOnRevert for certain positions in fixed fields
2. Ignore entire TextInstance 007 on revert
3. Change MovingImageInstance 007[16] to hasNote and ignore on revert
4. Set Electronic 007[10] to ignoreOnRevert
5. Ignore RemoteSensingImage 007 completely - both in import and export
6. Ignore KitInstance 007 completely - both in import and export
7. Ignore NotatedMusicInstance 007 completely - both in import and export

## Issue
[LXL-3114](https://jira.kb.se/browse/LXL-3114)

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
@@ -1335,6 +1335,7 @@ class MarcFixedFieldHandler {
                 columns << new Column(this, obj, colNum.first, colNum.second,
                         obj['itemPos'] ?: colNums.size() > 1 ? i : null,
                         obj['fixedDefault'],
+                        obj['ignoreOnRevert'],
                         obj['matchAsDefault'])
 
                 if (colNum.second > fieldSize) {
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
```diff
@@ -1402,18 +1403,21 @@ class MarcFixedFieldHandler {
         int end
         Integer itemPos
         String fixedDefault
+        Boolean ignoreOnRevert = null
         Pattern matchAsDefault
         MarcFixedFieldHandler fixedFieldHandler
 
         Column(MarcFixedFieldHandler fixedFieldHandler, fieldDfn, int start, int end,
-               itemPos, fixedDefault, matchAsDefault = null) {
+               itemPos, fixedDefault, ignoreOnRevert, matchAsDefault = null) {
```
lrosenstrom, line 1411: Perhaps initialize to null in the same way as 'matchAsDefault' to not mix conventions? If it needs to be initialized to null at all...

sarwik:
> LGTM, but I'm not sure i fully understand the implications of this.

I think the specs cover most of the implications, but it is also crucial to test the actual import and export behavior carefully - so we don´t miss anything crucial. I will merge this now so we have some time to test it before release 1.17.

PR #638, opened 2020-05-18
Add text-types cleanup script.

jannistsiroyannis:


whelktool/scripts/cleanups/2020/05/lxl-3139-text-to-BF2.groovy
```diff
@@ -0,0 +1,58 @@
...
+        instance["@type"] = "Manuscript"
+        changed = true
+    }
+
+    else if (instance["marc:mediaTerm"].contains("Kombinerat") || instance["marc:mediaTerm"].contains("kombinerat")) {
+        instance["@type"] = "Instance"
+        changed = true
+    }
+
+    else if (instance["hasPart"]) {
```
niklasl, line 39: This seems to assume that something with a printed text part makes the whole thing a printed text?

niklasl:
I had this in my WIP-script, might be useful:

```groovy

def deriveTypeFromMediaTerm(mediaTerm) {
    // TODO: map marc:mediaTerm to mediaType, carrierType or genreForm (sometimes of work)? return [type: X, carrier: Y, genreForm: Z]
    switch (mediaTerm) {
        case ~/(?i)ele[ck]tron.*.+|multimedia.*/:
        ELECTRONIC
        break

        case ~/(?i)\s*handskrift|\s*manus[ck]ript.*/:
        MANUSCRIPT
        break

        case ~/(?i).*tryck.*/:
        PRINT
        break

        case ~/(?i)\s*punktskrift.*/:
        TACTILE
        break

        case ~/(?i)\s*kombinerat material.*/: // work is Kit/MixedMaterial ?
        INSTANCE
        break
    }
}
```

whelktool/scripts/cleanups/2020/05/lxl-3139-text-to-BF2.groovy
jannistsiroyannis: Thus the "careful review point" :) The spec says:

in '{@graph,1,hasPart,@type}' =  'TextInstance'
     'Print' if 'carrierType' = https://id.kb.se/marc/RegularPrint
I guess its the "part" that should have type Print then?

niklasl:
I can put my entire (quite unfinished) script for comparison either in this PR, or elsewhere if you'd like? (It's a mix of statistics/research notes mostly converted to a whelktool script, at somewhere around "80% finished".)

jannistsiroyannis:
> I can put my entire (quite unfinished) script for comparison either in this PR, or elsewhere if you'd like? (It's a mix of statistics/research notes mostly converted to a whelktool script, at somewhere around "80% finished".)

That might be useful, but I will let Fredrik decide. He's the mastermind on these issues, i'm just the hired help :)

niklasl:
I here you. :) As I'm just a jack-in-the-box right now, I'll sync with him (my script was for "all of" LXL-1193 so I'll probably put it elsewhere to avoid scope creep).

niklasl:
I added the whole hairball here: https://github.com/libris/librisxl/blob/feature/normalize-instance-types/whelktool/scripts/2020/normalize-instance-types/script.groovy

klngwll:
> I added the whole hairball here: https://github.com/libris/librisxl/blob/feature/normalize-instance-types/whelktool/scripts/2020/normalize-instance-types/script.groovy

Nice, thanks!  I am gonna take a peak to see if there are some more cases we can drag along kicking and screaming :) We just split it up to be able to focus slightly on each script, however this will also come in handy for the Instance transformation later on. One thing we have changed though is we are not mapping Tactile at this point. We are not really sure it will be a viable subClass for Instance since a Tactile thing can also be a Print etc.

PR #655, opened 2020-06-03
Map marc:mediaTerm to instance and work types

olovy:


olovy:
Dry run on QA
```
   3144 Affisch
   2330 Musiktryck
    195 TV-spel
    813 Video
```

PR #658, opened 2020-06-10
Normalize new documents in the same way as updated

olovy:
Normalize when saving new documents. 
For example automatic imports and manual import from file.


olovy:
FYI

PR #659, opened 2020-06-10
Add script for relinking tgm terms

lrosenstrom:


whelktool/scripts/cleanups/2020/06/lxl-3187-tgm-relink-saogf-links.groovy
```diff
@@ -0,0 +1,59 @@
...
+    boolean updated = false
+
+    relations.each { relation ->
+        if (!instance[relation]) {
+            return
+        }
+
+        //Objects containing @type, prefLabel and sameAs
+        def toRemove = instance[relation].findAll{ hasSaogfSameAs(it) }
+        def toAdd  = toRemove.collect{ ['@id' : gmgpcPrefixSwe + it.sameAs[0][ID].tokenize('/').last()] }
```
sarwik, line 28: It may be good to actually verify that the newly created uri exists. PrefLabel is created manually and may contain errors, although the terms should be considered stable. But if not verify then maybe at least log which ones are created? Or test it thoroughly!

whelktool/scripts/cleanups/2020/06/lxl-3187-tgm-relink-saogf-links.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+        //Objects containing a single id key-value pair
+        def toRemoveSimple = instance[relation].findAll{ it[ID]?.contains(saogfPrefix)}
+        def toAddSimple = toRemoveSimple.collect { ['@id' : gmgpcPrefixSwe + it[ID].tokenize('/').last()] }
+
+        instance[relation].removeAll{ hasSaogfSameAs(it) }
+        instance[relation].removeAll{ it[ID]?.contains(saogfPrefix) }
+
+        //Add if linked object exists
+        (toAdd + toAddSimple).each {
```
sarwik, line 38: Nice! But it would also be good to make sure that the local entities remains if linking is not possible. Or log these occurrences to be corrected manually. Sorry for nitpicking!

whelktool/scripts/cleanups/2020/06/lxl-3187-tgm-relink-saogf-links.groovy
```diff
@@ -0,0 +1,78 @@
...
+            if (exists(it)) {
+                instance[relation].add(it[ID])
```
niklasl, line 40: Unless I miss something, this turns the input reference objects (`{"@id": "https://..."}`) into just strings (`"https://..."`)? (I'd expect the code to be just `it` (or `[(ID): it[ID]]` if the input isn't a reference but a full description).)

lrosenstrom: Good catch! This slipped in when adding the exists logic.

PR #666, opened 2020-06-15
Make records in Libris have the equivalent of 003 set to SE-LIBR.

jannistsiroyannis:
Previously, when we imported a MARC record from elsewhere, we left
the record's 003 (@graph,0,identifiedBy,N,{@type=SystemNumber}), and
only changed 003 when exporting the record. The practical consequence
of this is that whatever ID was in 001 when we imported the record, it
was discarded instead of placed in 035$a to facilitate future matching
in other databases.

This is now being corrected, so that pre-import IDs end up in 035$a
and 003 is being set immediately on import.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/util/LegacyIntegrationTools.groovy
+++ b/whelk-core/src/main/groovy/whelk/util/LegacyIntegrationTools.groovy
@@ -1,6 +1,13 @@
 package whelk.util
 
 import groovy.transform.CompileStatic
+import se.kb.libris.util.marc.Controlfield
+import se.kb.libris.util.marc.Field
+import se.kb.libris.util.marc.MarcRecord
+import se.kb.libris.util.marc.Subfield
+import se.kb.libris.util.marc.impl.ControlfieldImpl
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/whelk/util/LegacyIntegrationTools.groovy
```diff
@@ -105,4 +112,35 @@ class LegacyIntegrationTools {
...
+            f.getSubfields("a").each { sf ->
+                if (sf.getData().contains("(SE-LIBR)") || sf.getData().contains("(LIBRIS)"))
+                    hasLibris035aAlready = true
+            }
+        }
+
+        if (!hasLibris035aAlready && field001 != null && field003 != null && field001.getData()
+                != null && field003.getData() != null) {
+            Field field035 = new DatafieldImpl("035")
+            Subfield a = new SubfieldImpl("a".charAt(0), "(" + field003.getData() + ")" + field001.getData())
```
olovy, line 137: =)

PR #667, opened 2020-06-15
Dont return created on fail

lrosenstrom:
Ok, ugly solution that solves https://jira.kb.se/browse/LXL-2035, i.e. no longer return 201 - created when trying to add holding to a record that has been deleted. 
BUT isn't https://jira.kb.se/browse/LXL-2938  implying: "Don't fix LXL-2035 since it needs to be broken for the imports to work"? @jannistsiroyannis 
I guess the preferred solution would only involve one type of failure handling instead of either throwing, returning a boolean or returning null.


olovy:
I think it would be better to make `createDocument` return `void` and throw appropriate exceptions for different errors. This would require a bit more review of other call sites.

But this PR solves the immediate problem I guess.

lrosenstrom:
After discussion: will merge this and refactor later.

PR #671, opened 2020-06-16
Better logging and fix broken return in closure

lrosenstrom:


lrosenstrom:
> Fix title.
> 
> ship
Do you mean the name of the script?

PR #673, opened 2020-06-17
Crudspec: update unused storage mock

lrosenstrom:


lrosenstrom:
Either that or it needs some love. Merging this now to fix the build.

PR #675, opened 2020-06-30
Feature/combine qualifier title

sarwik:
Checklist:
- [x] I have run integTest

Description
Add functionality to merge properties on revert. 
This specific merge of qualifier and KeyTitle is needed by ISSN IC.  It´s the last step to enable catalogers on NB to link `otherPhysicalFormat` and get the Marc21 that ISSN IC needs.

See https://jira.kb.se/browse/LXL-3164 for more info.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
@@ -2882,6 +2882,8 @@ class MarcSubFieldHandler extends ConversionPart {
                 continue
             }
 
+            // TODO: Later, it may be desirable to add functionality
+            // to specify the punctuation mark to be used when merging.
             if (onRevertMergeWith) {
                 if (entity.containsKey(onRevertMergeWith)) {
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
```diff
@@ -2880,6 +2882,17 @@ class MarcSubFieldHandler extends ConversionPart {
                 continue
             }
 
+            // TODO: Later, it may be desirable to add functionality
+            // to specify the punctuation mark to be used when merging.
+            if (onRevertMergeWith) {
+                if (entity.containsKey(onRevertMergeWith)) {
+                    String valueToAppend = entity[onRevertMergeWith] instanceof List ?
```
niklasl, line 2889: Name-nitpicking: maybe call it something like `onRevertAppendValueFrom`?

sarwik: haha, I had actually named it to `onRevertAppendProperty` first which is almost the same as your suggestion. I'll change it to your suggestion :) Naming things is not the easiest! :P

PR #676, opened 2020-07-02
Add script for converting and moving bib249 to work

olovy:
Convert and move "original title" (obsolete marc field bib:249) from instance to work.


olovy:
We decided to first handle the simple case of on 249 and no existing work title.
The other rules might have to be tweaked.

PR #685, opened 2020-08-10
Add script for removing "narrower" links

olovy:


whelktool/scripts/2020/08/lxl-3213-remove-narrower.groovy
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+boolean hasBroader(String narrower, String broader) {
+    return (getThing(narrower)['broader'] ?: []).collect{ it['@id'] }.contains(broader)
+}
+
+boolean fixBroader(String narrower, String broader) {
+    boolean ok = false
+    selectByIds([narrower]) { auth ->
```
sarwik, line 71: Why not use getThing() here?

olovy: Because you have to be inside the closure passed to `selectByIds` to be able to save the record (`scheduleSave`).
I added `fixBroader` later and didn't care to refactor the script.

PR #688, opened 2020-08-12
Fix subfieldOrder in 411 (like 111)

klngwll:
- [x] I have run integTest

* Align subfieldOrder in 411 with 111.

See [LXL-2661](https://jira.kb.se/browse/LXL-2661)


sarwik (approved):
Just double checking.. It's the same subfieldOrder as **bib** 111 - is that correct? Or should it be **auth** 111? Otherwise, ship it!

PR #700, opened 2020-08-28
Apply inverses on revert

niklasl:
This is for LXL-3309. The gist of it is:

In MarcFrameConverter, if post-processing is turned on, go through the reverse map, look up key term in vocabIndex, and if the property has an owl:inverseOf, find the first value of that whose term is in vocabIndex, and add that property term to the thing as a regular property with all the incoming subjects as objects.

I've added a unit test, but this needs practical testing and considerations. For this to work at all, we need to ensure that the embellished data sent to revert also contains `@reverse` (and if it doesn't already, ensure that it isn't too costly to enable).

See LXL-3309 for more details and discussions about domain (e.g. to utilize this for removing redundant `hasReproduction` if those are generated from incoming `reproductionOf` links).

### DIFF
--- a/whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
+++ b/whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
@@ -392,6 +392,30 @@ class MarcFrameConverterSpec extends Specification {
         resultWithoutDollars != null
     }
 
+    def "should apply inverses"() {
+        Closure applyInverses = converter.conversion.&applyInverses
+        given:
+        def record = ['@id': null, controlNumber: "123"]
+        def thing = [
+            '@id': '1',
+            '@reverse': [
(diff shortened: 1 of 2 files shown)


jannistsiroyannis:
> For this to work at all, we need to ensure that the embellished data sent to revert also contains `@reverse` (and if it doesn't already, ensure that it isn't too costly to enable).

Attans.. _igår_ gjorde vi ändringar för få upp exporthastigheten (vilken förvärrats sen vi började med transitiv embellish, och nu orsakat en incident), som baserades på antagandet att _just det här_ inte behövdes - att reverteringen bara använder sig av frammåtlänkar. :(

Det vore olyckligt att behöva backa det nu, eller att försöka göra en väldigt mycket mer komplicerad cacheinvalidering för att även klara det här fallet.

olovy (approved):
LGTM!

export embellish cache has been changed to include reverse links

PR #703, opened 2020-09-01
When doing bulkStore() don't bother clearing the cache for each recor…

jannistsiroyannis:
…d, evict the whole thing after instead.

This speed definitions reloads back up again (about 130 sec on qa)


lrosenstrom (approved):
LGTM

PR #713, opened 2020-09-07
Fix DB connection deadlock in embellish cache

olovy:
loadEmbellished tries to grab two database connections leading to deadlock.


lrosenstrom (approved):
Looks good

PR #715, opened 2020-09-09
Use 'full' lens when calculating reverse links for doc

olovy:
This includes two fixes:
- Display lenses without `@id` didn't get any inherited properties
- Embellisher used lens of first embellish level instead of `full` when adding `@reverse` for starting document


klngwll (approved):
Approve the general idea 👍

sarwik (approved):
LGTM and ITY :) Ship it!

PR #716, opened 2020-09-09
Feature/languageNote

sarwik:
- [x] I have run integTest
- [x] I have dry-runned script

- Delete all occurrences of marc:languageNote
- On revert set `ind1=1` if `translationOf` exists
- Add subfieldOrder to bib 041

[LXL-3253](https://jira.kb.se/browse/LXL-3253)


whelktool/scripts/2020/09/lxl-3253-remove-languageNote.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+selectByCollection('bib') { bib ->
+    Map work = getWork(bib)
+    boolean changed = false
+
+    if (!work) {
+        return
+    }
+
+    if (work.remove('marc:languageNote')) {
```
klngwll, line 18: From what I can surmise from discussion in Issue the script does not need to take into account if translationOf is present?

klngwll (approved):
As far as I can see (In the very complex machinations of the languages) this works as expected and the specs tell the tale, checked out and built locally here as well. The script seem to be well discussed and yes there might be a possibility of missing some data in imports by dropping the indicator but then it was probably not very good to begin with. Good work! 🌟

whelktool/scripts/2020/09/lxl-3253-remove-languageNote.groovy
sarwik: Exactly. MSS has analyzed the data based on log files. They have come to the conclusion that it should be ok to remove all occurrences of `marc:languageNote`. In most cases where `marc:languageNote` is missing, it will be better and in some cases the data is incorrectly catalogued. But it will at least be consistent.

PR #719, opened 2020-09-10
Make export faster for auth (and bib)

olovy:
Only load and embellish documents that should actually be exported for the current profile.


lrosenstrom (approved):
Nice :ship:

PR #720, opened 2020-09-11
Move bib976 to Instance

klngwll:
- [x] I have runt integTest

If [LXL-3304](https://jira.kb.se/browse/LXL-3304) moves bib976 to Instance, this needs to be merged.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -20484,14 +20484,39 @@
 
     "976": {
       "Note:local": "Librisdefinierat fält. Fältet är avsett för maskinell uppdatering och förekommer normalt endast i äldre poster konverterade från ett ursprungligt LIBRISIII-format, respektive i de MTM-poster som maskinellt har lästs in från Bibliotekstjänst.",
-      "aboutEntity": "?work",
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -20484,14 +20484,39 @@
...
       "i2": {"property": "marc:bib976-i2", "marcDefault": " "},
       "$a": {"property": "marc:bib976-a"},
-      "$b": {"property": "marc:bib976-b"}
+      "$b": {"property": "marc:bib976-b"},
+      "_spec": [
+        {
+          "source": {
+            "976": {
+              "subfields": [
+                {"a": "a"},
```
lrosenstrom, line 20499: The only 967's that are moved to the instance are those that are missing 976$a but maybe that's irrelevant here.

lrosenstrom (approved):
LGTM!

whelk-core/src/main/resources/ext/marcframe.json
klngwll: If the others are becoming Classification objects yes :)

sarwik (approved):
LGTM!

PR #722, opened 2020-09-11
API: Add _applyInverseOf parameter to CRUD GET

olovy:
Add `_applyInverseOf` parameter to CRUD GET
When `_applyInverseOf=true`, reverse relations in the document get evaluated to their inverse relation.
e.g. `@reverse/broader` becomes `narrower`

This is so that we can display them on id.kb.se, see https://github.com/libris/lxlviewer/pull/688

We don't want docs PUT back in in their expanded form by mistake. So this is a "private" parameter only for internal use (and filtered by the python backend).


lrosenstrom (approved):
LGTM! Perhaps use 'evaluate' rather than 'expand' to use a less overloaded term?

olovy:
> LGTM! Perhaps use 'evaluate' rather than 'expand' to use a less overloaded term?

Agreed,
`_evaluateReverse`
`_evaluateInverse`
`_applyInverse`
?

PR #725, opened 2020-09-11
Search: Add or- parameter prefix

olovy:
TODO: tests
TODO: update API docs


lrosenstrom (approved):
Looks good technically. I guess we can revisit the discussion about implementation alternatives raised in  https://jira.kb.se/browse/LXL-3305 later.

PR #727, opened 2020-09-15
DB: Add index idx_lddb_controlnumbers

jannistsiroyannis:
…-list), since pgsql does not correctly optimize those queries.


olovy (approved):
Tested. Adding index is sufficient for speeding upp gui-whelktool.
14K bibs took less than one minute.

PR #728, opened 2020-09-15
Don´t revert auth 042 $9 when inCollection is uniformWorkTitle

sarwik:
## Checklist:

- [x] I have run integTests
- [ ] Script to add `inCollection` `uniformWorkTitle` to auth-Works (from Marc21)

## Description:

With https://github.com/libris/definitions/pull/235 we add temporary collection for old MARC works - but we do not want them to be converted to auth `042 $9 uniformWorkTitle `

## Issues:

[LXL-3357](https://jira.kb.se/browse/LXL-3357)

## PR involved:

https://github.com/libris/definitions/pull/235

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -21108,6 +21108,14 @@
       ]
     },
     "042": {
+      "match": [{
+          "when": {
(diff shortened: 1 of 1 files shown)


klngwll (approved):
LGTM, What a tricky one! ⭐

PR #729, opened 2020-09-15
Fix: specific value for property should replace wildcard in facet links

olovy:
Example: After a search with `@type=*`, generated facet links for `@type` should not contain wildcard

wrong:
`@id": "/find?q=*&@type=*&_limit=20&@type=Electronic`

corrrect:
`@id": "/find?q=*&_limit=20&@type=Electronic`


klngwll (approved):
🚢

PR #732, opened 2020-09-16
Fix dependency cache invalidation

olovy:
- Invalidate when creating new document
- Fix potential race where cache might be re-populated with old data because `invalidate` is called before commit of updated document.


sarwik (approved):
LGTM!

PR #733, opened 2020-09-17
Add script for making all languages into lists.

jannistsiroyannis:


klngwll (approved):
Seems to work as intended, this is probably reusable on a few other places I would guess 👍

PR #736, opened 2020-09-18
Feature/bib976 fixes

lrosenstrom:
Fixes kssb version being set to 2. After discussion with Harriet Aagaard, we decided to set kssb version = 6 för all bib976 objects.


olovy (approved):
:ship:

PR #738, opened 2020-09-22
Add a script that replaces all links to records with links to the cor…

jannistsiroyannis:
…responding mainEntities.

### DIFF
--- a/whelktool/scripts/cleanups/2020/09/lxl-3152-linkOnlyMainEntity.groovy
+++ b/whelktool/scripts/cleanups/2020/09/lxl-3152-linkOnlyMainEntity.groovy
@@ -28,7 +28,7 @@ boolean traverse(Object node, String idForLogging, Whelk whelk, PrintWriter link
                 map["@id"].startsWith(baseUri.toString()) &&
                 !map["@id"].endsWith("#it") ) {
             String correctLinkTarget = whelk.storage.getThingId(map["@id"])
-            if (correctLinkTarget != null) {
+            if (correctLinkTarget != null && correctLinkTarget != map["@id"]) {
                 linkChanges.println("In ${idForLogging}, changed a reference ${map["@id"]} , to ${correctLinkTarget}")
                 map["@id"] = correctLinkTarget
                 changed = true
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2020/09/lxl-3152-linkOnlyMainEntity.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+boolean traverse(Object node, String idForLogging, Whelk whelk, PrintWriter linkChanges) {
+    boolean changed = false
+
+    if (node instanceof Map) {
+        Map map = node
+
+        // Correct references are all id.kb.se or ends with #it.
```
olovy, line 26: Code is correct but comment is wrong, could also be an external link (e.g. http://id.loc.gov/authorities/label/Wolves)

olovy (approved):
LGTM

olovy (changes requested):
Should not change library and script ids
e.g.
https://libris.kb.se/library/NB
https://libris.kb.se/sys/globalchanges/cleanups/2020/01/lxl-2494/repair-uppercase-term-links.groovy

anything else?

whelktool/scripts/cleanups/2020/09/lxl-3152-linkOnlyMainEntity.groovy
jannistsiroyannis: :+1:

jannistsiroyannis:
> Should not change library and script ids
> e.g.
> https://libris.kb.se/library/NB
> https://libris.kb.se/sys/globalchanges/cleanups/2020/01/lxl-2494/repair-uppercase-term-links.groovy
> 
> anything else?

Good catch, I missed the https://libris.kb.se/library/NB ones. Fixed now. Can't think of anything else!

klngwll (changes requested):
Some other keys, especially so on record, is sensitive in this regards (derivedFrom etc.) I need to dive into this a bit more, but on mainEntity the sameAs is also "special". It is a bit scary with such a broad sweep hehe. I had hoped the problem was isolated to genreForm and perhaps subject.

jannistsiroyannis:
> Some other keys, especially so on record, is sensitive in this regards (derivedFrom etc.) I need to dive into this a bit more, but on mainEntity the sameAs is also "special". It is a bit scary with such a broad sweep hehe. I had hoped the problem was isolated to genreForm and perhaps subject.

It should be noted however, that for any link to a record, where there does not exist any mainEntity (with an id), the links will be left alone. So for example, the heldBy and generationProcess that were given special treatment does not actually need it - they have no mainEntity to link to, so nothing will happen anyway.

klngwll (approved):
LGTM! 🎈

PR #739, opened 2020-09-22
When a record changes, make sure any new URIs do not collide with exi…

jannistsiroyannis:
…sting URIs

Now that it is possible to change a records URI, we must make sure that
the changed URI does not conflict with an existing one.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -10,6 +10,7 @@ import whelk.component.ElasticSearch
 import whelk.component.PostgreSQLComponent
 import whelk.converter.marc.MarcFrameConverter
 import whelk.exception.StorageCreateFailedException
+import whelk.exception.WhelkRuntimeException
 import whelk.filter.LinkFinder
 import whelk.filter.NormalizerChain
 import whelk.search.ESQuery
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -352,6 +353,13 @@ class Whelk {
             preUpdateDoc = doc.clone()
             updateAgent.update(doc)
             normalize(doc)
+
+            boolean detectCollisionsOnTypedIDs = false
+            List<Tuple2<String, String>> collidingIDs = getIdCollisions(doc, detectCollisionsOnTypedIDs)
+            if (!collidingIDs.isEmpty()) {
+                log.info("Refused update of " + doc.getShortId() + ". Document would have ID collision with : " + collidingIDs)
+                throw new WhelkRuntimeException("Update of ${doc.getShortId()} failed, due to ID collision with : " + collidingIDs)
+            }
```
olovy, line 362: Maybe extract this to a method `checkIdCollsions(doc)` and use it in all three places?

jannistsiroyannis: But the error messages are slighlty different, depending on if its CREATE or UPDATE.

olovy: Do they have to be?
"`StorageCreateFailedException`" is not referenced in any other place.

jannistsiroyannis: It's referenced here: rest/src/main/groovy/whelk/rest/api/Crud.groovy:729, apparently so we can return a http 409: Conflict.
I don't mean to say that's necessarily a meaningful thing to do, but I'd rather not change the crud interface if I don't have to.

jannistsiroyannis: HAha, det behövs inte! PostgreSQL.verifyDocumentIdRetention() sköter redan det här. Förmodligen upkom den här i ett tidigare skede (den första stora inläsningen)!

PR #743, opened 2020-09-29
Feature/lxl 3149

jannistsiroyannis:


gui-whelktool/src/main/resources/templates/change-complex-subject.groovy
```diff
@@ -0,0 +1,68 @@
...
+                if (subject.termComponentList[0]["@type"] == "Topic"
+                        && subject.termComponentList[0]["prefLabel"] == "FROMTERM") {
+                    subject.termComponentList[0]["prefLabel"] = "TOTERM"
+                    fixedTermComponentList = true
+                }
+            }
+        }
+    }
+
+    return fixedSameAs && fixedPrefLabel && fixedTermComponentList
```
olovy, line 67: If some but not all of these are true, the subject will be partially updated or not depending on if another subject in the document is updated. Is this the desired behavior?

jannistsiroyannis: The record will only be saved if all three are true. If there's any thing "strange" going on, such that one of them is not, i don't think we should touch the record.

olovy: Ok, I wasn't clear.

If they are all true for one subject, all other "partially" updated subjects in the same document will also be saved since the data has been modified. Probably small risk in practice but possible.

jannistsiroyannis: I'll make some changes, so that any records where only some of these are true are logged separately, for manual review. Sound fair?

gui-whelktool/src/main/resources/templates/change-complex-subject.groovy
```diff
@@ -0,0 +1,77 @@
...
+
+        if (subject.prefLabel != null && subject.prefLabel instanceof String && subject.prefLabel.startsWith("FROMTERM")) {
+            subject.prefLabel = subject.prefLabel.replaceFirst("FROMTERM", "TOTERM")
+            fixedPrefLabel = true
+        }
+
+        if (subject.termComponentList != null && subject.termComponentList instanceof List) {
+            if (subject.termComponentList.size() > 0) {
+                if (subject.termComponentList[0]["@type"] == "Topic"
+                        && subject.termComponentList[0]["prefLabel"] == "FROMTERM") {
```
sarwik, line 60: Instead of changing the prefLabel of the local entity of the termComponentList it would be better to replace the local entity with a link to the actual subject ("TOTERM"?). We want to get rid of as many local entities as possible in ComplexSubjects.

gui-whelktool/src/main/resources/templates/change-complex-subject.groovy
```diff
@@ -0,0 +1,77 @@
...
+})
+
+boolean fixSubject(Object subject, PrintWriter manualReviewLog, String id) {
+    boolean fixedSameAs = false
+    boolean fixedPrefLabel = false
+    boolean fixedTermComponentList = false
+
+    if (subject["@type"] == "ComplexSubject") {
+
+        if (subject.sameAs != null && subject.sameAs instanceof List) {
```
sarwik, line 41: Could we have any sameAs that is a Map?

jannistsiroyannis: > Could we have any sameAs that is a Map?

Om inget ändrats radikalt, så ska det vara omöjligt (har varit omöjligt tidigare i alla fall).

gui-whelktool/src/main/resources/templates/change-complex-subject.groovy
jannistsiroyannis: I really think this needs to be a separate thing we do. The task here is only to update the term itself.

sarwik: ok, fair 👍

PR #744, opened 2020-10-07
When changing the main terms complex subject, affect only SAO-terms.

jannistsiroyannis:
Other terms are managed by other organizations and should not be changed.


olovy (commented):
LGTM

:ship:

PR #745, opened 2020-10-07
Avoid creating double links when replacing records.

jannistsiroyannis:
Sometimes when a record A is replaced by a record B, the situation occurs
that another record C was already linking to both A and B. When this happens
it should only link once to A after the operation, not twice.


olovy (approved):
:ship:

PR #746, opened 2020-10-08
Do not fail to remove recently changed records from Elastic.

jannistsiroyannis:
When a record has (very) recently been updated, it cannot be deleted
until Elastic has decided that it (internally) is done with it, even though
it already answered that it was earlier. When this happens you get a
conflicting version-error. When such an error is registered, we should
retry - until the delete is successful.


olovy:
This has do do with elastic refresh intervals[1]. We don't want to sync every request to the refreshes for performance reasons. This case of a refresh followed by a delete is split over different transactions so we can't know that we have to sync around a refresh.

The delete by id query is safe to retry so I think this is a good solution.

:ship: 

[1] https://discuss.elastic.co/t/elasticsearch-delete-by-query-409-version-conflict/174150

olovy (approved):

PR #749, opened 2020-10-13
Feature/shapes script

klngwll:
Update shapes script for work which break after moving all local work descriptions from graph2 into thing.instanceOf

Assertion of linked things is a pending question I have about the shapes, hence left the comment, but maybe save it for later? the script doesn't break now ;)

### DIFF
--- a/librisxl-tools/scripts/lddb_json_shape.py
+++ b/librisxl-tools/scripts/lddb_json_shape.py
@@ -19,12 +19,12 @@ def reshape(data):
         thing =graph[1]
         thing['meta'] = graph[0]
 
-        if 'instanceOf' in thing:
-            work = thing['instanceOf']
-            #assert thing['instanceOf']['@id'] == work['@id']
-            thing['instanceOf'] = work
-        else:
-            work = None
+        work = thing.get('instanceOf')
+        if work and len(work) == 1 and '@id' in work:
+            for item in graph[2:]:
+                if item.get('@id') == work['@id']:
+                    work = thing['instanceOf'] = item
+                    break
 
         return thing, work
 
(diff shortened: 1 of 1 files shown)


niklasl:
It's odd that the script failed, since it "should" only pick out the work from position 2 if it's there (an untested off-by-one maybe). Also, the change does a redunant thing now (it sets the instanceOf from itself). I can take a look if you want, tonight or tomorrow.

librisxl-tools/scripts/lddb_json_shape.py
```diff
@@ -4,26 +4,9 @@
 MAX_STATS = 512
 HARD_MAX_STATS = 8192
 STATS_FOR_ALL = {
-        # from auth 008
```
niklasl, line 7: Did these break anything or were they removed since they are already cleaned up from lddb?

klngwll: Removed because cleaned up.

niklasl:
The scripts copes fine until it blow up on https://libris.kb.se/sb4lr8645649p1c/data.jsonld?embellished=false which has a trail of ConceptScheme bnodes for some reason... That might be an issue. Anyway, I can adjust this script fix to handle both shapes better.

niklasl:
First added a simpler equivalent to the change (`work = thing.get('instanceOf')`), then re-added a backwards-compatible lookup of a work reference in the graph (tested with an old dump). That's more robust also in case we go for a flattened form sometime in the future.

klngwll:
> First added a simpler equivalent to the change (`work = thing.get('instanceOf')`), then re-added a backwards-compatible lookup of a work reference in the graph (tested with an old dump). That's more robust also in case we go for a flattened form sometime in the future.

Excellent, I can put this to work with a fresh dump later tonight.

PR #765, opened 2020-10-26
Add script for removing '375 and '377' from auth data

kwahlin:


whelktool/scripts/cleanups/2020/10/lxl-3391-remove-unhandled-375-and-377-from-auth.groovy
```diff
@@ -0,0 +1,38 @@
+PrintWriter scheduledForUpdating = getReportWriter("scheduled-updates")
+PrintWriter failedUpdating = getReportWriter("failed-updates")
+
+String where = "collection = 'auth' and deleted = false and data#>>'{@graph,0,_marcUncompleted}' ~ '\"375\"|\"377\"'"
```
andersju, line 4: This query might also fetch things that you don't necessarily want to process, like (contrived example):
```
      "_marcUncompleted": [
        {
          "700": {
            "ind1": "1",
            "ind2": "0",
            "subfields": [
              {
                "a": "375"
              },
```
This is in itself not a problem, but (see following comment)...

whelktool/scripts/cleanups/2020/10/lxl-3391-remove-unhandled-375-and-377-from-auth.groovy
```diff
@@ -0,0 +1,38 @@
...
+
+selectBySqlWhere(where) { auth ->
+    // Make '_marcUncompleted' a list
+    def marcUncompleted = asList(auth.graph[0]['_marcUncompleted'])
+
```
andersju, line 9: ...here you might want to add something like this:
```suggestion

    if (!isRemovable(marcUncompleted))
        return

```
Otherwise, things that just accidentally happen to have "375" or "377" as values somewhere in _marcUncompleted will be updated needlessly. This might be overly paranoid and not an issue in practice, but it wouldn't hurt :)

andersju (commented):
Looking good, just a minor (possible) issue! :+1:

kwahlin:
Good point @andersju ! Fixed in 613e489

andersju (approved):
:ship: :balloon:

PR #772, opened 2020-10-28
Add script for changing subject inScheme hsv -> ssif

olovy:
~700K subjects


jannistsiroyannis (approved):
I don't really follow what the statistics part is about. I can't yet see why you need it. Other than that it looks fine.

PR #774, opened 2020-10-29
Add script for linking KBslagord -> SAO 

olovy:
Statistics stuff depends on cb82535


olovy:
Need to check if KBslagord subject is ComplexSubject

jannistsiroyannis (approved):
I could find no fault at cursory glance!

PR #778, opened 2020-10-30
Feature/lxl 3214 add document validation

lrosenstrom:
TODO: add integration tests


olovy (commented):
Nice!

1) Have you run it on all docs in PROD?
2) We will need a script to fix the form of all repeatable terms? See 1)
  * Cat-client is ok for now since it normalizes them?
3) If we do 1) and 2) we could also do this at e.g. the same point as PostgreSQLComponent::normalizeDocumentForStorage

PR #785, opened 2020-11-04
Feature/validation analysis

lrosenstrom:


niklasl:
Might utilize or supersede https://github.com/libris/librisxl/blob/develop/whelk-core/src/main/groovy/whelk/JsonLd.groovy#L464

whelk-core/src/main/groovy/whelk/JsonValidator.groovy
```diff
@@ -38,16 +40,29 @@ class JsonValidator {
     }
 
     Set validateAll(Map map) {
```
olovy, line 42: This could be `List<String>` which would return errors in "logical" order (traversal order).

lrosenstrom:
> Might utilize or supersede https://github.com/libris/librisxl/blob/develop/whelk-core/src/main/groovy/whelk/JsonLd.groovy#L464

The validator (JsonValidator.groovy) used is essentially that code extracted to its own class, with some apis and added flexibility!

niklasl:
> > Might utilize or supersede https://github.com/libris/librisxl/blob/develop/whelk-core/src/main/groovy/whelk/JsonLd.groovy#L464
> 
> The validator (JsonValidator.groovy) used is essentially that code extracted to its own class, with some apis and added flexibility!

Great! I now see it already supersedes it (just a throw-away comment from my phone). (JsonValidator might be called JsonLdValidator? JsonValidator would be more apt if we used JSON Schema...)

olovy (approved):
:ship:

PR #789, opened 2020-11-06
Make sure holdings are created only on bib records, and nothing else.

jannistsiroyannis:
It is the case that a "controlNumber" is not in fact unique. It can be
shared by up to 3 record, one from "bib", "hold" and "auth" respectively.

This fixes a bug where this was not taken into account, and holdings were
created on non-bib-record. This has happened, which started a chain of
events leading to corrupt records (recoverable) in websök.


olovy (approved):
:ship:

PR #791, opened 2020-11-11
Add the option of limiting sigel changes to a list of bib ids.

jannistsiroyannis:
Additionally add an extra data sanity check to the delete-hold script.


olovy (approved):
:ship:

PR #793, opened 2020-11-11
Add script for fixing/removing broken holds

andersju:


whelktool/scripts/cleanups/2020/11/lxl-2791-fix-broken-holds/lxl-2791-fix-broken-holds.groovy
```diff
@@ -0,0 +1,61 @@
...
+// Turn file with XL id - old hold id - old bib id into a map, keyed by XL id
+new File(scriptDir, "id-oldid-bibid.tsv").each {
+    def (id, oldId, bibId) = it.split(/\t/).collect { it.trim() }
+    holds[id] = ["oldId": oldId, "bibId": bibId]
+}
+
+// Make sure the referenced bib records still exist. Fetch them using the legacy IDs
+// we got from Voyager, and map them to their XL shortIds.
+selectByIds(holds.collect { "http://libris.kb.se/resource/bib/" + it.value.bibId }) { data ->
+    if (data.graph[1]["@type"] == "Instance" && data.graph[0].controlNumber)
```
jannistsiroyannis, line 24: Are we sure that they're all "Instance" ? No "Electronic", "Print" or anything else?

JsonLd.getSubClasses() is a tip, in case you want "any" of these! :)

whelktool/scripts/cleanups/2020/11/lxl-2791-fix-broken-holds/lxl-2791-fix-broken-holds.groovy
```diff
@@ -0,0 +1,61 @@
...
+    String legacyBibId = holds[data.doc.shortId].bibId
+
+    // Broken holds from certain sigels, or ones that link to a nonexistent bib record,
+    // should shuffle off this mortal coil
+    if (data.doc.sigel in ["Hal", "SEK"]) {
+        shouldBeDeleted = true
+    } else if (legacyBibId in validBibs) {
+        if (!(instance.itemOf instanceof Map))
+            instance.itemOf = [:]
+        instance.itemOf << ["@id": baseUri.resolve(validBibs[legacyBibId]).toString()]
```
jannistsiroyannis, line 41: The tricky part here, is that if you resolve a short id, you get the "record" ID, for example:
https://libris.kb.se/aaaaaaa
But we actually (always) want to link to the mainEntity uri (@graph,1,@id), which would (normally) look something like:
https://libris.kb.se/aaaaaaa#it

jannistsiroyannis (changes requested):

whelktool/scripts/cleanups/2020/11/lxl-2791-fix-broken-holds/lxl-2791-fix-broken-holds.groovy
andersju: We're not! I removed it - checking seems pointless anyway, since I assume http://libris.kb.se/resource/bib/<id> should always resolve to either a bibliographic record or nothing, and we don't care what type it then happens to be.

whelktool/scripts/cleanups/2020/11/lxl-2791-fix-broken-holds/lxl-2791-fix-broken-holds.groovy
andersju: :+1:

jannistsiroyannis (approved):
:ship:

PR #794, opened 2020-11-12
Make multi-wildcard searches possible.

jannistsiroyannis:
For example "*kol*" should match "skolor".
To do this, we're switching all instances of simple query strings, to
query strings (ES nomenclature).

Side-effects of this are:
1. Queries can now fail (must be handled)
2. Character escaping becomes necessary in implict filters.
3. Character escaping will also be necessary in the "search box" if such
   chars are ever used there.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -496,8 +496,6 @@ class ESQuery {
         if (queryString.findAll('\\"').size() % 2 != 0) {
             throw new whelk.exception.InvalidQueryException("Unbalanced quotation marks")
         }
-
-        // The following chars are ES operators and need to be escaped to be used as literals: \+-=|&><!(){}[]^"~*?:/
     }
 
     /**
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -78,7 +78,7 @@ class ESQuery {
     @CompileStatic(TypeCheckingMode.SKIP)
     Map getESQuery(Map<String, String[]> queryParameters) {
         // Legit params and their uses:
-        //   q - query_string
+        //   q - query string, will be used as query_string or simple_query_string
         String q
         //   _limit, _offset - pagination
         int limit
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -78,7 +78,7 @@ class ESQuery {
     @CompileStatic(TypeCheckingMode.SKIP)
     Map getESQuery(Map<String, String[]> queryParameters) {
         // Legit params and their uses:
-        //   q - simple_query_string
+        //   q - query_string
         String q
         //   _limit, _offset - pagination
         int limit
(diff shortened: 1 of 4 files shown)


niklasl:
As I recall, we switched *to* `simple_query_string` for good reasons, being simpler (thus exposing a simpler logic in the API, something which ES also recommends "for search boxes").

If there is no other way, I'd probably rather switch to full `query_string` *only* if the search matches something like `/\*\S+/` (or a more involved regex to control exactly what we expose), instead of attempting to escape, clean up and error handle the full query string DSL syntax (which is too much for our regular use cases).While non-obvious to the end-user, we might explain that we (experimentally) allow leading wildcard queries (unless they prove too expensive).

olovy (changes requested):
LGTM except the escaping part.

It should be possible to use the operators described in [the search help](https://libris-stg.kb.se/katalogisering/help/search) when on e.g. prefLabel.
e.g. http://localhost:8080/katalogisering/search/libris?prefLabel=*kolor&%40type=Concept

We need to handle filtering on URIs as well so maybe something like
```
if (looksLikeUri(val))
    escape(val)
```
or
`ESreservedChars = "=><!{}[]^:/"` not enough?


I tried modifying `ESreservedChars` but this query still finds nothing when using `query_string` https://libris.kb.se/katalogisering/search/libris?prefLabel=%28tekniska%20%7C%20frist%C3%A5ende%29%20skolor&%40type=Concept
maybe we have to live with that.

olovy:
> LGTM except the escaping part.
> 
> It should be possible to use the operators described in [the search help](https://libris-stg.kb.se/katalogisering/help/search) when on e.g. prefLabel.
> e.g. http://localhost:8080/katalogisering/search/libris?prefLabel=*kolor&%40type=Concept
> 
> We need to handle filtering on URIs as well so maybe something like
> 
> ```
> if (looksLikeUri(val))
>     escape(val)
> ```
> 
> or
> `ESreservedChars = "=><!{}[]^:/"` not enough?
> 
> I tried modifying `ESreservedChars` but this query still finds nothing when using `query_string` https://libris.kb.se/katalogisering/search/libris?prefLabel=%28tekniska%20%7C%20frist%C3%A5ende%29%20skolor&%40type=Concept
> maybe we have to live with that.

This is of course obsoleted by the comment by Niklas

olovy:
I started doing something more fancy that could handle queries like `*foo | http://bar`, i.e. escaping characters in URIs when using `query_string` but I don't really see the need for it?

whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
```diff
@@ -477,6 +483,23 @@ class ESQuery {
...
+        def containsLeadingWildcard = queryString =~ /\*\S+/
+        return !containsLeadingWildcard
+    }
+
+    private static void checkNonSimpleQueryString(String queryString) {
+        if (queryString.findAll('\\"').size() % 2 != 0) {
+            throw new whelk.exception.InvalidQueryException("Unbalanced quotation marks")
+        }
+
+        // The following chars are ES operators and need to be escaped to be used as literals: \+-=|&><!(){}[]^"~*?:/
```
jannistsiroyannis, line 500: Remove comment if no longer doing this?

olovy: :+1:

olovy (approved):

PR #795, opened 2020-11-13
Activate jsonld validation: checks for nested graph

lrosenstrom:


rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -92,6 +96,7 @@ class Crud extends HttpServlet {
         vocabData = whelk.vocabData
         jsonld = whelk.jsonld
         search = new SearchUtils(whelk)
+        validator = JsonLdValidator.from(jsonld)
```
olovy, line 99: Need to stop processing if validation fails, i.e. return.
I think it's cleaner to throw and write error response at top level.

rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -499,8 +504,8 @@ class Crud extends HttpServlet {
         newDoc.normalizeUnicode()
         newDoc.trimStrings()
         newDoc.deepReplaceId(Document.BASE_URI.toString() + IdGenerator.generate())
-        // TODO https://jira.kb.se/browse/LXL-1263
         newDoc.setControlNumber(newDoc.getShortId())
+        validateJsonLd(newDoc, request, response)
```
olovy, line 508: Same as above

rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -548,6 +553,17 @@ class Crud extends HttpServlet {
...
     }
 
+    private void validateJsonLd(Document doc, HttpServletRequest request, HttpServletResponse response) {
+        List<JsonLdValidator.Error> errors = validator.validate(doc.data)
+        if (errors) {
+            def message = errors.collect { it.toStringWithPath() }
+            failedRequests.labels(request.getMethod(), request.getRequestURI(),
+                    HttpServletResponse.SC_BAD_REQUEST.toString()).inc()
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST,
+                    "Invalid JsonLd. Got errors:" + message)
```
olovy, line 563: Maybe return a structured response, i.e. JSON?
(inconsistent with rest of API though)

olovy (changes requested):
Need to stop processing if validation fails.

rest/src/main/groovy/whelk/rest/api/Crud.groovy
lrosenstrom: Oops!

rest/src/main/groovy/whelk/rest/api/Crud.groovy
lrosenstrom: We discussed this as a possible future feature, but skip it for now.

PR #797, opened 2020-11-16
Whelktool: Add default reports

olovy:
Reduce boilerplate in whelktool scripts
* Automatically create `MODIFIED.txt`, `CREATED.txt`, `DELETED.txt` with ids of touched records.
* Add a new method `incrementStats` that if used logs to `STATISTICS.txt`

This also opens up for more automated rollback/revert of whelktool runs.

Often we want to log some more information about modified records. That is done in the same way as today, i.e. by manually creating a report writer.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/util/Statistics.groovy
+++ b/whelk-core/src/main/groovy/whelk/util/Statistics.groovy
@@ -80,4 +80,7 @@ class Statistics {
         return this
     }
 
+    boolean isEmpty() {
+        c.isEmpty()
+    }
 }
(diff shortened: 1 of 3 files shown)


jannistsiroyannis (approved):
We used to call this "scheduled.." because at the time of writing, nothing had yet been changed (it was just about to). This is better in that regard, but would actually be (more) reliable if println() statements consitently came _after_ the modification instead of before, as they are now.

olovy:
> We used to call this "scheduled.." because at the time of writing, nothing had yet been changed (it was just about to). This is better in that regard, but would actually be (more) reliable if println() statements consitently came _after_ the modification instead of before, as they are now.

:+1: 
Moved them to after. 

Too keep track of 100% in case process is killed we would need to log before and after...

PR #799, opened 2020-11-17
Add script for linking up isPartOf.

jannistsiroyannis:


klngwll (changes requested):
The 'isPartOf' entities has been moved from instanceOf to instance so the node need to be changed. There is no longer any need to check graph,2 for controlNumber and maybe more "economic" to do the selectBySQLWhere check on {graph,1,isPartOf} to not get the records of other relations with controlNumber?). 

As a point of discussion, I wonder if we need to do any additional sanity checks on the target?

jannistsiroyannis:
> The 'isPartOf' entities has been moved from instanceOf to instance so the node need to be changed. There is no longer any need to check graph,2 for controlNumber and maybe more "economic" to do the selectBySQLWhere check on {graph,1,isPartOf} to not get the records of other relations with controlNumber?).
> 
> As a point of discussion, I wonder if we need to do any additional sanity checks on the target?
There is no @graph,2 anymore anyway, we removed them all :)

What sanity checks did you have in mind?

klngwll:
> > The 'isPartOf' entities has been moved from instanceOf to instance so the node need to be changed. There is no longer any need to check graph,2 for controlNumber and maybe more "economic" to do the selectBySQLWhere check on {graph,1,isPartOf} to not get the records of other relations with controlNumber?).
> > As a point of discussion, I wonder if we need to do any additional sanity checks on the target?
> 
> There are no @graph,2 anymore anyway, we removed them all :)
> 
> What sanity checks did you have in mind?

Well since I don't "trust" controlnumbers 100% as unique identifiers, these could also have come from any external source. Perhaps the risk is not overwhelming because most of these will not link anyway so it might be overkill but we could weigh it against perhaps logging linked ids where isPartOf.hasTitle.mainTitle diff from target mainEntity.hasTitle.mainTitle or something like that? Or possibly matching the identifiedBy.value on both entities but I guess there is also a risk that this yields alot of false errors.

klngwll (approved):
LGTM!

PR #804, opened 2020-11-23
Make XL compatible with Elasticsearch 7.x

andersju:
Various changes to make XL work with Elasticsearch 7.x (and 7.x **only**).

Index changes in `libris_config.json`:

* Elasticsearch 7.0 removed mapping types, so all properties are now directly under `mappings`
* Collection name now stored in custom field `_collection`
* The `_all` special field is gone, so a custom `_all` field is added to `libris_config.json`; other fields now use `copy_to` to copy to `_all`; `index.query.default_field` set to use the new `_all` field (otherwise ES searches thousands of fields rather than, like, a dozen as before)
* Field type `edgeNGram` is now `edge_ngram`
* In ES 5.x type `string` was replaced with `keyword` and `text`, so `string` has been changed to `keyword` in `soft_template`

The removal of mapping types means ES query URLs have changed, so paths in `ElasticClient`, `ElasticSearch`, `ESQuery` and `manage-whelk-storage.sh` are updated accordingly.

Additionally:

* Aggregation order key `_term` renamed to `_key`
* `hits.total` in ES response now an object, not a number; `track_total_hits=true` now sent in queries, otherwise we only get a maximum of 10 000 hits
* Ingest pipeline (`libris_ingest_configuration.json`) removed, because it served no purpose
* @olovy improved ElasticSearch error handling (new exception: `whelk.exception.ElasticStatusException`) and removed things no longer necessary in a post-mapping-type world

Any existing index needs to be recreated and reindexed.

See LXL-2527 for more details.


olovy (approved):
LGTM :star: 

Basic verification done on dev2.

PR #805, opened 2020-11-25
Reindex affected IDs *after* going through addedLinks

andersju:
Otherwise we might set an incorrect reverse links count in Elastic in certain situations.


olovy (approved):
LGTM! :ship:

PR #807, opened 2020-12-01
ES: Add keyword field for prefLabel to enable sorting on it

olovy:
Use both text and keyword so that we still have full-text search for `prefLabel`.
Whelk automatically uses `prefLabel.keyword` for sorting.

Tested locally.


andersju (approved):
LGTM!

PR #808, opened 2020-12-01
Add script for fixing descriptionUpgrader broken by a previous script

andersju:


niklasl:
Nice! We *may* also want to check if the latest descriptionUpgrader list has had any new items added since the old script was run? (Something like fixedList += newestList - brokenResultList...) Not sure about the likelihood of that though.

olovy (approved):
LGTM! :ship:

olovy:
> Nice! We _may_ also want to check if the latest descriptionUpgrader list has had any new items added since the old script was run? (Something like fixedList += newestList - brokenResultList...) Not sure about the likelihood of that though.

Or just step through the rest of the versions and check that it hasn't changed?

andersju:
> Nice! We _may_ also want to check if the latest descriptionUpgrader list has had any new items added since the old script was run? (Something like fixedList += newestList - brokenResultList...) Not sure about the likelihood of that though.

Ah, you mentioned that before but it slipped my mind. Should be fixed now!

whelktool/scripts/cleanups/2020/12/lxl-3457-fix-descriptionUpgrader.groovy
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+
+    if (changed) {
+        // Check whether something has been added to descriptionUpgrader since it was broken
+        if ((currentBrokenList - listWhenItBroke).size() > 0) {
+            fixedList += currentBrokenList - listWhenItBroke
+            hasAdditions.println("${data.doc.getURI()}")
+        }
```
olovy, line 55: :star:

olovy (approved):
:ship:

PR #811, opened 2020-12-04
Feature/lxl 3442 definitions as dataset

olovy:
* As a temporary solution delete definitions from ES based on the new `meta.inDataset` (until "DatasetImporter" is ready). 
* This makes it possible to get rid of collection in ES index. 
* Use this to do some reindexing in bulk.


jannistsiroyannis (approved):
Looks good!

PR #812, opened 2020-12-04
Feature/extend validator statistics

niklasl:
This is mainly for adding known irregular terms to the validator statistics report.

Specifically, this now collects all terms that are sometimes, but not always, used as repeatable (set-container) terms from the MarcFrameConverter, and list all paths to these.

It also reports full paths on validation errors, and marks those that are known as problematic (from MarcFrameConverter as per above), so we may ignore them and treat those that are unknown separately (this might be because their sources aren't MARC21, or that they've been edited or processed).

The paths are also annotated with "1", "2+" or "5+" to represent one, more than one or more than four list items respectively.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
@@ -87,7 +87,7 @@ class JsonLdValidator {
     private void checkIsNotNestedGraph(String key, value, validation) {
         if (key == jsonLd.GRAPH_KEY) {
             if (validation.seenGraph) {
-                handleError(new Error(Error.Type.NESTED_GRAPH, key, value?.toString()), validation)
+                handleError(new Error(Error.Type.NESTED_GRAPH, key, value), validation)
             }
             validation.seenGraph = true
         }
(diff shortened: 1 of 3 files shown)


niklasl:
If we need this much data (on the order of 5K lines), we may want to configure the sort order in the Statistics report. I currently sort by last key in path per section to get a fairly good overview (`sort /\v\.\w+\ze(:\S+)?\s/ r` when viewing the file in Vim). Or that might be good enough if we're happy with what we get, and just chomp up the file into digestible sub-tasks as we go further.

lrosenstrom:
LGTM! Perhaps we should keep a copy of the original script as a starting point for more complex explorations in one way or the other and to have a condensed summary where changes in the numbers are visible inspection.

whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
```diff
@@ -87,7 +87,7 @@ class JsonLdValidator {
     private void checkIsNotNestedGraph(String key, value, validation) {
         if (key == jsonLd.GRAPH_KEY) {
             if (validation.seenGraph) {
-                handleError(new Error(Error.Type.NESTED_GRAPH, key, value?.toString()), validation)
+                handleError(new Error(Error.Type.NESTED_GRAPH, key, value), validation)
```
lrosenstrom, line 90: :+1:

niklasl:
> LGTM! Perhaps we should keep a copy of the original script as a starting point for more complex explorations in one way or the other and to have a condensed summary where changes in the numbers are visible inspection.

Sounds good! Done in  252a0c4.

lrosenstrom:
:ship:

lrosenstrom (approved):

PR #813, opened 2020-12-10
Search: Add 'exists-' filter prefix

olovy:
Make it possible to search for documents having a specific field or not.

### DIFF
--- a/rest/API.md
+++ b/rest/API.md
@@ -171,6 +171,7 @@ for the same field means `OR`. Specifying multiple values for the same field can
 by repeating the parameter or by giving a comma-separated list as value.
 
 * `<field>` - The record has exactly this value for `field`.  
+* `exists-<field>` - The field exists in the record. Value should be `true` or `false`.
 * `min-<field>` - Greater or equal to.
 * `minEx-<field>` - Greater than (exclusive minimum).
 * `max-<field>` - Less or equal to.
@@ -209,6 +210,15 @@ $ curl -XGET -H "Accept: application/ld+json" \
 
 #### Example
 
+Has mediaType but not carrierType.
+```
+$ curl -XGET -H "Accept: application/ld+json" \
+    'https://libris-qa.kb.se/find?exists-mediaType=true&exists-carrierType=false'
+...
+```
+
+#### Example
+
 Published in the 1760s.
 ```
 $ curl -XGET -H "Accept: application/ld+json" \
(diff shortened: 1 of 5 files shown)


lrosenstrom (approved):
Nice! Maybe substitute the query parameter 'exists' for 'has' so that the query would read better?  However, 'exists' might fit better in our ubiquitous language... :thinking:

klngwll (approved):
LGTM as far as I can see, nice with examples.

klngwll:
> Nice! Maybe substitute the query parameter 'exists' for 'has' so that the query would read better? However, 'exists' might fit better in our ubiquitous language... 🤔

Isn't 'exists' the elastic name for the querytype? that makes me want to keep it, to make any referencing easier later on even though 'has' might be easier to read/write in the actual query.

andersju (approved):
LGTM! "exists" is indeed what it's called in ES ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-exists-query.html)) so +1 on sticking to that.

lrosenstrom:
In that case I agree! :ship: :star:

olovy:
Well, IMO, ES is just an "implementation detail" in this context so we should pick the prefix that makes the best API. 
But which is it? :thinking:

andersju:
> Well, IMO, ES is just an "implementation detail" in this context so we should pick the prefix that makes the best API.
> But which one is it? thinking

From that perspective I think I do prefer `has-<field>`, and it pairs better with the existing `matches-<field>`. Or maybe `has-field-<field>` as a more future-proof solution, in case there could be things other than fields (perhaps more abstract things) that a record "has" that we'll want to be able to filter on?

niklasl (approved):
LGTM!

The XL API is (loosely) based on [The Linked Data API](https://github.com/UKGovLD/linked-data-api/blob/wiki/API_Query_Parameters.md) by the uk.gov, which uses `exists-` for this (with the same design). It is (likely) inspired by [EXISTS in SPARQL](https://www.w3.org/TR/sparql11-query/#func-filter-exists), which the LD-API was designed to be a GET-based query front for.

So while the equivalence in ES is incidental, and it is most certainly an implementation detail, I am very much in support of the `exists-` prefix, for these reasons.

Also, while `has` can read better, it becomes less obviously so with all the `has*`-named properties in KBV (`hasPart`, `hasVariant`, `hasReproduction` and some 24 more).

olovy:
Thank you all for the input.

PR #815, opened 2020-12-14
Feature/lxl 3452

jannistsiroyannis:
Lägger till OAIPMH-seten "sao" och "nb", och exportformatet "ttl".


jannistsiroyannis:
Detta har felaktigeheter, vänta med att granska.

olovy (approved):
:ship: 

Will it be possible in practice to get the full "nb" set this way or is it too slow?

jannistsiroyannis:
> ship
> 
> Will it be possible in practice to get the full "nb" set this way or is it too slow?

We'll have to try it. It won't be very fast for certain. But my guess is it will be possible. :)

PR #819, opened 2020-12-17
Fix bib 610-i1 conversion for subject organization and add spec

klngwll:
- [x] I have run marcframe integTest

Fixes conversion error of 610-i1 for organization with subordinate unit to not be a part of Work contribution in subject.

[LXL-2335](https://jira.kb.se/browse/LXL-2335)


olovy (approved):
LGTM!

PR #820, opened 2020-12-17
Fix 755 inScheme and add spec

klngwll:
- [x] I have run marcframe integTest

Removes closeMatch saogf urlmatching for external gf-systems in auth#755. 

[LXL-3448](https://jira.kb.se/browse/LXL-3448)


lrosenstrom (approved):
LGTM!

PR #821, opened 2020-12-17
Add scripts for linking contentType, mediaType, carrierType

olovy:
This could ofc be made into a single script but now it is three more or less copy-pasted scripts.


klngwll (approved):
LGTM!

PR #823, opened 2020-12-17
Add a cleanup script for prefLabels on indian terms.

jannistsiroyannis:


klngwll (approved):
Reeeally quick look before checkout. Does seem to look at the right places and work the right things. My inner safety-controller would maybe double check that '%ndian%' does not pick up more than intended, even though I can't think of anything right now.

olovy (approved):
LGTM!

Since it just rebuilds prefLabels it should be safe.

olovy:
We really should just remove all prefLabels on ComplexSubject? But that's a separate discussion.

PR #827, opened 2020-12-21
Update virtuoso triples continuously

olovy:
See #817

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -56,6 +56,13 @@ class PostgreSQLComponent {
     interface UpdateAgent {
         void update(Document doc)
     }
+
+    interface QueueHandler {
+        enum Result { HANDLED, FAIL_RETRY, FAIL_REQUEUE }
+        
+        Result handle(Document doc)
+    }
+
     public static final int STALE_UPDATE_RETRIES = 10
 
     public static final String PROPERTY_SQL_URL = "sqlUrl"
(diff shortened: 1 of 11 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -309,11 +316,26 @@ class PostgreSQLComponent {
...
             """.stripIndent()
 
+    private static final String SPARQL_QUEUE_ADD = "INSERT INTO lddb__sparql_q (id) VALUES (?)"
+
+    private static final String SPARQL_QUEUE_REMOVE = """
+            DELETE FROM lddb__sparql_q
+            WHERE pk = (
+              SELECT pk
+              FROM lddb__sparql_q
+              ORDER BY pk
```
jannistsiroyannis, line 326: Maybe add 'ASC' explicitly? I can never remeber the deafult sort order and had to look it up.

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -858,6 +878,24 @@ class PostgreSQLComponent {
         return doc
     }
 
+    private Document lockAndLoad(String id, Connection connection) throws DocumentNotFoundException {
```
jannistsiroyannis, line 881: :gun:

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -1354,6 +1396,87 @@ class PostgreSQLComponent {
...
+    boolean sparqlQueueTake(QueueHandler reader, int num, DataSource connectionPool) {
+        Connection connection = null
+        try {
+            // The queue contains document ids.
+            // Items (rows) are locked and then finally removed from the queue when we commit the transaction.
+            // If the handler fails on any item, the transaction is cancelled and all items remain in the queue.
+            // SKIP LOCKED in the query guarantees that we will take the first <num> unlocked rows.
+            //
+            // Lock the actual documents while the reader is working on them. Otherwise two readers could end up
+            // working on two different versions of the same document concurrently.
```
jannistsiroyannis, line 1429: :+1:

whelk-core/src/main/groovy/whelk/component/SparqlUpdater.groovy
```diff
@@ -0,0 +1,149 @@
...
+                if (storage.sparqlQueueTake(handler, QUEUE_TAKE_NUM, connectionPool)) {
+                    pollNow()
+                }
+            }
+            catch (Exception e) {
+                log.warn("Error executing task: $e", e)
+            }
+        }
+
+        timer.scheduleAtFixedRate({ pollNow() }, PERIODIC_CHECK_MS, PERIODIC_CHECK_MS)
```
jannistsiroyannis, line 113: This calls into executorService.submit, which CAN throw, in which case (it seems) all future timer events are stopped for ever (until program restart).
pollNow() should probably have a try/catch{_everything_}.

jannistsiroyannis (approved):
See comment about executor.submit possibly throwing, otherwise: :ship:

whelk-core/src/main/groovy/whelk/component/SparqlUpdater.groovy
olovy: Good catch (pun intended)

olovy: I think it should be safe since we're using `ThreadPoolExecutor.DiscardPolicy` but I will add a catch.

PR #830, opened 2021-01-11
Replace use of deprecated Apache HttpClient methods

olovy:
They are deprecated since HttpClient 4.3.

* Refactor `SparqlUpdater`+`Virtuoso` a little bit.
* Handle random 501 NOT IMPLEMENTED from Virtuoso


jannistsiroyannis (approved):
:ship:

PR #831, opened 2021-01-12
Make sparql-reloading possible without taking the system down in the …

jannistsiroyannis:
…meantime.


olovy (approved):
LGTM.

Checking `@graph,0,generationDate` will take quite long long time since it's not indexed but that's OK I guess.

PR #832, opened 2021-01-13
Add script for replacing work with organization in subjects

kwahlin:


jannistsiroyannis (approved):
The script looks good to me, but why are we removing the secret.properties.in file?
(I see that it needs a little love in order to be up to date)

kwahlin:
Good question :) Slipped through by mistake. Fixed!

PR #835, opened 2021-01-19
Add a normalization step to deal with multi-types (which are not allo…

jannistsiroyannis:
…wed).

### DIFF
--- a/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
+++ b/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
@@ -6,6 +6,7 @@ import whelk.JsonLd
 import whelk.Whelk
 import whelk.component.DocumentNormalizer
 import whelk.exception.InvalidQueryException
+import whelk.exception.ModelValidationException
 import whelk.filter.BlankNodeLinker
 import whelk.filter.LanguageLinker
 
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
```diff
@@ -58,6 +58,46 @@ class Normalizers {
...
 
+    static int getTypePrecisionScore(String type, JsonLd jsonLd) {
+        List superClasses = []
+        jsonLd.getSuperClasses(type, superClasses)
+        return superClasses.size()
+    }
+
+    // If an @type property containing a list is found anywhere in the data, replace it with the most
+    // "specific" (least general) member of the list.
+    static void enforceTypeSingularity(node, jsonLd) {
```
niklasl, line 69: Nothing guarantees that the given types are in the same type subset, so formally we'd have to check that the dropped types are implied by the kept type (i.e. the dropped ones are within its superclasses).

jannistsiroyannis: I agree, but:

Say that the types are unrelated. We still have to choose one (having two breaks the system). And lacking a definitive way of determining which is the more appropriate one, i believe this serves as a decent heuristic.

If you have a better heuristic for this (or a definitive way to decide) I'm all ears :)

niklasl: Multiple types are allowed (and not to uncommon) in RDF (used by, among others, the ISSN central), so it is *our* system that breaks, based on a convenient restriction (given our method of programming). So if we decide to forbid multiple unrelated types, I'd suggest throwing an error here if we can't normalize the given types. This avoids a surprising silent dropping of types somewhere down the line if we attempt to integrate other data utilizing such forms.

(As a heuristic it is better than a random selection, but since defined supertypes are merely "points of interest" and not logically defined, little to nothing can be said about which is to be preferred given two unrelated types.)

jannistsiroyannis: I see.

I will change this then, to only remove types that have another subtype in the same list, or else error out.

andersju (approved):
Looks good, as far as I can tell!

PR #836, opened 2021-01-25
LGC-39 Udix remove print

poacherone:


poacherone:
Just ignore the commits history, I didn't realise a bunch of old/irrelevant changes stuff was left there.

kwahlin (commented):
Shouldn't there be more records in Udix_ID2.txt? Also, those three records are already deleted. Maybe it's the wrong file?

poacherone:
It is a tmp file, just to test the script. I wasn't sure if it was needed for a code review. But I'll update it.

whelktool/scripts/syslib/2021/01/Udix_remove_print.groovy
```diff
@@ -0,0 +1,22 @@
+PrintWriter failedUpdating = getReportWriter("failed-updating")
+PrintWriter scheduledForUpdating = getReportWriter("scheduled-updates")
+
+File HoldIDs = new File(scriptDir, "Udix_ID2.txt")
+
+List propertiesToCheck = ["associatedMedia", "marc:hasTextualHoldingsBasicBibliographicUnit"]
+
+selectByIds( HoldIDs.readLines() )  { hold ->
+def holds = hold.graph[1]
```
andersju, line 9: Nitpick: the use of hold (an individual hold) vs holds (here a part of that individual hold) is slightly confusing. Maybe "holds" => "item" would be better. (Doesn't matter for a one-off script, but for the future :smile:).

kwahlin:
Looks good then! Maybe you should check also if the property is left empty after removing the "Externt magasin" object, and if that's the case, remove the property from graph[1]. If you think there are cases where associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit contains only one ("Externt magasin") object?

andersju (commented):
Looks fine, just a couple of minor things you might want to look at. Also be mindful of indentation (makes it easier to read / spot problems). And +1 on what Kalle wrote :)

whelktool/scripts/syslib/2021/01/Udix_remove_print.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+propertiesToCheck.each { prop ->
+    holds[prop].removeIf { it["marc:publicNote"].contains("Externt magasin / Closed stacks") }
```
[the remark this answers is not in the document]
poacherone: Good point!

whelktool/scripts/syslib/2021/01/Udix_remove_print.groovy
poacherone: > Looks good then! Maybe you should check also if the property is left empty after removing the "Externt magasin" object, and if that's the case, remove the property from graph[1]. If you think there are cases where associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit contains only one ("Externt magasin") object?

In general this is probably a very good idea, but the case here is that if there is only one object in associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit ("Externt magasin") the entire holdings record should be removed. That will be done separately (i.e. registervårdsklienten).

poacherone:
> Looks good then! Maybe you should check also if the property is left empty after removing the "Externt magasin" object, and if that's the case, remove the property from graph[1]. If you think there are cases where associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit contains only one ("Externt magasin") object?

In general this is probably a very good idea, but the case here is that if there is only one object in associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit ("Externt magasin") the entire holdings record should be removed. That will be done separately (i.e. registervårdsklienten).

kwahlin:
> > Looks good then! Maybe you should check also if the property is left empty after removing the "Externt magasin" object, and if that's the case, remove the property from graph[1]. If you think there are cases where associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit contains only one ("Externt magasin") object?
> 
> In general this is probably a very good idea, but the case here is that if there is only one object in associatedMedia/marc:hasTextualHoldingsBasicBibliographicUnit ("Externt magasin") the entire holdings record should be removed. That will be done separately (i.e. registervårdsklienten).

I see! That was the point with writing a script right...

PR #837, opened 2021-01-26
Use softmatcher for mainTitle; explicitly define responsibilityStatem…

andersju:
…ent and top-level _str.

responsibilityStatement uses softmatcher, and is copied to _str (and _all).


olovy (approved):
LGTM

PR #839, opened 2021-01-26
Feature/sync datasets

jannistsiroyannis:
This adds the feature that you can load/sync a dataset from a jsonld.lines file, into XL.

Sync, in this context, means that all records will be created/updated/deleted as needed.

This should work without issue for any "normal" datasets, but it cannot (yet) replace the definitionsloader.

There are two reasons for this:
1. The output records from generating our "definitions" are not all actually in the dataset "https://id.kb.se/dataset/definitions" (see for example context/vocab). These would need to be "broken out" into jsonld.lines files of their own, or otherwise adapted in form so that they actually "work" when saved through the normal means.
2. The existing definitionsloader uses a very _very_ dirty trick. It bypasses all forms of indexing/denormalising and just writes it's data straight into the database. That's a good thing, because if these were "saved" normally, all the rest of libris, (having dependencies on the definitions) would need to be recarded/reindexed too. We cannot use this diry trick for other arbitrary datasets however. There is a mitigation here, that if you reload definitions using this new method, only records that actually changed will be updated, which redcuces "all of libris" to "all records depending on actually changed defintions". That however might not be enough, if more than a few definitions are updated.

### DIFF
--- a/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
+++ b/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
@@ -1,0 +1,72 @@
+package whelk.importer
+
+import org.codehaus.jackson.map.ObjectMapper
+import whelk.Document
+import whelk.IdGenerator
+import whelk.Whelk
+import whelk.exception.CancelUpdateException
+
+class DatasetImporter {
+    static final ObjectMapper mapper = new ObjectMapper()
+
+    static importDataset(Whelk whelk, String filePath, String dataset, String marcCollection) {
+
+        if ( ! ["auth", "bib", "hold", "definitions"].any{ it.equals(marcCollection) } )
+            throw new RuntimeException("Unknown marc collection.")
+
+        Set<String> idsInInput = []
+
+        File inDataFile = new File(filePath)
+        inDataFile.eachLine { line ->
+            Map data = mapper.readValue(line.getBytes("UTF-8"), Map)
+            Document incomingDoc = new Document(data)
+            ensureAbsoluteSystemId(incomingDoc)
+            idsInInput.add(incomingDoc.getShortId())
+            incomingDoc.setInDataSet(dataset)
+
+            // This race condition should be benign. If there is a document with
+            // the same ID created in between the check and the creation, we'll
+            // get an exception and fail early (unfortunate but acceptable).
+            Document storedDoc = whelk.getDocument(incomingDoc.getShortId())
+            if (storedDoc != null) {
+
+                // Update (potentially) of existing document
+                whelk.storeAtomicUpdate(incomingDoc.getShortId(), true, "xl", null, { doc ->
+                    if (doc.getChecksum() != incomingDoc.getChecksum()) {
+                        //System.err.println("Did not match checksums:\n\n" + doc.getDataAsString() + "\n\n" + incomingDoc.getDataAsString())
+                        doc.data = incomingDoc.data
+                    }
+                    else {
+                        throw new CancelUpdateException() // Not an error, merely cancels the update
+                    }
+                })
+            } else {
+
+                // New document
+                whelk.createDocument(incomingDoc, "xl", null, marcCollection, false)
+            }
+
+            // Clear out anything that was previously stored in this dataset, but was not in the in-data now.
+            whelk.storage.doForIdInDataset(dataset, { String storedIdInDataset ->
+                System.err.println(storedIdInDataset)
+            })
+        }
+    }
+
+    private static void ensureAbsoluteSystemId(Document doc) {
+        def sysBaseIri = Document.BASE_URI
+        // A system id is in place; do nothing and return.
+        if (doc.recordIdentifiers.any { it.startsWith(sysBaseIri.toString()) }) {
+            return
+        }
+        // A relative system id (slug) exists; make it absolute and return.
+        def slug = doc.recordIdentifiers.find { it =~ /^\w+$/ }
+        if (slug) {
+            doc.setId(slug)
+            return
+        }
+        // Mint a new system Id and set it.
+        def newId = sysBaseIri.resolve(IdGenerator.generate()).toString()
+        doc.addRecordIdentifier(newId)
+    }
+}
--- a/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
+++ b/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
@@ -14,7 +14,7 @@ class DatasetImporter {
 
         if ( ! ["auth", "bib", "hold", "definitions"].any{ it.equals(marcCollection) } )
             throw new RuntimeException("Unknown marc collection.")
-        
+
         if (Runtime.getRuntime().maxMemory() < 2l * 1024l * 1024l * 1024l) {
             Log.warn("This application may require substantial amounts of memory, " +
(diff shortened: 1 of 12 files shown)


olovy (commented):
* Shouldn't the main thing IRI be the identity of an item in the dataset?
* I think we need an index on `inDataset`?
    * e.g. `CREATE INDEX idx_lddb_indataset on lddb using gin ((data#>'{@graph,0,inDataset}'));`
* Nice to have: some progress indication (for large datasets with many dependers)

importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
```diff
@@ -0,0 +1,126 @@
...
+import whelk.IdGenerator
+import whelk.Whelk
+import whelk.exception.CancelUpdateException
+
+class DatasetImporter {
+    static final ObjectMapper mapper = new ObjectMapper()
+
+    static void importDataset(Whelk whelk, String filePath, String dataset, String marcCollection) {
+
+        if ( ! ["auth", "bib", "hold", "definitions"].any{ it.equals(marcCollection) } )
```
niklasl, line 15: Introducing datasets is another step towards possibly removing our LDDB collection notion (or at least reduce it to an internal optimization). I assume we don't want to do that in one step, but at this point we reasonably don't want to allow datasets to be anything other than in "definitions" either (which today basically means "from a dataset" (and is close to "any record not in auth, bib, or hold", albeit I've gathered the abstract Works fell in between for auth export reasons)).

So I suggest hardcoding the "definitions" collection for datasets (and note that in the issue about removing collections).

whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
```diff
@@ -77,7 +77,11 @@ class Normalizers {
                         if (typeList.size() == 1)
                             node[key] = typeList[0]
                         else {
-                            throw new ModelValidationException("Could not reduce: " + typeList + " to a single type (required) by removing superclasses.")
+
+                            //throw new ModelValidationException("Could not reduce: " + typeList + " to a single type (required) by removing superclasses.")
+                            // This must be reduced to a warning, because the assumption does not hold true: There are records in
+                            // libris that legitimately have unrelated multi-types (see definitions).
+                            log.warn("Could not reduce: " + typeList + " to a single type by removing superclasses.")
```
niklasl, line 84: If that was only because of the `a owl:Class, rdfs:Datatype` items, we could do something about that in the definitions (also, `rdfs:Datatype rdfs:subClassOf rdfs:Class` but our system doesn't know that :/).

whelk-core/src/main/groovy/whelk/Document.groovy
```diff
@@ -109,6 +110,31 @@ class Document {
         return mapper.writeValueAsString(data)
     }
 
+    void setInDataSet(String dataset) {
```
niklasl, line 113:
```suggestion
    void setInDataset(String dataset) {
```
for consistent casing.

niklasl:
Regarding the points by @jannistsiroyannis:

1. Very true, we need to address this in LXL-3442. Note that with that, we may not have a definitions dataset in the end (in fact, we reasonably don't want to, depending on how we update our workflows). It makes sense to start out with that, provided that we can split it up into "core" (vocab and related enums), and then manage other updates based on "upstream" changes for the respective sources (which will be more diverse going forward with other external or externally derived datasets, such as bibdb, geographic names and so on.).
2. Also crucially important, and I agree with your assessment. Depending on where we're at regarding dependency tracking and checking if a dependent chip or card is changed, any administrative updates for e.g. `language/swe` won't be a problem (unless its prefLabel is changed), it may or may not suffice to rely on the mitigation.

niklasl:
> * Shouldn't the main thing IRI be the identity of an item in the dataset?

Actually no, as it is the *record* that belongs to the dataset for these structures.

(A main entity might belong to a dataset as part of the cataloguing, but that is crucially different, as the described entity isn't (regularly) kept in LDDB, only its description&mdash;in the record. In fact, for this dataset logic, we might even consider it a safety-check to ensure that the description of a given dataset to be updated actually *belongs to its main entity*. That is, given that we are to process `?ds`, it ought to hold that `?dsr :inDataset; ?ds ; :mainEntity ?ds`. But it is not necessary at this point, as this is the only use of `Dataset` and `inDataset` we employ.  I mainly want to accentuate that since a `Dataset` is a kind of `Work`, and may figure as types of regular subjects in the domain we describe, the way we collect and manage records belonging to datasets is kept under specific conditions. The crucial part of that is that *records* belong to them, not other main entities.)

> * I think we need an index on `inDataset`?
>   
>   * e.g. `CREATE INDEX idx_lddb_indataset on lddb using gin ((data#>'{@graph,0,inDataset}'));`

Makes sense. (Maybe append `,@id` to that, albeit in shouldn't (...) make that much of a difference as the flattened structure ought to keep as a reference node (i.e. having just the `@id` key).)

> * Nice to have: some progress indication (for large datasets with many dependers)

😎

whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
jannistsiroyannis: I'm working on the other stuff, but for reference, here are the remaining multi types in definitions:

[Class, Datatype] to a single type by removing superclasses.
[Class, IssuanceType] to a single type by removing superclasses.
[FunctionalProperty, ObjectProperty] to a single type by removing superclasses.
[marc:CollectionClass, Class] to a single type by removing superclasses.
[marc:HoldingStatusType, marc:AuthorityStatusType] to a single type by removing superclasses.
[marc:I1codeSource, marc:LanguageCode, marc:I2sourceOfCode, marc:NumberSource] to a single type by removing superclasses.
[marc:SummaryType, marc:AudienceType] to a single type by removing superclasses.
[ObjectProperty, DatatypeProperty] to a single type by removing superclasses.
[ObjectProperty, owl:SymmetricProperty] to a single type by removing superclasses.

importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
```diff
@@ -0,0 +1,135 @@
...
+        }
+        // A relative system id (slug) exists; make it absolute and return.
+        def slug = doc.recordIdentifiers.find { it =~ /^\w+$/ }
+        if (slug) {
+            doc.setId(slug)
+            return
+        }
+        // Mint a new system Id and set it.
+        def newId = sysBaseIri.resolve(IdGenerator.generate()).toString()
+        doc.addRecordIdentifier(newId)
```
olovy, line 133: Do we want to allow this case?
These will only be possible to import once (or import, read out, rebuild dataset I guess).

olovy (approved):
LGTM! :ship: 

See comment about docs with missing record ID.

Let's discuss what we can do about point #2 by Jannis.

 We should consider if there is any further documentation needed
* the notion of dataset and how we intend to use it
* expected input to (and maybe behavior of) the importer

importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
jannistsiroyannis: Good catch! That code was copy pasted from the definitions importer, but you're right - the IDs can't be "random" like that! I'll fix it.

PR #840, opened 2021-01-27
Feature/virtuoso namespace prefixes

kwahlin:
Se användning i https://github.com/libris/devops/pull/80


librisxl-tools/scripts/install_namespace_prefixes.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Remove predefined prefixes
+DELETE_PREFIXES=$(/usr/local/bin/isql -U dba -P "$2" MESSAGE=OFF BANNER=OFF VERBOSE=OFF "EXEC=SELECT NS_PREFIX FROM DB.DBA.SYS_XML_PERSISTENT_NS_DECL" \
+  | awk '{ print "DB.DBA.XML_REMOVE_NS_BY_PREFIX('\''"$0"'\'', 2);"}')
+/usr/local/bin/isql -U dba -P "$2" EXEC="$DELETE_PREFIXES"
```
jannistsiroyannis, line 6: You "shouldn't" need an absolute path to isql (it should be available under the standard $PATH). It would be one thing less you need to remeber if the paths ever change.

jannistsiroyannis (approved):
So I take it these are "default" namespace prefixes, that you don't need to specify with each query then ?
:ship:

librisxl-tools/scripts/install_namespace_prefixes.sh
kwahlin: Running the script with fab without stating the absolute path to isql did for some reason not work in the first place. However, now that I tried it again, it worked just fine! So all the '/usr/local/bin/isql' has now been replaced with just 'isql'. And the whole $PATH content ended up in the commit message...:smile:

kwahlin: > So I take it these are "default" namespace prefixes, that you don't need to specify with each query then ?
> ship

Correct!

PR #841, opened 2021-01-27
Add script for linking blank inScheme in hold

andersju:
Simply copied from #761 and modified to work with holds.


olovy (approved):
:ship:

PR #844, opened 2021-02-02
Feature/lxl 3484 redirect genre form links

kwahlin:


jannistsiroyannis (changes requested):
This replaces the URIs without preserving the old ones into sameAs-lists.
Also, there's no need for this. The "registervårdsklient" can do this already.

klngwll:
> This replaces the URIs without preserving the old ones into sameAs-lists.
> Also, there's no need for this. The "registervårdsklient" can do this already.

The old URI:s is not sameAs replaced by the "real saogf", they are just faulty auths. So there is no need to preserve a sameAs in this case.

kwahlin:
> This replaces the URIs without preserving the old ones into sameAs-lists.
> Also, there's no need for this. The "registervårdsklient" can do this already.

You mean the old URI should be added to the new URI's sameAs-list? In that case maybe the old URI should be flagged as obsolete rather than deleted? Or can we delete it and have a "dead link"?

kwahlin:
Ooops, didn't see Fredrik's comment :)

jannistsiroyannis:
> > This replaces the URIs without preserving the old ones into sameAs-lists.
> > Also, there's no need for this. The "registervårdsklient" can do this already.
> 
> The old URI:s is not sameAs replaced by the "real saogf", they are just faulty auths. So there is no need to preserve a sameAs in this case.

Unless I've completely misunderstood what's going on here, you're going to replace links to: "https://libris.kb.se/xv8cln7g41q4n6h#it" with links to another record, and delete xv8cln7g41q4n6h ?

In the (theoretical) case that some one out there "has" a link to the old URI, we still need to honor it (we can't change their data like we can our own). We cant simply "drop" URIs. They are supposed to be forever.

Or did I completely misundertand whats going on ?

klngwll:
> > This replaces the URIs without preserving the old ones into sameAs-lists.
> > Also, there's no need for this. The "registervårdsklient" can do this already.
> 
> You mean the old URI should be added to the new URI's sameAs-list? In that case maybe the old URI should be flagged as obsolete rather than deleted? Or can we delete it and have a "dead link"?



> > > This replaces the URIs without preserving the old ones into sameAs-lists.
> > > Also, there's no need for this. The "registervårdsklient" can do this already.
> > 
> > 
> > The old URI:s is not sameAs replaced by the "real saogf", they are just faulty auths. So there is no need to preserve a sameAs in this case.
> 
> Unless I've completely misunderstood what's going on here, you're going to replace links to: "https://libris.kb.se/xv8cln7g41q4n6h#it" with links to another record, and delete xv8cln7g41q4n6h ?
> 
> In the (theoretical) case that some one out there "has" a link to the old URI, we still need to honor it (we can't change their data like we can our own). We cant simply "drop" URIs. They are supposed to be forever.
> 
> Or did I completely misundertand whats going on ?

Haha first managed to edit your comment?!

The old URI:s is "non authorized" anonymous terms (in english no less) in auth that is not in any GF system so we move these to the closest Swedish saogf term which should be used. I consider these for this reason as deletable as a bib-record. We do not have a proper redirect for these terms AFAIK either?

jannistsiroyannis:
> > > This replaces the URIs without preserving the old ones into sameAs-lists.
> > > Also, there's no need for this. The "registervårdsklient" can do this already.
> > 
> > 
> > You mean the old URI should be added to the new URI's sameAs-list? In that case maybe the old URI should be flagged as obsolete rather than deleted? Or can we delete it and have a "dead link"?
> 
> > > > This replaces the URIs without preserving the old ones into sameAs-lists.
> > > > Also, there's no need for this. The "registervårdsklient" can do this already.
> > > 
> > > 
> > > The old URI:s is not sameAs replaced by the "real saogf", they are just faulty auths. So there is no need to preserve a sameAs in this case.
> > 
> > 
> > Unless I've completely misunderstood what's going on here, you're going to replace links to: "https://libris.kb.se/xv8cln7g41q4n6h#it" with links to another record, and delete xv8cln7g41q4n6h ?
> > In the (theoretical) case that some one out there "has" a link to the old URI, we still need to honor it (we can't change their data like we can our own). We cant simply "drop" URIs. They are supposed to be forever.
> > Or did I completely misundertand whats going on ?
> 
> Haha first managed to edit your comment?!
> 
> The old URI:s is "non authorized" anonymous terms (in english no less) in auth that is not in any GF system so we move these to the closest Swedish saogf term which should be used. I consider these for this reason as deletable as a bib-record. We do not have a proper redirect for these terms AFAIK either?

I get it :) So you don't actually want the old URIs to point to the new record (they're not the same)?
But, for the record (no pun), all records are deletable (auth too), the difference here is wheter we respond with a redirect on the old URI or a 410 gone, basically.

klngwll:
> > > > This replaces the URIs without preserving the old ones into sameAs-lists.
> > > > Also, there's no need for this. The "registervårdsklient" can do this already.
> > > 
> > > 
> > > You mean the old URI should be added to the new URI's sameAs-list? In that case maybe the old URI should be flagged as obsolete rather than deleted? Or can we delete it and have a "dead link"?
> > 
> > 
> > > > > This replaces the URIs without preserving the old ones into sameAs-lists.
> > > > > Also, there's no need for this. The "registervårdsklient" can do this already.
> > > > 
> > > > 
> > > > The old URI:s is not sameAs replaced by the "real saogf", they are just faulty auths. So there is no need to preserve a sameAs in this case.
> > > 
> > > 
> > > Unless I've completely misunderstood what's going on here, you're going to replace links to: "https://libris.kb.se/xv8cln7g41q4n6h#it" with links to another record, and delete xv8cln7g41q4n6h ?
> > > In the (theoretical) case that some one out there "has" a link to the old URI, we still need to honor it (we can't change their data like we can our own). We cant simply "drop" URIs. They are supposed to be forever.
> > > Or did I completely misundertand whats going on ?
> > 
> > 
> > Haha first managed to edit your comment?!
> > The old URI:s is "non authorized" anonymous terms (in english no less) in auth that is not in any GF system so we move these to the closest Swedish saogf term which should be used. I consider these for this reason as deletable as a bib-record. We do not have a proper redirect for these terms AFAIK either?
> 
> I get it :) So you don't actually want the old URIs to point to the new record (they're not the same)?
> But, for the record (no pun), all records are deletable (auth too), the difference here is wheter we respond with a redirect on the old URI or a 410 gone, basically.

Exactly, they are neither the same or replaced so in this particular case a 410 gone, I would consider fine :)

jannistsiroyannis (approved):

andersju (approved):
Given the above discussion (and local testing with some imported records): LGTM!

PR #845, opened 2021-02-02
Add script for recovering and linking isPartOf and supplementTo relat…

kwahlin:
…ions

### DIFF
--- a/whelktool/scripts/cleanups/2021/01/lxl-3476-lost-isPartOf-and-supplementTo.groovy
+++ b/whelktool/scripts/cleanups/2021/01/lxl-3476-lost-isPartOf-and-supplementTo.groovy
@@ -1,0 +1,125 @@
+import whelk.Document
+
+['isPartOf', 'supplementTo'].each { property ->
+    String where
+
+    if (property == 'isPartOf')
+        where = "collection = 'bib' and data#>>'{@graph,1,isPartOf}' LIKE '%{}%'"
+    if (property == 'supplementTo')
+        where = "collection = 'bib' and data#>'{@graph,1,supplementTo}' @> '[null]'"
+
+    selectBySqlWhere(where) { data ->
+        List propertyValueOfCurrentVersion = data.graph[1][property]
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2021/01/lxl-3476-lost-isPartOf-and-supplementTo.groovy
```diff
@@ -0,0 +1,125 @@
...
+        List propertyValueBeforeLoss
+        List propertyValueRightAfterLoss
+
+        // Get all versions sorted by generationDate, most recent first
+        List<Document> versions = data.whelk.storage.loadAllVersions(data.doc.shortId).sort { a, b ->
+            b.getGenerationDate() <=> a.getGenerationDate()
+        }
+
+        for (i in 0..<versions.size()) {
+            propertyValue = versions[i].data['@graph'][1][property] ?: versions[i].data['@graph'][2][property]
```
jannistsiroyannis, line 22: Aah, I get it. Historic versions could still have @graph,2, right, before we moved them? But couldn't the thing you're looking for also be in @graph,1,instanceOf then ? Or do we know that the "mistake" whatever it was happened before the move.

Also, what if they're both non-null (in which case you'll only check one of them later)? Perhaps I'm confused.

kwahlin: Not crystal clear, but yes, we are assuming that the relations were never in @graph,1,instanceOf. The relations are found in the instance (@graph, 1) in newer versions and before that in the work (@graph, 2), however never in both places, or in  @graph,1,instanceOf.

jannistsiroyannis: Fair enough!

jannistsiroyannis (approved):

klngwll (approved):
LGTM

PR #847, opened 2021-02-03
Add draft for sparql example queries

kwahlin:
Someone should review these before we make them public


jannistsiroyannis (approved):
:ship:

PR #857, opened 2021-02-22
Feature/validation starvation

klngwll:
Removal of obsolete marc properties from Marcframe. See individual commits for details.


olovy (approved):
:ship:

PR #859, opened 2021-02-24
Add script for linking bibliographies

olovy:


klngwll (approved):
LGTM! I have some open questions but we can get to that once we try this out in a proper environment since they might be "theoretical" anyway.

niklasl:
Do we want to change marcframe.json to generate these links on convert? Might just be theoretically good though, in case we never import instances with bibliographies... _Or_ we need to use e.g. LinkFinder to find them since they might not be "our" codes if they are given...

PR #861, opened 2021-02-26
Add script for mapping fields 655 and 856 in 'channel records' to the…

kwahlin:
…ir linked catalogue records


klngwll (commented):
Maybe also do an extra securitycheck that the record is actually part of EPLK, (graph,0):
bibliography": [
        {
          "@type": "Library",
          "sigel": "EPLK"
        }
      ],

kwahlin:
@klngwll good suggestion. Added this check now and it indeed filters out a few records.

klngwll (approved):
LGTM!

PR #864, opened 2021-03-08
Remove out-dated instructions

lrosenstrom:
Maybe we can remove some of the oldest scripts at 

https://github.com/libris/librisxl/tree/develop/librisxl-tools/scripts

such as 

https://github.com/libris/librisxl/librisxl-tools/scripts/get-record.sh ?

Please give some input in that case, or commit directly to this branch.

Although they might be interesting from a historical perspective, it might give a false picture of the current state of the system for an interested external visitor browsing the repository.


klngwll:
Maybe quarantine them in a scripts/$(outdated|archaic|historic) folder for starters unless we are very sure they have no use?

PR #866, opened 2021-03-09
Access control for ShelfMarkSequence

olovy:
* User can only write ShelfMarkSequence if
    * it is owned owned by sigel (meta.descriptionCreator).
    * user is "global registrant" (must be able to write any holding)
* Holding can only link to ShelfMarkSequence owned by the same sigel.


lrosenstrom:
LGMT but do we really want to throw a RTE?

PR #870, opened 2021-03-11
ES index: Also embellish "hold" when indexing

olovy:
Needed to be able to display `Item` better in search hit lists.

We'll have to test how this affects index size and indexing time.


andersju (approved):
LGTM!

PR #876, opened 2021-03-18
Don't accidentally create empty .venv files

andersju:
Makes sure `build-local-definitions.sh` doesn't create empty `.venv` files (this has occasionally messed up Jenkins builds).


olovy (approved):
LGTM!

PR #881, opened 2021-03-25
ISNI

poacherone:
Add ISNI identifiers to authority records, based on a list (from ISNI).


whelktool/scripts/syslib/ISNI.groovy
```diff
@@ -0,0 +1,43 @@
+PrintWriter failedAuthIDs = getReportWriter("failed-authIDs")
+PrintWriter scheduledForUpdating = getReportWriter("scheduled-updates")
```
jannistsiroyannis, line 2: We also used to do it this way (and there's nothing wrong with it!), but since a few months, this is now also done implicitly.

So there will be (other) logfiles corresponding to scheduled-to-save and failed-to-save, even if you don't do this. Might be good to know!

whelktool/scripts/syslib/ISNI.groovy
```diff
@@ -0,0 +1,43 @@
...
+def ids = selectByIds(listofIds) { auth ->
+
+def aukt = auth.graph[1]
+def uri = auth.graph[0]["@id"]
+
+id = uri.split('/')[3] // ta ut id från uri
+
+def ISNI = isnimap[id] // matcha ISNI-nummer mot id
+
+aukt["identifiedBy"]?.removeIf { it["typeNote"]?.contains("isni") || false } // radera element i identifiedBy som har ISNI-nummer (ofta felaktiga - dessa kan iofs vara korrekta, men läses in på nytt i senare steg))
```
jannistsiroyannis, line 24: This looks wrong to me!

it["typeNote"]?.contains("isni") reasonably returns true or false (or null which groovy considers false)

{ true or false } || false is, redundant at best. "OR false" is by definition a No-op.

You should probably remove the "|| false" part!

jannistsiroyannis (approved):
What editor are you using to write these. I ask only becuase the indentation is seriously off (it doesn't matter techincally, but still).

poacherone:
Heh. I use Komodo Edit. I think the weird indentation is mostly a combination of some copy-pasting and me forgetting to fix it properly. I'll be sure to look over it before submitting next time. =)

whelktool/scripts/syslib/ISNI.groovy
poacherone: Yes, I've seen this (other log-files). That these remain in this script is because we started on it quite some time ago, and I simply forgot to remove them at a later point.

whelktool/scripts/syslib/ISNI.groovy
poacherone: Thanks!

klngwll (commented):
There is also the form of identifiedBy with {"@type": "ISNI"} while only a hundred or so exist today, maybe it should be checked also? (When we fix the conversion for types, this will also be the preferred form).

whelktool/scripts/syslib/ISNI.groovy
poacherone: Actually, without || false, it will throw a NullPointer exception if it returns null. So added back again.

PR #882, opened 2021-03-25
Create new holding from existing holdings record

poacherone:
Libraries with multiple sigels may want to create new holdings based on another holdings record (from one of their other sigels). The use case is usually when the Libris holdings data is not up to date with their local data.

### DIFF
--- a/whelktool/scripts/syslib/Copy_holdings.groovy
+++ b/whelktool/scripts/syslib/Copy_holdings.groovy
@@ -1,0 +1,45 @@
+String CURRENT_SIGEL = 'https://libris.kb.se/library/Ssb'
+String NEW_SIGEL = 'https://libris.kb.se/library/Mlc'
+
+File bibIDs = new File(scriptDir, "Copy_hold_bibID.txt")
+
+def holdList = [] // Eller behövs Vector???
+
+String bibidstring = bibIDs.readLines().join("','")
+
+selectBySqlWhere("""id in
+
+(
+ select lh.id
+  from
+   lddb lb
+  left join
+   lddb lh
+  on lh.data#>>'{@graph,1,itemOf,@id}' = lb.data#>>'{@graph,1,@id}'
+ where lb.data#>>'{@graph,0,controlNumber}' in ( '$bibidstring' ) and lb.collection = 'bib'
+ and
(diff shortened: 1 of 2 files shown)


whelktool/scripts/syslib/Copy_holdings.groovy
```diff
@@ -0,0 +1,45 @@
...
+selectBySqlWhere("""id in
+
+(
+ select lh.id
+  from
+   lddb lb
+  left join
+   lddb lh
+  on lh.data#>>'{@graph,1,itemOf,@id}' = lb.data#>>'{@graph,1,@id}'
+ where lb.data#>>'{@graph,0,controlNumber}' in ( '$bibidstring' ) and lb.collection = 'bib'
```
jannistsiroyannis, line 19: To be clear: You are sure the list of bib-ids really are "bibids"?

Because @graph,0,controlNumber could look something like 12345 (for older records).

whelktool/scripts/syslib/Copy_holdings.groovy
```diff
@@ -0,0 +1,45 @@
...
+ and
+ lh.data#>>'{@graph,1,heldBy,@id}' = '{$CURRENT_SIGEL}'
```
jannistsiroyannis, line 21: This might be correct, but it looks off to me. Shouldn't it be:
either
${CURRENT_SIGEL}
or
$CURRENT_SIGEL
?

jannistsiroyannis (approved):
The indentation is a little off, and I had two comments you might want to take a look at. Otherwise it's fine :)

whelktool/scripts/syslib/Copy_holdings.groovy
poacherone: The List "bib-IDs" are actually controlnumbers. Maybe that should be more clearly stated in the naming of the file (and possibly the variable). That part of the code is lifted from the move-holds template actually.

whelktool/scripts/syslib/Copy_holdings.groovy
poacherone: That is a typo, I'll fix it. Interesting that it still seems to work. =)

PR #886, opened 2021-04-06
Take virtualdelete and exportdeleted settings from export profile.

jannistsiroyannis:
If any of these are passed as HTTP parameters (which used to be the
only way), the HTTP parameters still take precedence.

### DIFF
--- a/marc_export/src/main/java/whelk/export/marc/MarcHttpExport.java
+++ b/marc_export/src/main/java/whelk/export/marc/MarcHttpExport.java
@@ -124,9 +124,9 @@ public class MarcHttpExport extends HttpServlet
             deleteMode = ProfileExport.DELETE_MODE.EXPORT;
 
         // Virtual deletion means: If there exists no holding for any of the locations in the profile for a bib record, consider that bib record deleted.
-        boolean doVirtualDeletions = true; // Default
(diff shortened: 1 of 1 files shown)


marc_export/src/main/java/whelk/export/marc/MarcHttpExport.java
```diff
@@ -146,6 +119,44 @@ private void doPost2(HttpServletRequest req, HttpServletResponse res) throws IOE
...
+                    default:
+                        failedRequests.labels("Invalid option for 'deleted'",
+                                Integer.toString(HttpServletResponse.SC_BAD_REQUEST)).inc();
+                        res.sendError(HttpServletResponse.SC_BAD_REQUEST, "Valid options for \"deleted\" are \"ignore\", \"export\" or \"append\"");
+                        return;
+                }
+            }
+
+            if (parameterMap.get("virtualDelete") != null && parameterMap.get("virtualDelete").equalsIgnoreCase("true"))
+                doVirtualDeletions = true;
```
olovy, line 157: Need to check for false here

olovy (changes requested):

olovy (approved):

PR #887, opened 2021-04-08
Add an extra (secret) x-withSilentUpdates option to OAI-PMH ListRecor…

jannistsiroyannis:
…ds/ListIdentifiers calls, to also select any records that where _silently_ changed within the interval.


jannistsiroyannis:
Wait! There's a logic error here, don't reivew yet.

A modified outside the interval on one end, and a generation date outside on the other end will incorrectly be classified as "inside" the interval.

jannistsiroyannis:
Now it should work ok!

andersju (approved):
LGTM!

olovy (approved):
LGTM

PR #888, opened 2021-04-09
Add common field for sorting, based on chips

andersju:
Currently Instance/Work/Concept can be sorted alphabetically (using `hasTitle.mainTitle` / `prefLabel`), but we can't sort mixed results ("Alla", "Används i"), because we don't have a sort field common to all types. (Also, agents can't be sorted.)

This adds such a field to Elasticsearch, `_sortKeyByLang`. It contains two properties (sub-fields), `sv` and `en`. Each is a string containing the property values of the thing's chips in the order the properties are specified in the chip definition(s) `showProperties`. _byLang_ versions (e.g., `prefLabelByLang`, `titleByLang`) are used instead of the non-language-specific version when available (just like in the frontend).

The goal is to generate essentially the same as what the user sees in the frontend (in lxlviewer it's controlled by `header` in https://github.com/libris/lxlviewer/blob/develop/viewer/vue-client/src/resources/json/displayGroups.json).

Can be quickly tested with a whelktool script:
```groovy
import static groovy.json.JsonOutput.*
private pp(thing) {
    println(prettyPrint(toJson(thing)))
}
selectByIds(['bb59plxfd35ggq1t', 'v73tnvkjshb2hwp1', 'nm2hp58rq35gzd61', 'hvqllf5gf6dhdtkx']) { data ->
    pp(data.whelk.jsonld.toChipAsMapByLang(data.graph[1]))
}
```
=>
```
{
    "sv": "Datorbaserat multimedium",
    "en": "Interactive multimedia"
}
{
    "sv": "Statsvetenskap--teori, filosofi",
    "en": "Statsvetenskap--teori, filosofi"
}
{
    "sv": "Deskryptory Biblioteki Narodowej, dbn",
    "en": "National Library of Poland Descriptors, dbn"
}
{
    "sv": "Meghan, hertiginna av Sussex, 1981-",
    "en": "Meghan, hertiginna av Sussex, 1981-"
}
```

After recreating the index and reindexing, `&_sort=_sortKeyByLang.sv` and `&_sort=_sortKeyByLang.en` should work.

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -254,6 +254,10 @@
                 "type": "keyword",
                 "copy_to": "_all"
             },
+            "_sortKeyByLang": {
+                "type": "icu_collation_keyword",
+                "language": "sv"
+            },
             "_all": {
                 "type": "text",
                 "store": false,
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -255,8 +255,16 @@
                 "copy_to": "_all"
             },
             "_sortKeyByLang": {
-                "type": "icu_collation_keyword",
-                "language": "sv"
+                "properties": {
+                    "sv": {
+                        "type": "icu_collation_keyword",
+                        "language": "sv"
+                    },
+                    "en": {
+                        "type": "icu_collation_keyword",
+                        "language": "en"
+                    }
+                }
             },
             "_all": {
                 "type": "text",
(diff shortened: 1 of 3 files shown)


andersju:
Needs some more work, hold off on this.

andersju:
Should work better now!

olovy (approved):
LGTM!

I would like the list of languages (`['sv', 'en']`) to be a constant in `ElasticSearch` (also replacing use in `filterLanguages`) and passed as a parameter to `toChipAsListByLang`. The constant should later be replaced by configuration.

olovy (approved):
:ship:

PR #890, opened 2021-04-13
Add script for linking expressionOf

olovy:


whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
```diff
@@ -0,0 +1,194 @@
...
+                e['@id'] = worksLang[0]['@id']
+                modified |= true
+            }
+            else {
+                multiMatch.println("${bib.doc.id} ${toString(e)} ??? ${worksLang.collect {it['@id'] + " " + toString(it)}}")
+            }
+        }
+        else {
+            def contribution = asList(getPathSafe(bib.doc.data, ['@graph', 1, 'instanceOf', 'expressionOf'])).findAll{ it['@type'] == 'PrimaryContribution'}
+            def ee = contribution ? e + ['contribution': contribution] : e
```
kwahlin, line 72: What is the purpose of this part? Not sure that I understand.

kwahlin (approved):
LGTM!

whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
olovy: Keeping track of expressionOf that were not linked. Including contribution from the "main" work. Identical ones with many copies could be extracted to new records.

kwahlin: I see! But shouldn't the path to the "main" work contribution be just ['@graph', 1, 'instanceOf']?

kwahlin: Forget that, the "main" work is of course the entity in 'expressionOf', and different expressions of it can have different contributions while still pointing to the same main work right?

klngwll (approved):
LGTM! I guess we will have to tweak this as modeling goes alongs as well as see to reuse in subject for example.

PR #893, opened 2021-04-16
Add script validation analysis for auth collection

klngwll:
Copy of the original bib validation script by @lrosenstrom with all 'bib' replaced with 'auth'.

### DIFF
--- a/whelktool/scripts/analysis/validator-statistics-auth.groovy
+++ b/whelktool/scripts/analysis/validator-statistics-auth.groovy
@@ -1,0 +1,46 @@
+import whelk.JsonLdValidator
+import whelk.JsonLdValidator.Error as JsonError
+import whelk.util.Statistics
+
+class Script {
+    static PrintWriter report
+    static Statistics s = new Statistics(6).printOnShutdown()
+    static JsonLdValidator v
+    static boolean isInitialized = false
+}
+
+Script.report = getReportWriter("report.txt")
+
+selectByCollection('auth') { auth ->
+    synchronized (this) {
+        if (!Script.isInitialized) {
+            Script.v = JsonLdValidator.from(auth.whelk.jsonld)
+            Script.v.skipUndefined()
+            Script.isInitialized = true
+        }
+    }
+
+    try {
+        process(auth)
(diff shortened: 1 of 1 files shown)


whelktool/scripts/analysis/validator-statistics-auth.groovy
```diff
@@ -0,0 +1,46 @@
...
+    static PrintWriter report
+    static Statistics s = new Statistics(6).printOnShutdown()
+    static JsonLdValidator v
+    static boolean isInitialized = false
+}
+
+Script.report = getReportWriter("report.txt")
+
+selectByCollection('auth') { auth ->
+    synchronized (this) {
```
jannistsiroyannis, line 15: What is "this" in this context (since it's called from a function and not a method)? (Not trying to be snarky, genuine question!)

Is it some global object (which is fine), or is it some temporary object, owned by the executing thread (which makes the synchronized bit not have any effect)?

If the answer is "don't know", consider perhaps synchronizing over some "new Object();" in the global scope instead.

jannistsiroyannis (approved):
I had a comment, otherwise LGTM! :)

whelktool/scripts/analysis/validator-statistics-auth.groovy
lrosenstrom: If I remember correctly that construction is stolen (by me) from some @olovy - script. It's used to synchronize a block of code rather than a whole method to allow for concurrent execution in the rest of the code: https://stackoverflow.com/questions/13264726/java-syntax-synchronized-this.

lrosenstrom (approved):
:ship:

whelktool/scripts/analysis/validator-statistics-auth.groovy
olovy: In this case `this` is the magic groovy script object. (I hope!)

klngwll: @lrosenstrom ? I just did a superlazy search replace bib with auth in the script 😄

PR #894, opened 2021-04-16
Search: Hide already selected facets

olovy:
Don't return facets on fields that already have an active facet.
        Now multiple conditions on the same field are always ORed, which means facets will behave weird.
        This can be changed if we add a way to have x=a AND x=b in the API. 
        Now: if x=a is selected, the facet for x=b will show the number for (x=a AND x=b) but when selected the 
        results will be (x=a OR x=b).

Return a chip in `object` for already selected facets (`mappings`) that the frontend can use for display (see https://github.com/libris/lxlviewer/pull/735).


andersju (approved):
LGTM!

PR #895, opened 2021-04-16
Add script for merging series memberships

kwahlin:


whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
```diff
@@ -0,0 +1,256 @@
...
+
+    List validPairs = matchedPairs.findAll {
+        matchCount[it[0]] == 1 && matchCount[it[1]] == 1
+    }
+
+    // Remove matched objects and add the merged pair
+    validPairs.each {
+        seriesMembership.remove(it[0])
+        seriesMembership.remove(it[1])
+        seriesMembership << it[0] + it[1]
```
olovy, line 79: Is this the form we want for the merged SeriesMembership?

whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
```diff
@@ -0,0 +1,256 @@
...
+                    asList(sm.inSeries?.instanceOf)[0]?.hasTitle?.find { it["@type"] == "Title" }?.mainTitle
+            ]
+
+    if (values.any { it instanceof List && it.size() > 1 })
+        return true
+
+    return false
+}
+
+boolean aMatch(String a490, String a830) {
```
olovy, line 107: This could be a generalized matchStrings method?

olovy (approved):
See comment about the merged SeriesMembership, otherwise LGTM.

I think it might have been more clear to express everything in terms of the BIBFRAME/KBV terms instead of MARC fields. Also maybe to convert each SeriesMembership to a "canonical" form and then compare them instead of always checking 490 -> 830. But this works!

whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
```diff
@@ -0,0 +1,256 @@
...
+    int editDist = distMatrix.last().last()
+    // The edit distance doesn't consider string lengths
+    // E.g. "hej" and "bra" gets the same value (3) as "hej" and "hejsan"
+    // So we need to normalize to get a good metric
+    // The normalized value for "hej" and "bra" is 0, and for "hej" and "hejsan" it's 1/2
+    int longest = Math.max(w1length, w2length)
+    def normalizedDist = (longest - editDist) / longest
+
+    // Max value is 1 (exact match) but we consider 3/4 good enough
+    if (normalizedDist >= 3 / 4)
```
olovy, line 220: :star:

whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
kwahlin: Structurally I think it should be ok. Since we exclude all properties having more than one value, we are sure to not overwrite any additional values upon merging. Although I'm not sure whether certain values that have a "better" syntax should take precedence upon conflicts.

whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
kwahlin: How do mean generalized (at what level)?

kwahlin:
Agree about the "canonical" form, I'll do that next time!
Thought it would be less confusing if I stuck to the marc terms since the matching logic was originally explained in those terms. Maybe I was wrong ;)

whelktool/scripts/cleanups/2021/04/lxl-1931-merge-series-membership.groovy
olovy: just something like

```
boolean aMatch(String a490, String a830) {
    return stringsMatch(a490, a830)
}

boolean stringsMatch(String a, String b) {
...
```

kwahlin: Ah ok, fair enough!

PR #896, opened 2021-04-19
Embellish auth records that are to be merged into a bib for export.

jannistsiroyannis:
Without this, the auth records lack important linked information, which
results in odd behaviour. See lxl-3524 for an example.


olovy (approved):
:ship:

klngwll (approved):
LGTM! Resulting bib now looks correct ⭐

PR #902, opened 2021-04-22
Fix a bug, which would allow definitions to bechanged via API.

jannistsiroyannis:
No record is ever allowed "change" it's collection, and API writes
only ever allow updates of auth, bib or hold. That _should_ have
been enough to prevent definitions updates, but there's a loop hole.

It turns out that the API makes it's judgement of collection based
on the contents of the update. So you can pass the API something
that looks like, say, a hold and pass the first check. But when
that presumed hold is then saved, there's no check that thing being
overwritten was indeed a hold. So in fact you can update a definition
as long the new data you're updating with doens't look like a
definition. This commit fixes the loophole by also checking that
the record about to be overwritten is in fact auth, bib or hold.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/Crud.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/Crud.groovy
@@ -732,6 +732,14 @@ class Crud extends HttpServlet {
                 String activeSigel = request.getHeader(XL_ACTIVE_SIGEL_HEADER)
 
                 if (isUpdate) {
+
+                    // You are not allowed to change definitions via API.
+                    String storedCollectionForId = whelk.storage.getCollectionBySystemID(doc.getShortId())
+                    if (!(storedCollectionForId in ["auth", "bib", "hold"])) {
+                        log.warn("Refused API update of document previously stored with collection $storedCollectionForId")
+                        return null
+                    }
+
                     String ifMatch = CrudUtils.cleanEtag(request.getHeader("If-Match"))
                     log.info("If-Match: ${ifMatch}")
                     whelk.storeAtomicUpdate(doc, false, "xl", activeSigel, ifMatch)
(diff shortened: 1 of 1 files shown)


andersju:
I'm not very familiar with the Crud code but if a record is never allowed to change its collection, wouldn't simply `if (collection != whelk.storage.getCollectionBySystemID(doc.getShortId()))` be appropriate? Because from your description and the code it sounds like a record's collection can still change, just not to anything other than auth/bib/hold?

jannistsiroyannis:
> I'm not very familiar with the Crud code but if a record is never allowed to change its collection, wouldn't simply `if (collection != whelk.storage.getCollectionBySystemID(doc.getShortId()))` be appropriate? Because from your description and the code it sounds like a record's collection can still change, just not to anything other than auth/bib/hold?

You're right. That would be better, I'll change it. The reason I first didn't go that way is:

Well, all of this starts out with loading a dataset, say bibdb. When that happens, they're all saved with collection (the column) = defintions.
But when you calculate "collection" on the data of those same documents, you get something else (this is why they could in fact be updated using the API). I first figured, If i did what you suggest, it would mean that no bibdb record coult ever be updated again at all. But that is incorrect thinking, since the dataset loader doesn't use the API.

Your way is better, I'll fix it.

rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -732,6 +732,13 @@ class Crud extends HttpServlet {
                 String activeSigel = request.getHeader(XL_ACTIVE_SIGEL_HEADER)
 
                 if (isUpdate) {
+
+                    // You are not allowed to change collection when updating a record
+                    if (collection != whelk.storage.getCollectionBySystemID(doc.getShortId()))
```
andersju, line 737:
```suggestion
                    if (collection != whelk.storage.getCollectionBySystemID(doc.getShortId())) {
```
Missing `{`! Otherwise LGTM

jannistsiroyannis:
Thanks Anders! I must have been tired yesterday :)

PR #903, opened 2021-04-26
Add simple user data API

andersju:
This adds a simple REST API for storing user data (e.g., user settings). It makes sure a user can only GET/PUT stuff concerning itself, and only PUT valid JSON, and not PUT anything exceedingly large, but doesn't care about anything else. SHA-256 of user email is used as ID. ID is verified through the user map populated by `AuthenticationFilter` (in case of valid user). Data is stored in a `jsonb` field. 

To use:

```bash
# Devops: fab xl_local app.whelk.run_db_migrations
# Start local environment. Sign in.
# Copy token (Firefox: Inspector -> Storage -> Local Storage -> key `at`)
LIBRIS_TOKEN=<copied-from-browser>
EMAIL_HASH=$(echo -n 'foo@example.com'|sha256sum|head -c 64)

# Put stuff:
curl -i -XPUT "http://localhost:8180/_userdata/${EMAIL_HASH}" \
  -H "Authorization: Bearer ${LIBRIS_TOKEN}" \
  -d '{"foo":"bar", "hej":"hopp"}'
# => HTTP/1.1 200 OK

# Get stuff: 
curl -i -XGET "http://localhost:8180/_userdata/${EMAIL_HASH}" \
  -H "Authorization: Bearer ${LIBRIS_TOKEN}"
# => {"foo": "bar", "hej": "hopp"}

# Put something else:
curl -i -XPUT "http://localhost:8180/_userdata/${EMAIL_HASH}" \
  -H "Authorization: Bearer ${LIBRIS_TOKEN}" \
  -d '{"foo":"bar", "hej":"hoppppppppp"}'
# => HTTP/1.1 200 OK

# Get something with wrong user hash:
curl -i -XGET "http://localhost:8180/_userdata/dskfjsklfsdfs" \
  -H "Authorization: Bearer ${LIBRIS_TOKEN}"
# => HTTP/1.1 403 ID in request doesn't match ID from token

# Put something with invalid token:
curl -i -XPUT "http://localhost:8180/_userdata/${EMAIL_HASH}" \
  -H "Authorization: Bearer hunter2" \
  -d '{"foo":"bar", "hej":"hopp"}'
# => HTTP/1.1 401 Missing access token.

# Put too much stuff:
dd if=/dev/urandom bs=2M count=1 \
  | curl -i -XPUT "http://localhost:8180/_userdata/${EMAIL_HASH}" \
      -H "Authorization: Bearer ${LIBRIS_TOKEN}" \
      --data-binary @-
# => [...]
# => HTTP/1.1 413 Too much data (length 2097115, max 1000000)

# Put invalid stuff:
curl -i -XPUT "http://localhost:8180/_userdata/${EMAIL_HASH}" \
  -H "Authorization: Bearer ${LIBRIS_TOKEN}" \
  -d 'sdfskfjsdfsdf'
# => HTTP/1.1 400 Invalid JSON
```

We could use a more typical solution with multiple tables, but given our usage/access patterns and what we need to store, a simple JSON store like this might be sufficient and more flexible.

Sending the hash of the email with the request is not strictly necessary (because `AuthenticationFilter` fetches that through the token), but `PUT /_userdata/<user-hash>` is a bit more RESTful than `PUT /_userdata`.

PUT rather than POST because it should be create-OR-update and idempotent.

`web.xml` modified to add not just the servlet but also another filter, since we need to use `AuthenticationFilter` for GET requests to `_userdata` (but not any other GET requests). I've never worked with servlets before so there might be a nicer way of doing that.

### DIFF
--- a/librisxl-tools/postgresql/migrations/00000020-add-user-data-table.plsql
+++ b/librisxl-tools/postgresql/migrations/00000020-add-user-data-table.plsql
@@ -1,0 +1,35 @@
+BEGIN;
+
+DO $$DECLARE
+   -- THESE MUST BE CHANGED WHEN YOU COPY THE SCRIPT!
+
+   -- The version you expect the database to have _before_ the migration
+   old_version numeric := 19;
+   -- The version the database should have _after_ the migration
+   new_version numeric := 20;
+
+   -- hands off
+   existing_version numeric;
+
+BEGIN
+
+   -- Check existing version
+SELECT version from lddb__schema INTO existing_version;
+IF ( existing_version <> old_version) THEN
+      RAISE EXCEPTION 'ASKED TO MIGRATE FROM INCORRECT EXISTING VERSION!';
+ROLLBACK;
+END IF;
+UPDATE lddb__schema SET version = new_version;
+
+-- ACTUAL SCHEMA CHANGES HERE:
+
+CREATE TABLE IF NOT EXISTS lddb__user_data (
+    id text PRIMARY KEY,
+    data jsonb NOT NULL,
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    modified timestamp with time zone DEFAULT now() NOT NULL
+);
+
+END$$;
+
+COMMIT;
--- a/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
@@ -1,0 +1,113 @@
+package whelk.rest.api
+
+import groovy.json.JsonException
+import groovy.util.logging.Log4j2 as Log
+import groovy.json.JsonSlurper
+import whelk.Whelk
+import whelk.util.WhelkFactory
+
+import javax.servlet.http.HttpServlet
+import javax.servlet.http.HttpServletRequest
+import javax.servlet.http.HttpServletResponse
+
+import java.util.stream.Collectors
+
+@Log
+class UserDataAPI extends HttpServlet {
+    private Whelk whelk
+    private static final int POST_MAX_SIZE = 1000000
+    private static final String ID_HASH_FUNCTION = "SHA-256"
+
+    UserDataAPI() {
+    }
+
+    UserDataAPI(Whelk whelk) {
+        this.whelk = whelk
+    }
+
+    @Override
+    void init() {
+        log.info("Starting User Data API")
+        if (!whelk) {
+            whelk = WhelkFactory.getSingletonWhelk()
+        }
+    }
+
+    @Override
+    void doGet(HttpServletRequest request, HttpServletResponse response) {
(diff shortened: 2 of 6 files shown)


olovy (commented):
I think the API should support DELETE.

Other than that, LGTM!

jannistsiroyannis:
Neat!

A few thoughts:
1. It looks like it's possible to hash your colleagues email instead of your own, and change their settings (not a bad practical joke). Or did I miss something preventing this?
2. This "feels" like XL and the "katalogiseringsgränsnitt" are finally being married, after years of pretending that they're independent. Now they're joined at the hip, period. Perhaps the day of the monorepo is finally at hand?

andersju:
>     1. It looks like it's possible to hash your colleagues email instead of your own, and change their settings (not a bad practical joke). Or did I miss something preventing this?

[This](https://github.com/libris/librisxl/blob/15bce17932e698484204495b13231a94bbcff117/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy#L101) prevents that: it checks whether `/_userdata/<email_hash>` matches the hash of the email that's tied to the token  (`userInfo` comes from `AuthenticationFilter` which in turn gets it from xl_auth).

>     2. This "feels" like XL and the "katalogiseringsgränsnitt" are finally being married, after years of pretending that they're independent. Now they're joined at the hip, period. Perhaps the day of the monorepo is finally at hand?

:+1:

jannistsiroyannis (approved):
I've had it explained to me and now I get it :) :ship:

andersju:
> I think the API should support DELETE.

Done!

```
curl -i -XDELETE "http://localhost:8180/_userdata/${EMAIL_HASH}" -H "Authorization: Bearer ${LIBRIS_TOKEN}"
HTTP/1.1 204 No Content
```

olovy (approved):
:ship: 
Nice!

lrosenstrom (approved):
LGMT! :ship:

andersju:
> > ```
> > 1. It looks like it's possible to hash your colleagues email instead of your own, and change their settings (not a bad practical joke). Or did I miss something preventing this?
> > ```
> 
> [This](https://github.com/libris/librisxl/blob/15bce17932e698484204495b13231a94bbcff117/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy#L101) prevents that: it checks whether `/_userdata/<email_hash>` matches the hash of the email that's tied to the token (`userInfo` comes from `AuthenticationFilter` which in turn gets it from xl_auth).

(Addendum: the user-supplied email isn't actually used in get/put/delete; the one injected by AuthenticationFilter is currently always used. So it's just to be able to say "you're trying to do something you shouldn't" (and have a cleaner API), and be open to the possibility of users with certain permissions being able to change other users' settings in the future, perhaps.)

(Should also add some tests...)

olovy:
@oBlissing

PR #904, opened 2021-04-26
Search: Faster assembly of result by reducing number of DB calls

olovy:
When searching, for each request, hundreds of chips are loaded for e.g facets. This batches the DB requests and loads cards instead of full docs. ~40% faster when running locally on my machine.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
@@ -1,5 +1,7 @@
 package whelk.rest.api
 
+import com.google.common.collect.ArrayListMultimap
+import com.google.common.collect.Multimap
 import com.google.common.escape.Escaper
 import com.google.common.net.UrlEscapers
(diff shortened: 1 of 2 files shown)


jannistsiroyannis (approved):
I'm tired so proably not thinking straight.
I get that loading cards would be faster than full docs, but why would batching make a difference here (it's still the same number of queries right)? Is the idea to bypass the connectionPool (i thought that was supposed to be super cheap)?
Still, 40% is 40% :ship:

olovy:
> I'm tired so proably not thinking straight.
> I get that loading cards would be faster than full docs, but why would batching make a difference here (it's still the same number of queries right)? Is the idea to bypass the connectionPool (i thought that was supposed to be super cheap)?
> Still, 40% is 40% ship

No it's (best case) two queries instead of 700 =)

lrosenstrom (approved):
Nice :ship:

andersju (approved):
:racing_car: :ship:

PR #905, opened 2021-04-27
Document (types) no longer need to belong to a legacy MARC collection

olovy:
Set collection `none` for documents that don't belong to any of the legacy MARC collections (bib, auth, hold) or definitions.
So that we can store documents with other types e.g. `DataCatalog`

I've checked all the places where we look at collection and this shouldn't break anything.


jannistsiroyannis (approved):
:ship:

PR #908, opened 2021-04-29
Feature/lxl 1931 merge series membership 2

kwahlin:
The main improvement in this version is safe merging. In the prior we risked overwriting stuff.
We now also:
- remove exact duplicates
- compare any two SeriesMembership entities instead of only when they have seriesStatement and mainTitle respectively
- if a SeriesMembership contains multiple values we leave out only this entity from comparison instead of all entities in seriesMembership
- use bibframe terminology instead of marc


olovy (approved):
LGTM! :star:

PR #909, opened 2021-04-29
Add script for jaktskildringar.

jannistsiroyannis:


olovy (commented):
LXL-3536 was a bit vague but as I understood it we also want to change subject: sao/Jaktskildringar to genreForm: saogf/Jaktskildringar?

gernreForm: sao/Jaktskildringar  are only 5 records [in PROD](https://libris.kb.se/katalogisering/search/libris?o=https%3A%2F%2Fid.kb.se%2Fterm%2Fsao%2FJaktskildringar) while  subject: sao/Jaktskildringar to genreForm are 79. 

ComplexSubject termComponent (45) I guess we can't do nothing about.

jannistsiroyannis:
:+1: I've checked with Viktoria and its _only_ subject we're supposed to change. I'll fix it.

klngwll (commented):
I do agree the description of the issue is a bit vague. In general we don't have GenreForm terms in subjects so I would have guessed also the relation should have changed to genreForm.

klngwll (changes requested):
So what I figured out is that the relationship for these should be changed from subject to genreForm as well.

jannistsiroyannis:
> So what I figured out is that the relationship for these should be changed from subject to genreForm as well.

Better now? :)

whelktool/scripts/cleanups/2021/04/lxl-3536-jaktskildringar.groovy
```diff
@@ -0,0 +1,64 @@
...
+    if (containedBadLink) {
+
+        // Make sure genreForm is a list
+        if (work.genreForm == null)
+            work.put("genreForm", [])
+        else if ( !(work.genreForm instanceof List) )
+            work.put("genreForm", [work.genreForm])
+
+        // Add the correct link
+        work.genreForm.add(["id" : newLinkTarget])
```
olovy, line 43:
```suggestion
        work.genreForm.add(["@id" : newLinkTarget])
```

olovy (commented):
Almost!

The rest LGTM

klngwll (approved):
LGTM

jannistsiroyannis:
> Almost!
> 
> The rest LGTM

Good catch! :)

PR #910, opened 2021-04-29
Respect ES max_result_window in makePaginationLinks, and include it in response

andersju:
Fetches Elasticsearch `max_result_window` from index settings and uses it when generating pagination links. This fixes "next" and "last" links (can no longer go beyond what's possible), but some adjustments in the client are still necessary to make it not generate links beyond the limit, so a `maxItems` is now included in the response to make this possible.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
@@ -198,6 +198,8 @@ class SearchUtils {
             result['_debug'] = esResult['_debug']
         }
 
+        result['maxItems'] = esQuery.getMaxItems().toString()
+
         lookup.run()
         
         return result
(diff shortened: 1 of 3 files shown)


olovy (approved):
LGTM!

We should add maxItems to vocab, [here somewhere](https://github.com/libris/definitions/blob/b4b4ce2332e11c8ce83d1fea97ce81473109079d/source/vocab/platform.ttl#L282). I haven't found any matching property in other vocabularies. 

We are maybe breaking the opensearch definition of [opensearch:totalItems](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-totalresults-element)?

Hydra documentation is so bad that is hard to say [hydra:totalItems](https://www.hydra-cg.com/spec/latest/core/#hydra:totalItems).

PR #912, opened 2021-05-04
Compute _sortKeyByLang on framed document

olovy:
So that data added by embellish is available.
e.g. library name and instance title for holdings.


andersju (approved):
:ship:

PR #913, opened 2021-05-05
Allow GET /_remotesearch only when authenticated

andersju:
Previously, GET requests to /_remotesearch weren't actually authenticated. Now they are.


olovy (approved):
LGTM!

lrosenstrom (approved):
:ship:

PR #914, opened 2021-05-06
Add more validation for auth and bib

andersju:
- Makes it possible for JsonLdValidator to check a set of different scopes depending on legacy collection. Figured this would be the easiest way (it's just "temporary"). Makes POST/PUT in Crud send collection type along with the data to the validator.
- For auth: enables MISSING_DEFINITION, OBJECT_PROPERTIES, UNKOWN_VOCAB_VALUE
- For bib: enables MISSING_DEFINITION, OBJECT_PROPERTIES

OBJECT_PROPERTIES corresponds to the errors OBJECT_TYPE_EXPECTED and VOCAB_STRING_EXPECTED.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
@@ -6,6 +6,20 @@ import whelk.util.DocumentUtil
 class JsonLdValidator {
     private static JsonLd jsonLd
     private Collection skipTerms = []
+    private Map<String, Set> scopesPerLegacyCollection = [
+            "all": [
+                    Validation.Scope.NESTED_GRAPH,
+            ],
+            "auth": [
+                    Validation.Scope.MISSING_DEFINITION,
+                    Validation.Scope.OBJECT_PROPERTIES,
+                    Validation.Scope.UNKNOWN_VOCAB_VALUE,
+            ],
+            "bib": [
+                    Validation.Scope.MISSING_DEFINITION,
+                    Validation.Scope.OBJECT_PROPERTIES,
+            ],
+    ]
 
     private JsonLdValidator(JsonLd jsonLd) {
         this.jsonLd = jsonLd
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
```diff
@@ -41,9 +59,26 @@ class JsonLdValidator {
...
+
+        // Scopes that all things should have
+        scopes.addAll(scopesPerLegacyCollection['all'])
+
+        // Scopes that only some types of things should have (for the moment!)
+        if (collection && collection in scopesPerLegacyCollection) {
+            scopes.addAll(scopesPerLegacyCollection[collection])
+        }
+
+        return scopes.inject([], { errors, value ->
```
jannistsiroyannis, line 76: This is fine! But I had to go look it up, and it sort of hurt my head ;(
There are so many annoying ways to be clever in groovy.

jannistsiroyannis (approved):
:ship: good to get some more validation activated!

lrosenstrom:
Nice! A suggestion though would be to run doValidate once instead of once per scope. Now we traverse the graph and prevalidate everything several times per record. The scopes could be e.g. AUTH, BIB, ALL, or alternatively we could keep a list of scopes for the Validation object. Then we can decide what methods to call for a particular scope in the newly added switch/case!

whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
andersju: Was actually thinking that too. I'll make it more readable in the next iteration :)

andersju:
> Nice! A suggestion though would be to run doValidate once instead of once per scope. Now we traverse the graph and prevalidate everything several times per record. The scopes could be e.g. AUTH, BIB, ALL, or alternatively we could keep a list of scopes for the Validation object. Then we can decide what methods to call for a particular scope / set of scopes in the newly added switch statement!

Good idea, thanks! I'll cook up a new version.

andersju:
Simplified things and fixed wastefulness. This is assuming that we don't need "scope" to be a specific validation, but a set of validations like "all validations", "default validations", "default + collection-specific validations".

PR #919, opened 2021-05-12
Add findKey for multiple keys to DocumentUtil

olovy:
Old version only takes one key
`static boolean findKey(data, String key, Visitor visitor)`

Inspired by `clearGarbageProperties` in #918.


jannistsiroyannis (approved):
LGTM!

PR #923, opened 2021-05-19
Handle reindex getAffectedIds > ES max_result_window

olovy:
We find documents that need to be reindexed by querying _links and _outerEmbellishments.
This fails if the number of found documents exceed the max_result_window setting of the ES index (currently 100 000).

Fall back to using [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) and [Point in time API](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html) if number of results > max_result_window.


andersju (approved):
Nice work! LGTM!

PR #926, opened 2021-05-20
Add "heuristic linker" document normalizer

olovy:
Implemented an old idea from https://github.com/libris/librisxl/pull/591. 
Configure types and properties for "BlankNodeLinker" normalizers with vocab instead of hardcoding them.

Link blank nodes based on "heuristic identifiers"
e.g. `{ "@type": "Role", "label": "Þýðandi" }` matches https://id.kb.se/relator/trl on `prefLabelByLang.is`

For all types that have :category :heuristicIdentity in vocab:
Link all blank nodes with that `@type` that match on a property that has `:category :heuristicIdentifier`.
Only check blank nodes in properties where `@type` is in range (`range` or `rangeIncludes`).

Depends on https://github.com/libris/definitions/pull/284

First use case is `country` in entities from bibdb.
Replace the hardcoded linker for `Role`

### DIFF
--- a/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
+++ b/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
@@ -14,26 +14,14 @@ import static whelk.JsonLd.ID_KEY
 
 /*
 TODO: add support for linking blank nodes based on owl:hasKey
-example (only hasKey defined in vocab at the moment): 
-{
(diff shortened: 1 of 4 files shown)


olovy:
Also handle `isReplacedBy` in `BlankNodeLinker`

olovy:
We could of course also tag properties ('code', 'label', 'prefLabelByLang', 'altLabelByLang', 'hiddenLabel') with e.g. `heuristicIdentitifier` and make it completely generic?

PR #929, opened 2021-05-26
Search: Handle ORCID with spaces

olovy:
This is a follow-up on 1de4449c3d4e4a5397291eaefe34cb6d128213af and f286c2c17a1f5cf20db8f9210bc4d6fe3e3d009c which I accidentally pushed to `develop`...

ISNIs are always 16 digits and can be displayed with out without grouping of digits. ORCID is a reserved subset of ISNI.

Handle ISNI and ORCID, formatted with or without spaces in search
- i.e. both 0000000084029206 and 0000 0000 8402 9206 works
- but 0000 8402 0000 9206 will also match because identifiedBy.value is a text fields in elastic so every group is treated as a word (token)...

Normalize ISNI and ORCID on save, always store without spaces


klngwll (approved):
LGTM!

PR #930, opened 2021-05-26
Fix MARC conversion of ISNI/ORCID/VIAF/WikidataID

olovy:
* Fix MARC conversion of ISNI/ORCID/VIAF/WikidataID
* Map to `@type` instead of `@type`: Identifier with typeNote for these identifiers
* Add a document normalizer that replaces all obsolete typeNotes (e.g. cataloging according to old practise)
* Add a script for converting existing auth records


klngwll (approved):
LGTM! ⭐

PR #932, opened 2021-05-26
Avoid null values in lddbToTrig; clean up IRIs in JsonLdToTurtle

andersju:
vcopyImporter's `lddbToTrig` can produce invalid triples due to bad data. We should obviously fix these broken records (some are even hardcoded in `trigExcludeList` at the moment); however, we might also want to use the output of `lddbToTrig` as input to [Apache Jena](https://jena.apache.org/). Apache Jena's behavior is to completely stop processing when an invalid triple is encountered -- skipping is not an option (some discussion [here](https://users.jena.apache.narkive.com/srym8jU4/getting-rid-of-triples-with-bad-uris)).

Loading/updating SPARQL data is something that should be done routinely and that preferably shouldn't involve manual "fix data - restart import" interventions every now and then, à la sisyphean whack-a-mole, so I think we should make some effort to avoid outputting completely invalid things.

I've been trying to import the entire lddb-as-trig dataset into Jena, but encountered some errors. Some records have `null` values that make it through the converter and end up in the TriG version (e.g. [4nggh9tg3l3kgbp](https://libris.kb.se/4nggh9tg3l3kgbp#it), [q26xxwhfn63n4jkd](https://libris.kb.se/q26xxwhfn63n4jkd#it)), some have IRIs with [illegal](https://www.w3.org/TR/turtle/#grammar-production-IRIREF) characters (e.g. [x67fk77lvckfnf4z](https://libris.kb.se/x67fk77lvckfnf4z#it), [r128d23zpmxvb662](https://libris.kb.se/r128d23zpmxvb662#it) -- `https://id.kb.se/marc/GovernmentPublicationType-\\`).

This small change just removes the problem-causing nulls when using `lddbToTrig` and removes illegal characters from IRI references (while logging a warning) in `JsonLdToTurtle`. There might be more things to fix, we'll see.

### DIFF
--- a/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
+++ b/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
@@ -186,6 +186,7 @@ class ImporterMain {
                 filterProblematicData(data[property])
             }
         } else if (data instanceof List) {
+            data.removeAll([null])
             data.each {
                 filterProblematicData(it)
             }
--- a/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
+++ b/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
@@ -159,7 +159,7 @@ class ImporterMain {
                     System.err.println("$i records dumped.")
                 }
                 ++i
-                filterProblematicData(doc.data)
+                filterProblematicData(id, doc.data)
                 try {
                     serializer.objectToTrig(id, doc.data)
                 } catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


olovy (approved):
:ship: LGTM!

We should clean (or block) broken JSON-LD on the way in too, e.g. null values in lists.

klngwll (approved):
LGTM, a sound approach to not grind everything to a halt, yet also giving us feedback for better conversion handling and/or validation.

andersju:
Encountered more errors and realized that we need more proper validation. I noticed that we're actually already including `apache-jena-libs` as a dependency, and it includes a handy [IRI library](https://jena.apache.org/documentation/notes/iri.html), so let's use it! Now we try to fix bad IRIs a bit, but if they're broken beyond (sensible) repair they're replaced with `https://BROKEN-URI/`.

Also, changed it so that things with `null` values have their document ID included in the logged warning.

lrosenstrom (approved):
LGTM!

PR #933, opened 2021-05-26
Reindex: Only use Point in time API if available

olovy:
Apparently it isn't included in the OSS version of ES.
I happened to have the "default" version installed when testing this
locally...

https://github.com/elastic/elasticsearch/issues/65906

It isn't needed to be able to use "search_after" but we don't
get a snapshot of the index state when looking for docs to reindex.

Maybe we should remove the point in time code completely.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -41,6 +41,7 @@ class ElasticSearch {
     private String elasticCluster
     private ElasticClient client
     private ElasticClient bulkClient
+    private boolean isPitApiAvailable = false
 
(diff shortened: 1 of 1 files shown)


andersju (approved):
LGTM! *But*, 7.10 was the very [last](https://www.elastic.co/pricing/faq/licensing) elasticsearch-oss, and 7.11 the last version with a distinction between the "free / basic" (default) and "basic" (formerly oss) versions. Meaning the Point In Time API *is* included in the only version that's now publicly available. So if we stick with Elastic we'll have to use the free-but-not-actually version (which includes PIT) at some point, if we ever want to upgrade. (I noticed that the [7.13](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html) documentation dropped the "X-Pack" tag)

On a related note Amazon's Elasticsearch 7.10 fork is now live at https://www.opensearch.org/ (but no PIT obviously...)

PR #935, opened 2021-05-27
Add missing matchUriTokens

klngwll:
Add matchUriTokens where missing to prevent linking unknown values like `https://id.kb.se/marc/GovernmentPublicationType-\\`. Also removed erroneous match on "_".

We could still need a more finetuned way of filtering bad data to not slip useless codes into the descriptions but at least they will not be represented as faulty URI:s


olovy (approved):
:ship:

PR #938, opened 2021-05-27
Add empty fixedDefault for serial 008/21 and spec

klngwll:
Producing blank field in mandatory position for Type of continuing resource in 008/006 for serials instead of "|", as per ISSN-IC MARC profile:
```
Codes used:  
d: updating database  
l: updating loose-leaf  
m: monographic series  
n: newspaper  
p: periodical  
w: updating Web site  
#: none of the above types

Notation “#” means that the field should remain blank. The hash “#” should never be used in the MARC record.
```
https://www.issn.org/wp-content/uploads/2016/11/ISSN-MARC-21-ENG-Revised-September-2016.pdf


olovy (approved):
:ship:

PR #946, opened 2021-06-04
Don't update changedBy on minor (silent) updates

andersju:
Assumes `changedBy` should _never_ be updated if it's a minor update (so affects more than just whelktool).


olovy (approved):
:ship:

PR #951, opened 2021-06-09
Feature/lxl 2938

jannistsiroyannis:


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -1826,6 +1832,22 @@ class PostgreSQLComponent {
         }
     }
 
+    boolean iriIsLinkable(String iri, Connection connection) {
+        PreparedStatement preparedStatement = null
+        ResultSet rs = null
+        try {
+            preparedStatement = connection.prepareStatement(GET_IRI_IS_LINKABLE)
```
olovy, line 1839: connection is null here if called from `void normalizeIdentifiers(Document document)` ?

olovy (changes requested):
See comment about NPE.

olovy:
I guess this is potentially quite slow, since we get two extra DB calls per link?
(But I think correctness is top priority)

olovy (approved):
:ship:

PR #955, opened 2021-06-11
Add documentation for or- in search-API

olovy:
Add documentation for `or-` in search-API. Started with Swedish version for review. Will translate if OK.
Not super happy about this. Can it be made more clear?


jannistsiroyannis (approved):
I think it's fine! It's complicated, but that's because it _is_ complicated (with URLsyntax).

andersju (approved):
+1 on what Jannis said

olovy:
Ålright, I think I will add an example too.

PR #956, opened 2021-06-15
Remove extra whitespace from 001/003 when creating 035$a

jannistsiroyannis:
OCLC sometimes deliver records where 001 carries an extra
space character at the end. XLs code did not expect this and
as a result created extra (double) 035$a-fields.

This commit trims values from 001/003 before using them to
construct (and double check) 035$a.


kwahlin (approved):
:+1:

olovy (approved):
:ship:

PR #957, opened 2021-06-24
Add a safety margin to export time intervals.

jannistsiroyannis:
Without this, the only safety margin is the network transport
time, which can be very small. The consequence is that a client
could miss being told of records that have changed.

Clients should always use NTP, but even then one second diffs
could easily occur, due to rounding.

### DIFF
--- a/marc_export/marc-export-documentation.md
+++ b/marc_export/marc-export-documentation.md
@@ -16,12 +16,15 @@ Dessa parametrar är:
    1. `append` som innebär att borttagna posters ID:n exporteras som en CSV-fil _efter_ den vanliga exportdatan (separerat av en null-byte).
 1. `virtualDelete` som kan ha värde `true` eller `false`. Är `virtualDelete` satt till `true` så kommer poster anses vara borttagna i den genererade exporten, i de fall där de sigel som anges i `locations` i profilen inte längre har bestånd på posterna. Flaggan används förslagsvis tillsammans med `deleted=export`.
 
-Exempel på anrop:
+## Exempel på anrop
 ```
 $ curl -Ss -XPOST "https://libris.kb.se/api/marc_export/?from=2019-10-05T22:00:00Z&until=2019-10-06T22:00:00Z&deleted=ignore&virtualDelete=false" --data-binary @./etc/export.properties > export.marc
 ```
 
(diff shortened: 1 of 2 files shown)


andersju (approved):
LGTM!

olovy:
I don't follow the reasoning behind this. The `from` and `until` parameters are server/system time (?). So the client's view of the current time should not matter?

andersju:
> I don't follow the reasoning behind this. The `from` and `until` parameters are server/system time. So the client's view of the current time should not matter?

From the email thread I believe the reasoning is:

It's 12:59:59 (actual and server time), but client thinks it's 13:00:00. Client asks server for records updated between 12:00-13:00. Server returns those records. Afterwards, at 12:59:59.5 (actual time), some records are updated. When client later asks for records updated between 13:00-14:00, those few records won't be included.

olovy:
> > I don't follow the reasoning behind this. The `from` and `until` parameters are server/system time. So the client's view of the current time should not matter?
> 
> From the email thread I believe the reasoning is:
> 
> It's 12:59:59 (actual and server time), but client thinks it's 13:00:00. Client asks server for records updated between 12:00-13:00. Server returns those records. Afterwards, at 12:59:59.5 (actual time), some records are updated. When client later asks for records updated between 13:00-14:00, those few records won't be included.

Ah, ok.
I think it would be cleaner to add a note on usage in the API documentation and implement this on the client side. 
But in practice, perhaps this solution will lead to less missed records.

This will to lead to some surprising results since consecutive intervals will overlap by one second.
But duplicate exports/imports shouldn't be a problem. Some statistics somewhere might get a bit off.
Intervals were already slightly overlapping since both from and until are inclusive.

Let's describe the extra second in the API docs and :ship: .

olovy (approved):

jannistsiroyannis:
> I think it would be cleaner to add a note on usage in the API documentation and implement this on the client side.
> But in practice, perhaps this solution will lead to less missed records.

Indeed! I feel the same unease with this! But you would not belive the number of times I've had to explain to users of this API, that they can't just input local times and call them "utc" and get correct results (even though it says to watch carefully for this in the documentation).

But you're of course right that we need to document the fact that we're fiddling with the input. I'll add that!

PR #958, opened 2021-06-29
Add script for adding identifiers with mimer packet uris

kwahlin:


klngwll (approved):
🚢

whelktool/scripts/cleanups/2021/06/lxl-3640-add-mimer-uri-identifiers/script.groovy
```diff
@@ -0,0 +1,57 @@
...
+}
+
+selectByIds(mimerUriById.keySet()) { data ->
+    Map instance = data.graph[1]
+    List identifiedBy = instance.identifiedBy
+
+    Map identifier =
+            [
+                "@type": "Identifier",
+                "typeNote": "uri",
```
niklasl, line 36: Is this `typeNote` to get desired MARC out of this? Otherwise I'd say `"@type": "URI"` here (and skip the `typeNote`).

klngwll: I had actually forgot we added the URI class. Should we "unpend" it? :)
We could add "uri" code to the normalization we have in place as well.

kwahlin: Very reasonable, fixed!

PR #960, opened 2021-07-01
marc2bfConversionUpdate3xx-v1.6

klngwll:
**Description**

Align with LC marc2Bibframe2 conversion v1.6

- 300: Ignore $6 (Small occurence in "extent", but this needs further analyzing because they could be wrong. All $6 was added to extent in conversion even though other properties are possible, like hasDimensions, hasNote etc.).
- 306: Ignore $6 (No occurence in "duration").
- 310: Add $0 (controlNumber), $2 (source), Ignore $6  (No occurence in "frequency").
- 321: Add $0 (controlNumber), $2 (source), Ignore $6 (No occurence in "replaces").

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -11285,9 +11285,9 @@
       },
       "$a": {"about": "_:frequency", "property": "label"},
       "$b": {"about": "_:frequency", "property": "date"},
-      "$0": {"property": "marc:recordControlNumber"},
+      "$0": {"about": "_:frequency", "property": "marc:recordControlNumber"},
       "$1": null,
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
LGTM! For my understanding, what does the null value say/do in this context, e.g. "$1": null?

klngwll:
> LGTM! For my understanding, what does the null value say/do in this context, e.g. "$1": null?

It used to drop fields if I remember correctly (it still do for fixed fields and indicators). Now we use ignore to do that properly and for fields and subfields now just a way to say we don't do anything (probably more like LC:s nac), ultimately we are left with a marcUncompleted. It would be a good idea to revise the usage of 'null' down the road since 'ignored' used to be the marcUncompleted one. We switched because with ignored we can easier leave a rationale for why we drop the field.

PR #962, opened 2021-07-14
Feature/lxl 3526 wikidata links

kwahlin:
The changes to whelktool/build.gradle are copied from whelk-core/build.gradle. xml-apis is necessary for the Jena libraries to work. Otherwise see LXL-3526


klngwll (approved):
Good work! Can't comment on the code details but have checked some of the statistical analysis and will check the result once we do a QA run of this ✨

olovy (approved):
:ship:

PR #964, opened 2021-07-22
Add support for ES autocompletion/suggestions and source filtering

andersju:
We want to provide search suggestions a.k.a. autocompletion a.k.a. search-as-you-type when people search on id.kb.se. This needs to be reasonably fast with low overhead since each keystroke triggers a new search. It needs to feel instant(-ish) to the user.

There are a few ways to do this with Elasticsearch, the most obvious ones being the [suggest](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html) feature's [completion suggester](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#completion-suggester) and the [`search_as_you_type`](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-as-you-type.html) field type. The completion suggester should be the fastest, but it's limited - we can't use filters, for example. And in this case we just want autocompletion on a small subset of the index. The [context suggester](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#context-suggester) can help with this, but would be pretty messy for our purposes.

So, after experimenting, I found that the `search_as_you_type` field used with a [`match_phrase_prefix`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase-prefix.html) query seemed to give the most decent results in the least hackish way. I added a field called `suggest` to `prefLabel` in `libris_config.json`, so the index needs to be recreated.

A new query parameter, `_suggest`, is added. The value should be the field on which suggestion/autocompletion should be performed. If specified, a `match_phrase_prefix` query is performed on the specified field, and no boosting is done. Example:

```
curl -XGET -H "Accept: application/ld+json" \
      'http://kblocalhost.kb.se:5000/find.jsonld?q=musik&_limit=10&inScheme.@id=https://id.kb.se/term/sao&_suggest=prefLabel.suggest'
```

Additionally, to reduce network overhead, a `_fieldsToInclude` parameter is added. It takes a comma-separated list of fields. By default ES returns the entire JSON document that was indexed, but if `_fieldsToInclude` is specified then [source filtering](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#source-filtering) is used to only retrieve the selected fields.

The query above resulted in 67662 bytes of JSON when I tested locally. But for autocompletion we're only interested in a few fields - perhaps just `@id`, `prefLabel` and `inScheme.titleByLang`. So let's get only those fields:

```
curl -XGET -H "Accept: application/ld+json" \
      'http://kblocalhost.kb.se:5000/find.jsonld?q=musik&_limit=10&inScheme.@id=https://id.kb.se/term/sao&_suggest=prefLabel.suggest&_fieldsToInclude=@id,prefLabel,inScheme.titleByLang'
```

Now it's just 8955 bytes of JSON, or about 87% smaller. Still, 2/3 of that is taken up by the `stats` and `search` objects that we don't need for autocompletion, so we can do better (but that requires adding a few special cases which might or might not be worth it).

(`_fieldsToInclude` might also be useful in other situations where we are only interested in a few specific fields.)

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -238,6 +238,10 @@
                     "keyword": {
                         "type": "icu_collation_keyword",
                         "language": "sv"
+                    },
+                    "suggest": {
+                        "type": "search_as_you_type",
+                        "analyzer": "softmatcher"
                     }
                 },
                 "copy_to": "_all"
--- a/rest/API.md
+++ b/rest/API.md
@@ -183,6 +183,8 @@ means `OR`, `*` is used for prefix queries, `""` matches the whole phrase, and
   Default is 200.
 * `_offset` - Number of hits to skip in the result, used for pagination.
   Default is 0.
+* `_fieldsToInclude` - Only include the specified fields (comma-delimited) in the results.
+  If not set, all fields are returned.
 
 Records can be filtered on field values. Specifying multiple values for the same field can be done
 by repeating the parameter or by giving a comma-separated list as value. Specifying multiple fields 
@@ -285,6 +287,16 @@ Containing 'Aniara' and held by sigel APP1.
 ```
 $ curl -XGET -H "Accept: application/ld+json" \
     'https://libris-qa.kb.se/find.jsonld?q=Aniara&@reverse.itemOf.heldBy.@id=https://libris.kb.se/library/APP1'
+...
+```
+
+#### Example
+
+Instances containing 'Jansson'. Only the fields `hasTitle` and `identifiedBy`
+are included in each returned item.
+```
+$ curl -XGET -H "Accept: application/ld+json" \
+    'https://libris-qa.kb.se/find.jsonld?q=Jansson&@type=Instance&_fieldsToInclude=hasTitle,identifiedBy'
 ...
 ```
 
--- a/rest/API_sv.md
+++ b/rest/API_sv.md
@@ -188,6 +188,8 @@ innebär `ELLER`, `*` används för prefixsökningar, `""` matchar hela frasen o
   paginering. Standardvärdet är 200.
 * `_offset` - Antal träffar att hoppa över i resultatet, används för
   paginering. Standardvärdet är 0.
+* `_fieldsToInclude` - Inkludera enbart angivna fält (kommaseparerade) i  resultatet.
+  Om ej angivet returneras alla fält.
   
 Sökningen kan filtreras på värdet på egenskaper i posten. Samma egenskap kan anges flera gånger genom
 att upprepa parametern eller genom att komma-separera värdena. Om olika egenskaper anges innebär det 
@@ -289,6 +291,16 @@ Innehåller 'Aniara' och har ett bestånd med sigel APP1.
 ```
 $ curl -XGET -H "Accept: application/ld+json" \
     'https://libris-qa.kb.se/find.jsonld?q=Aniara&@reverse.itemOf.heldBy.@id=https://libris.kb.se/library/APP1'
+...
+```
+
+#### Exempel
+
+Instanser som innehåller 'Jansson'. Endast fälten  `hasTitle` och `identifiedBy`
+inkluderas i resultatet för varje post.
+```
+$ curl -XGET -H "Accept: application/ld+json" \
+    'https://libris-qa.kb.se/find.jsonld?q=Jansson&@type=Instance&_fieldsToInclude=hasTitle,identifiedBy'
 ...
 ```
 
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -23,7 +23,7 @@ class ESQuery {
     private static final ObjectMapper mapper = new ObjectMapper()
     private static final int DEFAULT_PAGE_SIZE = 50
     private static final List RESERVED_PARAMS = [
-        'q', 'o', '_limit', '_offset', '_sort', '_statsrepr', '_site_base_uri', '_debug', '_boost', '_lens'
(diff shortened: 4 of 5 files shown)


niklasl:
Good progress! I would like to eliminate some moving parts and leaking internals here if possible. Might we get by by switching over to this mode of searching automatically when `&lens=chip` (`chips`?) and a `@type` is given? Given that e.g. the chip definition for `Concept` includes only `"showProperties": [ "prefLabel", "label", "code", "termComponentList", "inScheme" ]`, that's pretty close.

That in turn makes me wonder if the internal `_text` key (we still add that, right?) can be tinkered with to have an extra `.suggest` instead, thus eliminating the need for source filtering, not requiring (so much) more reindexing if we want to suggest-query on e.g. agents, and just "suggest"-query on that `_text` value if the `_lens=chip` and `@type` parameters are given. If so, we can also drop at least `stats` when `_lens=chip` is given (since one is "obviously enough" looking for a reduced result payload if one only wants chips).

andersju:
These sound like good and doable ideas, thanks for the feedback! Would certainly be cleaner. We do add the `_str` field still (and `_str.exact`) in ES. So adding `.suggest` along with `.exact` is one option. One problem is that `_str` includes a bunch of stuff that's not in the chip. For example, the top-level `_str` of https://libris-qa.kb.se/gzrqmngs4jfv1x9#it is

"96 s. Vad du kan lära dig om livet av din katt 10 spiritual lessons you can learn from your cat. Kina 2005 Valentin Västerljung ISBN 91723312 83 inb. 17 cm"

However, in #888 we added a new ES field for sorting purposes, to (reasonably) accurately reflect what the user sees when browsing, based on the actual chip:
```
    "_sortKeyByLang": {
      "sv": "Vad du kan lära dig om livet av din katt",
      "en": "Vad du kan lära dig om livet av din katt"
    }
```
This field should work well for autocompletion purposes, so maybe it's best to add a `.suggest` field to `_sortKeyByLang.sv` and `_sortKeyByLang.en` (as a bonus this would also make it possible to do per-language autocompletion, if we ever want that). (In retrospect `_sortKeyByLang` was perhaps not the ideal name - "naming things" and all that...).

niklasl:
Ah, yes, that sounds even better! I see about `_str` being overeager; we might want to fine-tune these if we want them to converge. But go for `_sortKeyByLang` for now, I'd say.

Naming things indeed, but at least it's clear what it means (`_str` is even worse; I named it and can't even remember its name :see_no_evil: ). And they're all `_private` enough. We don't want them exposed in the API.

andersju:
Alright, I pushed another stab at it. Sample query:
```
curl -XGET -H "Accept: application/ld+json" \
  'http://kblocalhost.kb.se:5000/find.jsonld?q=musik&_lens=chips&@type=Concept&_limit=10'
```

niklasl (commented):
I like this a lot! Clean changes too. I can try it out more in depth tomorrow; for now just a quick question: is the `.sv` value picked from non-language-strings as well (e.g. regular `prefLabel` and `hasTitle.mainTitle`)? If that's the case it might be logically incorrect and a hidden source of future trouble, but we can investigate that separately. Apart, of course, that this feature now depends on that structure. Not publicly though, so as far as design goes I think this is good!

andersju:
If there is a language string (e.g. `prefLabelByLang.{sv,en}`), then that one is used instead of the regular `prefLabel`. If there isn't, both `.sv` and `.en` will use the regular version. This is what lxlviewer does, so it was necessary to attempt to do the same in ES to be able to alphanumerically sort things in a way that would make sense to the end user (i.e. correspond to what the end user sees as the "title" of a thing in the browser). So `.en` here isn't necessarily in English, but simply what a user, having selected English as the UI language, would see :)

The actual logic is [here](https://github.com/libris/librisxl/blob/develop/whelk-core/src/main/groovy/whelk/JsonLd.groovy#L747-L834) (`toChipAsListByLang` and `toChipAsMapByLang`); some examples [here](https://github.com/libris/librisxl/pull/888).

niklasl:
I see; that makes a lot of sense! Thanks for the references too (saved me time being away from a desk at the moment)!

I believe the whole "search with language or not" question is quite open in the RDF/SPARQL community at large, and AFAIK it's not uncommon to store doubles (e.g. indexing plain strings next to their lang-tagged sources). That's somewhat the inverse of this, but since it's internal state it only matters if more languages are to be used, or for language-aware search as you've mentioned. And either way would need to be adapted. Good to talk about further in the future, but this PR works perfectly well as implemented!

niklasl (approved):
LGTM!

olovy (commented):
I'm a bit late to the party here but I'll add my 2 cents.

I don't agree with implicitly switching to "suggestion mode" based on the _lens parameter.
I think it would be better to be explicit about that a completely different type of search is being performed. It feels weird to me that setting `_lens=chips` gives a completely different set of results (implicit trailing wildcard (prefix search), leading wildcards don't work, different sort order etc.).

This could still be used together with `_lens=chips`.

andersju:
@olovy Fair point. How about something like: switch to autocompletion/suggestion search if `_lens=chips` and `_suggest=true`? `_lens` doesn't _have_ to be `chips` of course, but it'd be very wasteful with anything else. Maybe with an `InvalidQueryException` if `_suggest=true` but `_lens` isn't `chips`?

olovy:
Yes, that or a separate "query parameter" that is mutually exclusive with `q`.

Perhaps 
`_suggest=true&_sortKeyByLang.sv=<query>`
`_suggest=true&_sortKeyByLang.en=<query>`
So that we also can use the .en field which is already indexed?

olovy:
Yes, that or a separate "query parameter" that is mutually exclusive with `q`. e.g. `_suggest_q`

Or perhaps 
`_sortKeyByLang.en=<query>&_suggest={sv,en}&_lens=chips`
So that we also can use the `.en` field which is already indexed?

PR #967, opened 2021-08-11
CRUD: Send error response bodies as json

olovy:
Send error response bodies as json instead of letting servlet container generate a plaintext/html response.

Include stack trace on HTTP 500. Tradeoff between leaking internals and easing trouble-shooting.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/Crud.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/Crud.groovy
@@ -563,7 +563,7 @@ class Crud extends HttpServlet {
         try {
             doPut2(request, response)
         } catch (Exception e) {
-            failedRequests.labels("POST", request.getRequestURI(),
+            failedRequests.labels("PUT", request.getRequestURI(),
                     HttpServletResponse.SC_INTERNAL_SERVER_ERROR.toString()).inc()
             sendError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage(), e)
         } finally {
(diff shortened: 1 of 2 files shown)


rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -943,15 +950,43 @@ class Crud extends HttpServlet {
...
+
+    @Memoized
+    static String httpStatus(int statusCode) {
+        for (Field f : HttpServletResponse.class.declaredFields) {
+            if (f.getName().startsWith("SC_") && f.getType() == int.class && f.getInt(null) == statusCode) {
+                return f.getName().substring("SC_".length()).replace("_", " ")
+            }
+        }
+        return ""
     }
```
olovy, line 990: ¯\_(ツ)_/¯

rest/src/main/groovy/whelk/rest/api/Crud.groovy
```diff
@@ -943,15 +950,44 @@ class Crud extends HttpServlet {
...
-            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, wre.message)
+            sendError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage(), e)
+        }
+
+    }
+
+    static void sendError(HttpServletResponse response, int statusCode, String msg, Exception e = null) {
+        Map json = [
+                "status_code": statusCode,
+                "status": httpStatus(statusCode),
```
andersju, line 966: Custom method can be replaced by a thing we're already pulling in (did some digging):
```suggestion
                "status": org.eclipse.jetty.http.HttpStatus.getMessage(statusCode),
```

andersju (approved):
LGTM (see comment about httpStatus)! Maybe `HttpTools` is a better place for `sendError()`? (Especially if we decide that some of the other APIs should also return errors as JSON?)

PR #971, opened 2021-08-16
Update expressionOf script

olovy:
Updates:
* Identical expressionOf that occur in multiple records are extracted into new "uniformWorkTitle" works. 
* Languages in expressionOf.hasTitle.mainTitle strings are moved to expressionOf.language
* Handle blank languages e.g. `{ "@type": "Language", "label": ["English & Tamil."] }`
* It's ok if instanceOf.expressionOf.language is a subset of instanceOf.language

### DIFF
--- a/whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
+++ b/whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
@@ -499,7 +499,7 @@ Map substitutions() {
             'tyska (lågtyska)'                : 'lågtyska',
             'tyska (medelhögtyska)'           : 'medelhögtyska',
             'tyska (medellågtyska)'           : 'medellågtyska',
-            
+
             // Added in lxl-3547
             'polyglott'                       : 'flera språk',
             'pehlevi'                         : 'pahlavi',
@@ -508,9 +508,9 @@ Map substitutions() {
             'english (anglo-saxon)'           : 'fornengelska',
             'judetyska'                       : 'jiddisch',
             'italiensk'                       : 'italienska',
-            'german (middle high german)'     : 'medelhögtyska'
-            'skotsk-gaeliska' : 'skotsk gäliska'
-            
+            'german (middle high german)'     : 'medelhögtyska',
+            'skotsk-gaeliska'                 : 'skotsk gäliska'
+
             // Also seen
             // Fornnorska
             // Fornryska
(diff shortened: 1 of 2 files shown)


whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
```diff
@@ -355,6 +447,170 @@ private String titleStr(Map thing) {
...
+                l += whichLanguageVersion
+            }
+
+            Map m = ['l': l]
+            languageLinker.linkLanguages(m, 'l')
+            l = m.l
+            if (l.size() == languages.size() && l.every {it['@id'] }) {
+                languageFromTitle.println("$id ${work.hasTitle.mainTitle} --> $title $l")
+                asList(work['hasTitle'])?.first().mainTitle = title
+                work.language = l
```
kwahlin, line 493: Since expressionOf.language is overwritten, are we sure it never contains any language different to those found in expressionOf.hasTitle.mainTitle?

whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
```diff
@@ -288,11 +376,15 @@ class Norm {
         def t = normalize(title.findAll {it.key in TITLE_KEYS})
         
         // Some records have the language in the title e.g. "Tusen och en natt. Svenska" drop the language part
+        // These should normally already have been moved by moveLanguagesFromTitle()
         t.mainTitle?.with { String m ->
             def s = m.split(' ')
             if (s.last() in languageNames) {
                 t.mainTitle = normalize(s.dropRight(1).join(' '))
             }
+            if (s.size() > 3 && s[-1] in languageNames && s[-2] == "&" && s[-3] in languageNames) {
```
kwahlin, line 385: What about languages conjoined with "och", "and" etc.? Is it that moveLanguagesFromTitle() never fails to handle those?

kwahlin (approved):
Good job! Looks correct as far as I can tell. Just a couple of question marks, see comments.

whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
olovy: Good catch!.
I meant to pass languages from expressionOf.hasTitle.mainTitle + expressionOf.language to `languageLinker` but accidentally removed the expressionOf.language part at some point

whelktool/scripts/cleanups/2021/04/lxl-3547-link-expressionOf.groovy
olovy: I think this whole section should be deleted. It's a thing from before languages were moved from titles.
Will just log instad

PR #973, opened 2021-08-19
Feature/lxl 3654 accession number

kwahlin:
- [x] I have run integTest


klngwll (changes requested):
Looks good except that the identifiedBy value is dependent of the context so unfortunately we cannot move it outside its Acquisition entity. A thought could be that it would work for the item-level but in the definition the identification is assigned to the transfer itself and therefore each item could then record several transfers. From MARC docs: "Identification code assigned to materials acquired in a single and separate transfer of custody. ".

kwahlin:
@klngwll Fair enough, I was a bit hesitant about that part.

klngwll (approved):
LGTM!

PR #975, opened 2021-08-25
Optionally exclude hold records when using copywhelk

andersju:
For e.g. working with works it'd be nice to be able to copy all instance records without dragging 50M+ hold (item) records along. This simply adds `--exclude-items` (default false) to `copywhelk` to make that possible. Related to LXL-3613.


olovy (approved):
:ship:

PR #978, opened 2021-08-26
Change property/class names in 351 conversion

kwahlin:
To match the updates in https://github.com/libris/definitions/pull/311


klngwll (approved):
LGTM! I was a bit perplexed at first since it was not in LC:s spec for 1.6 but it is correct according to their updated conversion xslt :)

PR #979, opened 2021-08-30
Feature/fix prepared statement leaks

andersju:
Using fab's `app.whelk.import_work_example_data` with a dataset much larger (~13M docs) than the "usual", small one resulted in Postgres eventually failing with an out of memory error during the `reDenormalize` step, after about 2027500 records processed, on a machine with 48 GB RAM. Suspicious:

```
  CacheMemoryContext: 1040384 total in 7 blocks; 406960 free (20 chunks); 633424 used
...
    2027750 more child contexts containing 20763196416 total in 10138316 blocks; 10235581288 free (2027632 chunks); 10527615128 used
```

`saveIdentifiers()` in `PostgreSQLComponent` wasn't closing its `PreparedStatement` resources, which meant that they would only disappear when the connection was closed, so at the time of error there were a couple of million of those (instead of 5 or so). Fixed that. Also:

- Added a missing `withDbConnection { ... }` to `WhelkCopier`
- Made `reDenormalize()` use `getOuterConnection()` instead of `withDbConnection`, since `loadAll()` uses the former which can't be inside the latter. Could of course make `loadAll()` use `withDbConnection`, but perhaps that's not desirable. In any case, `reDenormalize()` is _only_ used by `WhelkCopier`, so I don't think it matters much.
- Disabled autocommit in `reDenormalize()`, which should speed things up a bit

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -721,20 +721,28 @@ class PostgreSQLComponent {
     void reDenormalize() {
         log.info("Re-denormalizing data.")
         Connection connection = getOuterConnection()
-        connection.setAutoCommit(false)
-        boolean leaveCacheAlone = true
+        try {
+            boolean autoCommit = connection.getAutoCommit();
+            connection.setAutoCommit(false)
+            boolean leaveCacheAlone = true
 
-        long count = 0
-        for (Document doc : loadAll(null, false, null, null)) {
-            refreshDerivativeTables(doc, connection, doc.getDeleted(), leaveCacheAlone)
+            long count = 0
+            for (Document doc : loadAll(null, false, null, null)) {
+                refreshDerivativeTables(doc, connection, doc.getDeleted(), leaveCacheAlone)
 
-            ++count
-            if (count % 500 == 0)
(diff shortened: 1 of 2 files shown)


lrosenstrom (approved):
Nice :ship:

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -720,20 +720,21 @@ class PostgreSQLComponent {
...
-                    log.info("$count records re-denormalized")
-            }
-            clearEmbellishedCache(connection)
+            ++count
+            if (count % 500 == 0)
+                log.info("$count records re-denormalized")
         }
+        clearEmbellishedCache(connection)
+        connection.commit()
+        connection.close()
```
jannistsiroyannis, line 737: Shouldn't this also be in a finally{}-block? In case something after getOuterConnection() throws?

jannistsiroyannis (commented):
I had one comment, other than that, it looked good. Good find! :)

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
andersju: Good point, fixed!

PR #981, opened 2021-09-01
Do not accept hold imports of nonregistrant sigels.

jannistsiroyannis:
The registrant information is (periodically) synced with (from)
bibdb, and will now decide if the incoming record is allowed
or not.

This blocks such imports specifically in the batch import.
Other mechanisms prevent this from happening through the normal APIs.

Existing bad holds still need to be cleaned up.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Document.groovy
+++ b/whelk-core/src/main/groovy/whelk/Document.groovy
@@ -68,6 +68,7 @@ class Document {
     static final List generationDatePath = ["@graph", 0, "generationDate"]
     static final List descriptionCreatorPath = ["@graph", 0, "descriptionCreator", "@id"]
     static final List descriptionLastModifierPath = ["@graph", 0, "descriptionLastModifier", "@id"]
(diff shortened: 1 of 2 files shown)


olovy (approved):
I can't comment on the use of category `http://id.kb.se/bibdb/registrant` but LGTM

whelk-core/src/main/groovy/whelk/Document.groovy
```diff
@@ -519,6 +520,20 @@ class Document {
...
+        Object catObj = get(categoryPath)
+        if (catObj == null)
+            return false
+        List categories
+        if (catObj instanceof List)
+            categories = catObj
+        else
+            categories = [catObj]
+        return categories.any {
+            it["@id"] == "http://id.kb.se/bibdb/registrant"
```
jannistsiroyannis, line 533:
```suggestion
            it["@id"] == "http://id.kb.se/term/bibdb/Registrant"
```

PR #983, opened 2021-09-02
Fix ISO-639-2(T) in language linker

olovy:


klngwll (approved):
These propertynames are really giving me a headache. The langCode is actually the MARC-code if I remember correctly which makes me think this LGTM! 🎈

olovy:
> These propertynames are really giving me a headache. The langCode is actually the MARC-code if I remember correctly which makes me think this LGTM! balloon

I would prefer to call them iso6392b etc but have forgotten the details of that discussion.

PR #984, opened 2021-09-02
Use correct connection methods

kwahlin:


olovy (approved):
LGTM. 

Should also close result set and statment inside a finally-block.

jannistsiroyannis (approved):

PR #987, opened 2021-09-14
Generate a history profile given a records history.

jannistsiroyannis:
The profile will tell you "who" is responsible for each piece
of information in the record. The idea is to use this to decide
if parts of incoming updates should be merged/added/discarded.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -12,6 +12,8 @@ import org.postgresql.PGStatement
 import org.postgresql.util.PGobject
 import org.postgresql.util.PSQLException
 import whelk.Document
+import whelk.history.DocumentVersion
+import whelk.history.History
 import whelk.IdType
 import whelk.JsonLd
 import whelk.Link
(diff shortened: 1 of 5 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2290,6 +2292,29 @@ class PostgreSQLComponent {
...
+            ResultSet rs = null
+            List<DocumentVersion> docList = []
+            try {
+                selectstmt = connection.prepareStatement(GET_ALL_DOCUMENT_VERSIONS)
+                selectstmt.setString(1, id)
+                rs = selectstmt.executeQuery()
+                int v = 0
+                while (rs.next()) {
+                    def doc = assembleDocument(rs)
+                    doc.version = v++
```
olovy, line 2308: Never used as far as I can tell, but it numbering is backwards?
v 0 == current
Or is that intentional?

whelk-core/src/main/groovy/whelk/history/Ownership.java
```diff
@@ -0,0 +1,76 @@
...
+import java.time.ZonedDateTime;
+
+public class Ownership {
+    public String m_lastManualEditor;
+    public Instant m_lastManualEditTime;
+
+    public String m_lastEditor;
+    public Instant m_lastEditTime;
+
+    public Ownership(DocumentVersion version, Ownership previousOwnership) {
```
olovy, line 13: Libris-specific stuff, but I think that is ok for now. Let's just note increased "Libris debt" in whelk-core.
(We could later e.g. move interpretation of the strings xl/Voyager/vcopy/APIX etc. to importers?)

olovy (approved):
LGTM.

A good starting point.
Let's add some tests later.

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
olovy: Ok, same numbering as in `doLoadAllVersions`

jannistsiroyannis: > Ok, same numbering as in `doLoadAllVersions`

Yeah, its used here (where that needs to be the order):
java/whelk/export/marc/ProfileExport.java:174

PR #989, opened 2021-09-22
Consistent version handling in load / loadDocumentByMainId

andersju:
`whelk.storage.load(<id>, 0)` returned the latest version of a document while `whelk.storage.loadDocumentByMainId(<id>, 0)` returned the oldest. Now they both return the oldest.

* Changed the underlying query `GET_ALL_DOCUMENT_VERSIONS` to return things in ascending rather than descending order 
* Made `GET_ALL_DOCUMENT_VERSIONS_BY_MAIN_ID` consistent with a recent change to `GET_ALL_DOCUMENT_VERSIONS` (`ORDER BY modified` => `ORDER BY GREATEST(modified, (data#>>'{@graph,0,generationDate}')::timestamptz)`)
* Adapted two instances where `loadAllVersions` was used (WhelkTool and marc_export). The marc_export integration test passes. I _think_ that's all :)
* Made it possible to use a negative version number. So now: 0 = oldest, 1 = second oldest, ..; -1 = newest, -2 = second newest, etc.

### DIFF
--- a/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
+++ b/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
@@ -173,14 +173,16 @@ public class ProfileExport
         {
             List<Document> versions = m_whelk.getStorage().loadAllVersions(id);
 
-            // The 'versions' list is sorted, with the most recent version first.
+            // The 'versions' list is sorted, with the oldest version first.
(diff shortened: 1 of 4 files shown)


olovy (approved):
LGTM!

jannistsiroyannis (approved):

PR #993, opened 2021-09-27
Get rid of Flask

andersju:
This PR covers a bit more than originally intended (sorry), but the changes are all related and it should be easier to test this way.

The goal is to replace the old [Flask layer](https://github.com/libris/lxlviewer/). The Flask app functions as both

* https://id.kb.se (static pages / user-facing website). This is replaced by https://github.com/libris/id.kb.se, a server-side rendered JS app.
* a middle-man handling requests to the backend XL REST API (proxying them and doing stuff to the requests/responses). This is replaced by proxying API requests directly to XL rather than Flask, achieved through the changes in this PR plus suitable Apache configuration (see below for instructions).

So we're going from:

Browser -> Apache -> Flask -> XL

To:

Browser -> Apache -> XL

Flask did a bunch of magic and we don't want breaking changes in the public API. The goal here is to be able to remove Flask without affecting anything that uses the API, so some changes are in order:

### What site am I
One and the same Flask instance could handle both, say, id-dev.kb.se and libris-dev.kb.se and add different query parameters to the request based on that, before proxying it to XL ("am I id.kb.se or libris.kb.se?").

In the Flask app the site configuration is defined in [viewer/dataaccess.py](https://github.com/libris/lxlviewer/blob/develop/viewer/dataaccess.py). In this PR the configuration is now in `rest/src/main/resources/site_config.json` and read by `Crud` on init. `Crud.getActiveSite()` is used to determine which site is "active", and the result of that is used in a few places in `Crud` to fetch appropriate parameters from the site config. 

"Active site" can be set with `&_site=<foo>` in the query, e.g., `&_site=https://id.kb.se/`. If that's missing, it matches the base URI of the request with the sites in the config. Failing that, it looks for a site alias (like in the Flask app's `BASE_URI_ALIAS` from `instance/config.cfg`, except flipped around!). Failing that, it defaults to whatever is set as `default_site` in `site_config.json` (= https://libris.kb.se/).

(`&site=?` might be unnecessary.)

Certain paths - and id.kb.se - get special treatment by Apache (see configuration below).

### Content negotiation and conversion
Previously, XL's REST API would only give you JSON or JSON-LD, while the Flask app also supports Turtle, N3, TriG and RDF/XML, converting the JSON-LD response from XL before sending it back to the client.

 I added `JsonLDTrigConverter` (nearly identical to `JsonLDTurtleConverter`) and `JsonLD2N3Converter` (nearly identical to `JsonLD2RdfXml`) and adapted the converters to work properly with the REST part (= optionally taking an existing Whelk as an argument). (The converters seem ripe for refactoring but I didn't want to mess with them too much.)
 
Also added `whelk.rest.api.ConverterUtils` as a helper thing, loaded by `Crud` on init and used by `getFormattedResponseBody()` if the determined content type is not JSON or JSON-LD.
 
 To actually use these I added more types in `MimeTypes` and `CrudUtils`, and adapted the latter to take both the `Accept:` header and the file extension of the path into account in `getBestContentType()`. So now one can do:
 
 ```
 ~ curl -sI http://localhost:8180/6ph30v6j0bdfr1t/data.ttl | grep Content-Type
Content-Type: text/turtle;charset=utf-8

~ curl -H 'Accept: text/turtle' -sI http://localhost:8180/6ph30v6j0bdfr1t | grep Content-Type
Content-Type: text/turtle;charset=utf-8

~ curl -sI http://localhost:8180/6ph30v6j0bdfr1t/data.xml | grep Content-Type
Content-Type: application/rdf+xml;charset=utf-8

~ curl -H 'Accept: application/rdf+xml' -sI http://localhost:8180/6ph30v6j0bdfr1t | grep Content-Type
Content-Type: application/rdf+xml;charset=utf-8
 ```
 
etc. (Note that - unlike with Flask - this is currently _not_ supported for `/find`, but it was decided not to care about that. For the moment, anyway. Either way that usage was to my knowledge not documented anywhere.)

Note that Flask would simply add `@context: "/context.jsonld"` to any JSON-LD response from XL:

```python
@negotiator.add(JSONLD_MIMETYPE, 'jsonld', group='RDF')
def render_jsonld(path, data):
    data[CONTEXT] = CONTEXT_PATH
    return _to_json(data)
```

`Crud` now does the same in a few places (search for `CONTEXT_PATH`), but in a way that shouldn't affect ETag calculation.

### Embellish by default

Previously, `/6ph30v6j0bdfr1t` in the backend meant "embellish this thing", while `/6ph30v6j0bdfr1t/data.jsonld` meant "_don't_ embellish this thing".

In the Flask app (and thus the public API), `/6ph30v6j0bdfr1t/data.jsonld` means _do_ embellish this thing (the app would drop `/data.jsonld` from the path before proxying the request).

It was decided to keep the Flask/public behaviour: so, `/6ph30v6j0bdfr1t/data.jsonld` is now embellished by default.

### How to test

Assuming:

* `librisxl/secret.properties` with `baseUri = http://kblocalhost.kb.se:5000/`
* XL running on 8180
* `lxlviewer/viewer/vue-client` running on 8080
* https://github.com/libris/id.kb.se running on 3000
* Ubuntu (otherwise just adapt slightly)

Make sure `/etc/hosts` includes the following:

```
127.0.0.1 id.kblocalhost.kb.se
127.0.0.1 kblocalhost.kb.se
```

Install Apache.

Add `Listen 5000` to `/etc/apache2/ports.conf`.

Edit `/etc/apache2/sites-enabled/000-default.conf` and add:

```apache
<VirtualHost *:5000>
  ServerName kblocalhost.kb.se
  ProxyRequests Off
  ProxyPreserveHost On
  RewriteEngine On

  # NOTE! ProxyPass and ProxyPassMatch preferred over RewriteRule for performance reasons.

  ProxyPassMatch ^/vocab/(data.*) http://localhost:8180/https://id.kb.se/vocab//$1
  ProxyPass /vocab http://localhost:8180/https://id.kb.se/vocab
  ProxyPass /context.jsonld http://localhost:8180/https://id.kb.se/vocab/context

  RewriteCond %{REQUEST_METHOD} ^(POST|PUT|DELETE|OPTIONS)$
  RewriteRule ^/data(.*)$ http://localhost:8180/$1 [P,L]

  # Matches /data, /data.jsonld, ..
  ProxyPassMatch ^/(data\..+)$ http://localhost:8180/$1
  ProxyPassMatch ^/data$ http://localhost:8180/

  # Matches /<something>/data, /<something>/data.jsonld, ..
  ProxyPassMatch ^/(.+/data.+)$ http://localhost:8180/$1

  # Matches XL IDs
  ProxyPassMatch ^/([bcdfghjklmnpqrstvwxz0-9]{15,16}.*) http://localhost:8180/$1

  ProxyPassMatch ^/(http.*)$ http://localhost:8180/$1
  ProxyPassMatch ^/find(.*) http://localhost:8180/find$1

  # Catch-all if nothing above matched
  ProxyPass / http://localhost:8180/

  AddOutputFilterByType DEFLATE text/css text/html text/plain text/xml
  AddOutputFilterByType DEFLATE application/x-javascript text/x-component application/javascript
  AddOutputFilterByType DEFLATE application/json application/ld+json
</VirtualHost>

<VirtualHost *:5000>
  ServerName id.kblocalhost.kb.se
  ProxyRequests Off
  ProxyPreserveHost On
  RewriteEngine On

  # Careful now to avoid recursion
  RewriteCond %{HTTP_ACCEPT} (text/html|application/xhtml)
  RewriteCond %{REQUEST_URI} !\.(json|jsonld)$
  RewriteCond %{REQUEST_URI} !data\..+$
  RewriteRule ^/(.*)$ http://localhost:3000/$1 [P,L]
  ProxyPass /_nuxt http://localhost:3000/_nuxt

  ProxyPassMatch ^/vocab/(data.*) http://localhost:8180/https://id.kb.se/vocab//$1
  ProxyPass /vocab http://localhost:8180/https://id.kb.se/vocab
  ProxyPass /vocab/display/data.jsonld http://localhost:8180/https://id.kb.se/vocab/display
  ProxyPass /context.jsonld http://localhost:8180/https://id.kb.se/vocab/context

  ProxyPassMatch ^/(data.*)$ http://localhost:8180/$1
  ProxyPassMatch ^/find(.*) http://localhost:8180/find$1
  ProxyPassMatch ^/(http.*)$ http://localhost:8180/$1
  ProxyPassMatch ^/(.*) http://localhost:8180/https://id.kb.se/$1
</VirtualHost>
```

(Re)start Apache: `sudo systemctl restart apache2`
(If it complains about a module, just do `sudo a2enmod rewrite` or whatever it wants)

Now http://id.kblocalhost.kb.se:5000/ (or http://localhost:3000/ - same thing) and http://localhost:8080/katalogisering/ _should_ work.

Apache config can probably be improved, and in production it needs to be slightly different (and devops changes are necessary and so on), but this should be enough for a local setup.


jannistsiroyannis (approved):
Vilket jäkla :horse: jobb :)
Ship it!

olovy (approved):
Great work! :star2: 

We need to handle `site_alias` in some other way (not hardcode it in `site_config.json`). But let's merge all of this first.

:ship:

PR #994, opened 2021-09-28
Feature/merge import rules

jannistsiroyannis:
This adds a commandline option to the batch-import which enables selective merging of incoming (bibliograhpic) records with existing ones.
This is done using a (json) rule-file and the computed history of existing records.

### DIFF
--- a/batchimport/integtest/batch14.xml
+++ b/batchimport/integtest/batch14.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<collection xmlns="http://www.loc.gov/MARC21/slim">
+  <record type="Bibliographic">
+    <leader>02625cam a22007458i 4500</leader>
+    <controlfield tag="001">20203738</controlfield>
+    <controlfield tag="003">SE-LIBR</controlfield>
+    <controlfield tag="005">20170713141440.0</controlfield>
+    <controlfield tag="008">170327s2017    sw ||||j     |00| 0|swe d</controlfield>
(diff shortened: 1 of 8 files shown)


olovy (commented):
So if I understand this correctly the following configuration would allow any properties to be added to existing docs?

```
{
    "rules": [
	{
            "operation": "add_if_none",
            "path": ["@graph", 1]
        }
    ]
}
```

olovy (approved):
:ship:

jannistsiroyannis:
> So if I understand this correctly the following configuration would allow any properties to be added to existing docs?
> 
> ```
> {
>     "rules": [
> 	{
>             "operation": "add_if_none",
>             "path": ["@graph", 1]
>         }
>     ]
> }
> ```

Under @graph,1 yes, but not under @graph,0 (unless I've made a mistake).
But it's less .. invasive than it sounds, because you can never do deeper than the first list (except for the @graph one).

olovy:
> > So if I understand this correctly the following configuration would allow any properties to be added to existing docs?
> > ```
> > {
> >     "rules": [
> > 	{
> >             "operation": "add_if_none",
> >             "path": ["@graph", 1]
> >         }
> >     ]
> > }
> > ```
> 
> Under @graph,1 yes, but not under @graph,0 (unless I've made a mistake). But it's less .. invasive than it sounds, because you can never do deeper than the first list (except for the @graph one).

I was thinking that it is a good thing. So that it is possible to let all properties be added without having to list them explicitly.

olovy:
:ship:

PR #995, opened 2021-09-28
Fix: Don't check for multiple holds for sigel when deleting hold

olovy:
It should be possible to delete existing bad holdings.

e.g. https://libris-qa.kb.se/katalogisering/8sl27z5l5s089r0


kwahlin:
:ship:

andersju (approved):
LGTM!

PR #999, opened 2021-09-30
Make embellished responses cachable

olovy:
Make ETag/If-None-Match (304 NOT MODIFIED) work with embellish.

(ETag is an opaque string so we can put whatever we want inside)
When `embellish==false` ETag is checksum of document
When `embellish==true` include checksum of both document and embellished document in ETag.

Both ETag forms can be used in PUT.

### DIFF
--- a/rest/build.gradle
+++ b/rest/build.gradle
@@ -32,11 +32,6 @@ sourceSets {
 }
 
 test {
-    systemProperty 'xl.secret.properties', '../secret.properties'
-}
-
-test {
-    systemProperty 'xl.secret.properties', '../secret.properties'
     jacoco {
         destinationFile = file("$buildDir/jacoco/jacocoTest.exec")
         classDumpDir = file("$buildDir/jacoco/classpathdumps")
--- a/rest/src/main/groovy/whelk/rest/api/Crud.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/Crud.groovy
@@ -1,6 +1,5 @@
 package whelk.rest.api
 
-
 import groovy.transform.PackageScope
 import groovy.util.logging.Log4j2 as Log
 import io.prometheus.client.Counter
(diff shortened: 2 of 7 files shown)


andersju:
Some failing tests:
```
    CrudSpec. PUT to /<id> - Global registrant can correct missing sigel in holding
    CrudSpec. PUT to /<id> - Registrant can correct missing sigel in holding
    CrudSpec. PUT to /<id> should update content of type Work
    CrudSpec. PUT to /<id> should update content of type holding
    CrudSpec. PUT to /<id> should update document if it exists
    CrudSpec. PUT to /<id> should update holding if user has global registrant permission active
    CrudSpec. PUT to /<id> should update holding if user has kat permission for code
```

andersju:
Now a few different ones:
```
    CrudSpec. GET document with If-None-Match should return 200 Ok or 304 Not Modified[0]
    CrudSpec. GET document with If-None-Match should return 200 Ok or 304 Not Modified[1]
    CrudSpec. GET document with If-None-Match should return 200 Ok or 304 Not Modified[2]
    CrudSpec. GET document with If-None-Match should return 200 Ok or 304 Not Modified[5]
```

olovy:
:thinking:

olovy:
Ok, the tests passed in IntelliJ test runner but failed in `./gradlew test`...

:scream_cat: 

This was because the gradle test task set `xl.secret.properties`
```
test {
    systemProperty 'xl.secret.properties', '../secret.properties'
}
```
which made the test document get a different base URI and a different checksum...
(the tests maybe shouldn't use hardcoded checksums but anyway)

```
{
  "@graph": [
    {
      "@id": "https://libris.kb.se/1234",
      "mainEntity": {
        "@id": "https://libris.kb.se/1234#it"
      }
    },
    {
      "@id": "https://libris.kb.se/1234#it",
      "@type": "Instance",
      "prop1": {
        "@id": "https://foo"
      }
    }
  ]
}
```

```
{
  "@graph": [
    {
      "@id": "http://kblocalhost.kb.se:5000/1234",
      "mainEntity": {
        "@id": "http://kblocalhost.kb.se:5000/1234#it"
      }
    },
    {
      "@id": "http://kblocalhost.kb.se:5000/1234#it",
      "@type": "Instance",
      "prop1": {
        "@id": "https://foo"
      }
    }
  ]
}
```

I've removed secret.properties from the gradle task and fixed RemoteSearchApiSpec which depended on it.

When troubleshooting I realized I also got bitten by `GString` once again. The test data contained GStrings (interpolated groovy string) which gave different checksums than String. Didn't really affect the result in the end but I've updated the Checksum code to handle them.

andersju (approved):
:ship:

PR #1002, opened 2021-10-06
REST: Render things as HTML

andersju:
Makes it possible to get (minimal) HTML versions of resources, e.g.:

```bash
~ curl -I -H 'Accept: text/html' http://localhost:8180/bvnnfbjn3kq7j31
HTTP/1.1 200 OK
...
Content-Type: text/html;charset=utf-8

~ curl -I http://localhost:8180/bvnnfbjn3kq7j31/data.html
HTTP/1.1 200 OK
...
Content-Type: text/html;charset=utf-8
```
The actual template used is `whelk-core/src/main/resources/html_resource.tpl`.

Up until now the Flask app has provided this, but now XL itself needs to do it.

Consider this a quick PoC, just wanted to get some input. (Like, using constants from the `ElasticSearch` class is a little hmmmm) 

Note that with this change, fetching a document like http://kblocalhost.kb.se:5000/7qjgtwzk4p74qr6/data.jsonld through the Flask app _in a browser_ won't work, because the Flask app drops the `/data.jsonld` part before proxying and doesn't change the original `Accept` header from the browser (though `data.json` will work, because in that case it _is_ changed), instead just asking XL for `7qjgtwzk4p74qr6` with `Accept` being `text/html, ...`, which _used_ to work only because XL sent JSON-LD by default if no filename extension was specified and the `Accept` value was something not known by XL :shrug:

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/JsonLD2HtmlConverter.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/JsonLD2HtmlConverter.groovy
@@ -20,13 +20,16 @@ class JsonLD2HtmlConverter implements FormatConverter {
         Map thingChipMap = whelk.jsonld.toChipAsMapByLang(source['@graph'][1],
                 ElasticSearch.LANGUAGES_TO_INDEX, ElasticSearch.REMOVABLE_BASE_URIS)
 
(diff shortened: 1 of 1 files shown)


olovy:
What Anders said

PR #1020, opened 2021-10-20
Add initial digital reproduction tooling

niklasl:
This  contains:

- Updated marcframe mappings and mechanisms to produce the same MARC from simplified descriptions (utilizing embellished data as source for the redundancies in MARC).
- A process for creating digital reproductions (just a whelktool script; might be obsolete if we use the REST API).
- Batch job for "digibest" using the a sample backlog.


niklasl (commented):
Note that this is a manual mapping which mirrors (one of) the mappings defined in https://github.com/libris/definitions/pull/341

Later on, if those mappings have landed and have a more powerful mechanism of getting at term-matching relationships in core, we could switch to using that here, and remove these hard-coded patterns and term references.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -2136,109 +2245,329 @@
...
                   }
                 ]
               }
             }
+          }
+        ]
+      },
+      {
+        "type": "InjectWhenMatchingOnRevert",
+        "_TODO": "This should be closeMatch/broader-driven (and computed on search)",
```
niklasl, line 2412: Note that this is a manual mapping which mirrors (one of) the mappings defined in libris/definitions#341

Later on, if those mappings have landed and have a more powerful mechanism of getting at term-matching relationships in core, we could switch to using that here, and remove these hard-coded patterns and term references.

whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -2136,60 +2245,127 @@
...
+        "produceMissing": {
+          "produceProperty": "marc:publicNote",
+          "showProperties": [
+            {"property": ["usageAndAccessPolicy", "label"]},
+            {
+              "property": ["publisher", "name"],
+              "useValueFormat": {
+                "contentFirst": " via ",
+                "contentAfter": " och ",
+                "contentLast": ""
```
olovy, line 2259: :sunglasses:

olovy (approved):
LGTM!

PR #1031, opened 2021-11-03
Change bib 653[i2=5] from Geographic to Place

niklasl:
This is mainly to be able to use general Place links as geographic subjects when reverting to MARC21.

While incoming MARC21 in practise might intend for these to be Geographic subjects (unlinked), we do not foresee any practical issues with this slight generalization. The semantic difference is just that the implicit Geographic type should/might indicate that the thing is defined by a controlled concept scheme while Place does not imply such formalization. And in practise, this distinction does not hold for incoming data with unlinked geographic names (i.e. they are "just names referring to places").

This depends on https://github.com/libris/definitions/pull/335 to work upon revert.

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -16582,7 +16582,7 @@
         },
         {
           "when": "i2=5",
-          "resourceType": "Geographic"
+          "resourceType": "Place"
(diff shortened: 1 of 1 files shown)


olovy (approved):
LGTM

klngwll (approved):
LGTM!

PR #1034, opened 2021-11-10
delete-holdings-for-T

poacherone:
Delete holdings for sigel T, based on list with controlnumbers and LibrisIII-numbers.


jannistsiroyannis (approved):
:ship:

PR #1042, opened 2021-11-22
Feature/update sparql prefixes

kwahlin:
Mainly to set `:` as the prefix for `https://id.kb.se/vocab/` but adds a couple of more prefix too. Some changes on the installation was needed too.

_Important_ that the blank prefix comes last in `namespace_prefixes.tsv`, otherwise it won't work (seems to get overwritten by any other prefix).


olovy (approved):
LGTM!

PR #1044, opened 2021-11-22
Search: make matches- prefix match narrower terms

olovy:
Before it only worked on date fields. Now it will match all `narrower`
(really `@reverse/broader`) things on non-date fields.

i.e. `?matches-instanceOf.genreForm.@id=https://id.kb.se/term/saogf/Fantasy`
will find both "Fantasy" and "Urban fantasy"

`Ranges` and `RangeParameterPrefix` should probably be renamed at some
point since they don't deal only with ranges anymore.

`matches-` still doesn't work together with `or-`, we might want to fix
this at some point.


whelk-core/src/main/groovy/whelk/search/Ranges.groovy
```diff
@@ -132,4 +142,18 @@ class Ranges {
...
+    private class OrNarrower implements Query {
+        String value
+
+        OrNarrower(String value) {
+            this.value = value
+        }
+        
+        @Override
+        Map toQuery() {
+            def values = [value] + whelk.relations.followReverseBroader(value).collect() // default max size in ES 65,536...
```
olovy, line 155: Note to self: Limit it instead

niklasl (approved):
LGTM! See my added form for the query results (and reasoning of).

niklasl:
Added a form for the search result path in 10df18429608b76a5cbe0510a26aa4d5ea0feef3. This uses the new property defined in https://github.com/libris/definitions/pull/359.

olovy:
> Added a form for the search result path in [10df184](https://github.com/libris/librisxl/commit/10df18429608b76a5cbe0510a26aa4d5ea0feef3). This uses the new property defined in [libris/definitions#359](https://github.com/libris/definitions/pull/359).

I buy the reasoning behind the property. LGTM!

PR #1045, opened 2021-11-23
Feature/lxl 2483 extended place descriptions 2

kwahlin:
Adds closeMatch relations to DDC, LCSH and FAST.
Adds exactMatch relations to Freebase, Geonames, TORA and YSO.
Something else that should be included?


olovy (approved):
:ship:

klngwll (approved):
LGTM!
One addition on top of my mind would be closeMatch(?) in getty: https://www.wikidata.org/wiki/Property:P1667

olovy:
> LGTM! One addition on top of my mind would be closeMatch(?) in getty: https://www.wikidata.org/wiki/Property:P1667

exactMatch?

kwahlin:
Sure, I'll add Getty IDs too. Think closeMatch should be the preferred relation as the resources there are subjects/concepts.

PR #1046, opened 2021-11-24
Fix confusing error message in 401 from AuthenticationFilter

olovy:
"Missing access token" sounds like the token is missing in request
when the problem actually is that it cannot be verified (found) in
xl_auth.


andersju (approved):
:ship:

niklasl (approved):
LGTM!

PR #1047, opened 2021-11-24
DigitalReproductionAPI: set generationProcess on extracted works

olovy:


rest/src/main/groovy/se/kb/libris/digi/DigitalReproductionAPI.groovy
```diff
@@ -308,7 +308,11 @@ class XL {
                 }
             }
 
-            def workId = create([:], work)
+            def record = [
+                    'generationProcess': ['@id': 'https://libris.kb.se/api/_reproduction']
```
olovy, line 312: ¯\_(ツ)_/¯

niklasl: Can't think of anything better! (It's at least as good as our current "made up" wheltool script URI:s. And anything "better" would be formalizing an internal API, which might lead down a path we're not sure about anyway.)

rest/src/main/groovy/se/kb/libris/digi/DigitalReproductionAPI.groovy
```diff
@@ -302,13 +304,18 @@ class XL {
...
+                def title = asList(instance.hasTitle).find { it['@type'] == 'Title' }
+                if (title) {
+                    work.hasTitle = [ title + [source: [['@id': instanceId]]] ] 
                 }
             }
 
-            def workId = create([:], work)
+            def record = [
+                    'generationProcess': ['@id': DigitalReproductionAPI.API_LOCATION],
+                    'derivedFrom'      : [['@id': instanceId]]
```
niklasl, line 315: We haven't defined this as repeatable... but perhaps we ought to?

PR #1049, opened 2021-11-25
Feature/improve produce if missing

niklasl:
This is done to improve the use of `showProperties` for the `marc:publicNote` added on `associatedMedia` upon revert, specifically handling various combinations and missing values thereof.

This required an adjustment of the generated`seriesStatement` configuration (now requires that to explicitly allow for presence of partial values).

This also dropped the need for the singular (albieit configured) `locale` in `marcframe.json`, instead relying on `locales` in `JsonLd` (alas not yet configurable).


olovy (approved):
:ship:

PR #1053, opened 2021-11-29
DigitalReproductionAPI: Parameter for disabling work extraction

olovy:


kwahlin (approved):
LGTM

rest/src/main/groovy/se/kb/libris/digi/DigitalReproductionAPI.groovy
```diff
@@ -92,7 +92,8 @@ class DigitalReproductionAPI extends HttpServlet {
         def service = new ReproductionService(xl : new XL(headers : forwardHeaders, apiLocation: getXlAPI(request)))
 
         try {
-            String id = service.createDigitalReproduction(parse(request))
+            boolean extractWork = !Boolean.parseBoolean(request.getHeader("x-dont-extract-work"))
```
niklasl, line 95: Is this preferable over a request parameter? (It is possible to combine those and POST payloads).

olovy: I did it as request parameter first but the I started writing it as a header in my test request so I changed it...
I have no preference but I guess a request parameter is more idiomatic.

niklasl: I can certainly see that it may be (since it is less "obtrusive"), but it depends on how this is to be handled (having it as a URI parameter is more akin to having two "endpoints"). The reason for why I thought of a request parameter was mostly because I thought of the earlier idea of a general "extract entity"-API; but since that design would do a POST of the path+entity to be extracted, we probably can't marry these in this design anyway...

niklasl: I see! Ignore my subsequent pondering then, I suppose our intuitions converging is good enough. :slightly_smiling_face:

niklasl (approved):
LGTM!

PR #1055, opened 2021-11-30
Add data injection to items copied on revert

niklasl:
Use to add the text "[originalversion]" as `qualifier` to identifiers and `appliesTo` to extent and publication.


klngwll (approved):
LGTM!

PR #1057, opened 2021-12-06
Add script for removing broken SCB links

kwahlin:
This script removes any object that points to a non-resolving URL starting with http://www.scb.se/

### DIFF
--- a/whelktool/scripts/cleanups/2021/11/lxl-2537-delete-broken-scb-links.groovy
+++ b/whelktool/scripts/cleanups/2021/11/lxl-2537-delete-broken-scb-links.groovy
@@ -1,0 +1,55 @@
+import whelk.util.DocumentUtil
+
+PrintWriter removed = getReportWriter('removed.tsv')
+
+selectByCollection('bib') { data ->
+    Map instance = data.graph[1]
+    String id = data.doc.shortId
+
+    List badPaths = []
+
+    // Remove broken links
+    boolean modified = DocumentUtil.traverse(instance) { value, path ->
+        if (value in String && value.startsWith('http://www.scb.se/')) {
+            int respCode = getResponseCode(value)
+            if (respCode in [400, 404]) {
+                badPaths << path
+                removed.println("${id}\t${respCode}\t${path}\t${value}")
+                return new DocumentUtil.Remove()
+            }
+        }
+    }
+
+    // Remove the objects containing broken links altogether
+    badPaths.reverseEach { bp ->
+        if (instance[bp[0]] in List) {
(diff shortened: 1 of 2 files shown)


whelktool/scripts/cleanups/2021/11/lxl-2537-delete-broken-scb-links.groovy
```diff
@@ -0,0 +1,55 @@
...
+                incrementStats("removed", it)
+            }
+            if (instance[bp[0]].isEmpty())
+                instance.remove(bp[0])
+        }
+        else if (instance[bp[0]] in Map) {
+            Map removedObj = instance.remove(bp[0])
+            removedObj.each {
+                incrementStats("removed", it)
+            }
```
niklasl, line 37: Are we sure there are no links more deeply nested in the data? Make an else-clause throwing an error to ensure we don't miss any paths already removed in the initial "traverse" section?

kwahlin: Good point, however it can be verified from the report "removed.tsv" (attached in Jira) that we won't miss any paths this way.

niklasl: In practise I can imagine it's safe now, but the pattern is brittle in that if someone adds a link in a more nested object (say in an `hasPart` or `inSeries`), that link will be removed without the containing object being removed? If so, I'd say add an Exception just to make the pattern robust (if for nothing else then for the case where someone copies this pattern for another cleanup later on).

lrosenstrom: Containing objects are removed automatically if they are empty after removal (https://github.com/libris/librisxl/blob/develop/whelktool/README.md#L370). Is that the case here?

niklasl: Not entirely, this is in the cleanup step which removes the container outside if it is empty.

kwahlin: @niklasl it's rather the opposite since we remove the first or second step in the path from the instance, e.g. if the path to a removed url was `instance.associatedMedia.0.hasPart.0.uri.0`, then we would remove `instance.associatedMedia.0`. So it's still brittle since we could then wrongly remove e.g. `instance.associatedMedia.0.hasPart.1`. I'll make it more robust.

olovy (approved):

PR #1058, opened 2021-12-08
Fix 'null' in string from produceMissing

olovy:
Test case previously produced "null via Kungliga biblioteket"

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
@@ -54,7 +54,7 @@ abstract class MarcFramePostProcStepBase implements MarcFramePostProcStep {
             if (key instanceof Integer) {
                 return source[key]
             }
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/MarcFramePostProcStep.groovy
```diff
@@ -54,7 +54,7 @@ abstract class MarcFramePostProcStepBase implements MarcFramePostProcStep {
             if (key instanceof Integer) {
                 return source[key]
             }
-            return source.collect { getValue(it, key) }
+            return source.collect { getValue(it, key) }.grep()
```
niklasl, line 57:
```suggestion
            return source.findResults { getValue(it, key) }
```

niklasl: I did like that. 😉 🤷

niklasl (approved):
LGTM! (Could use suggested variant.)

PR #1061, opened 2021-12-14
Make it possible to write merge rules targeting list members.

jannistsiroyannis:
When an incoming record is to be merged with an existing record,
and the rules in play say to target a member of a list, say for
example you want to replace:
["@graph", 1, "instanceOf", "contribution", 0, "agent"].

This doesn't work, because there is no way to know that item 0
in the incoming "contribution" list is really the same entity
as item 0 in the existing list. After all there could be any
number of items in either list, and in any order.

What this does, is to allow the rules to make certain assumptions
in cases where there is only a single item in both of the lists of a
certain type.

For example, the above (not working) path could now instead be
written as:
["@graph", 1, "instanceOf", "contribution", "@type=PrimaryContribution", "agent"].

which will now work _IF AND ONLY IF_ both records have only a
single PrimaryContribution. If this is not the case, there's
nothing we can do, and the rule has no effect for that record.

### DIFF
--- a/batchimport/src/main/java/whelk/importer/Merge.java
+++ b/batchimport/src/main/java/whelk/importer/Merge.java
@@ -18,7 +18,7 @@ public class Merge {
         "rules": [
             {
                 "operation": "replace",
-                "path": ["@graph", 1, "hasTitle"],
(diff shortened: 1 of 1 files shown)


olovy (approved):
:ship: !

PR #1063, opened 2021-12-16
Add script for bumping year based ShelfMarkSequence

olovy:
Should we check `description` as well?


kwahlin (approved):
LGTM! In what way would the descriptions change? You mean if `description` is something like "Suecana broschyrer 2021"?

olovy:
> LGTM! In what way would the descriptions change? You mean if `description` is something like "Suecana broschyrer 2021"?

Exactly. But let's just say that's not how it works.

PR #1064, opened 2021-12-17
Feature/264conversion

klngwll:
Specs for 264. Elib example will always fail in current postprocess. PR for discussion and further investigation.


klngwll:
Superseded by https://github.com/libris/librisxl/pull/1729

PR #1065, opened 2021-12-17
Feature/faster exports

olovy:
The main improvements are
- Only embellish `Item` one level cards instead of [cards, chips] ~50% faster for holdings
- Don't put checksums in `887#a` ~10% faster (but less after fix above)
- Build MARC records in parallel in marc_export API ~linear speedup with number of CPUs

See individual commits.

For webbsök sync most of time is spent on holdings
`marc_export/metrics`
```
marc_export_single_doc_latency_seconds_sum{collection="auth",} 34544.14953423608
marc_export_single_doc_latency_seconds_sum{collection="bib",} 9298.385669556103
marc_export_single_doc_latency_seconds_sum{collection="hold",} 326076.8567143171
```


whelk-core/src/main/groovy/whelk/converter/marc/JsonLD2MarcXMLConverter.groovy
```diff
@@ -51,34 +51,25 @@ class JsonLD2MarcXMLConverter implements FormatConverter {
...
             def df = record.createDatafield("887")
-            df.addSubfield("a".charAt(0), mapper.writeValueAsString(["@id":identifier,"modified":modified,"checksum":checksum]))
+            df.addSubfield("a".charAt(0), mapper.writeValueAsString(["@id":identifier,"modified":modified]))
             df.addSubfield("2".charAt(0), "librisxl")
             record.addField(df)
         }
         return record
     }
 
-    DocumentFragment convertToFragment(final Document doc) {
```
olovy, line 73: unused

olovy (commented):
Revert "Compile DocumentUtil.DFS statically" since it makes the code less clear and doesn't change performance much,

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -422,7 +422,7 @@ class PostgreSQLComponent {
             config.setJdbcUrl(sqlUrl.replaceAll(":\\/\\/\\w+:*.*@", ":\\/\\/"))
             config.setDriverClassName(driverClass)
             config.setConnectionTimeout(0)
-            config.setMetricsTrackerFactory(new PrometheusMetricsTrackerFactory())
+            //config.setMetricsTrackerFactory(new PrometheusMetricsTrackerFactory())
```
olovy, line 425: Investigate this. Profiling says this metrics builder blocks a lot.

marc_export/src/main/java/whelk/export/marc/ProfileExport.java
```diff
@@ -60,77 +74,93 @@
...
-                    String id = resultSet.getString("id");
-                    String collection = resultSet.getString("collection");
-                    String mainEntityType = resultSet.getString("mainEntityType");
-                    Timestamp createdTime = resultSet.getTimestamp("created");
-                    Boolean deleted = resultSet.getBoolean("deleted");
-
-                    boolean created = zonedFrom.toInstant().isBefore(createdTime.toInstant()) &&
-                            zonedUntil.toInstant().isAfter(createdTime.toInstant());
-
-                    int affected = exportAffectedDocuments(id, collection, created, deleted, fromTimeStamp,
```
olovy, line 112: TODO: metrics counters need to be fixed...

jannistsiroyannis (commented):
The embellishment changes look good, and if they make things faster, great! :ship: 

I'm less sure of the changes to the "ProfileExport" (see comment below, for one). I'm confident it could have been done in a simpler way. I don't agree with the idea that "more threads would just slow us down", it seems to be based on the misconception that threads will be given timeslices to waste while they're blocked. But that's not how it works. The kernel will never wake up a thread while it is waiting on a read() or write(). It will wake them up only when they can actually run.

olovy:
No. I forgot to add that this is based on the fact that the worker threads are CPU bound (MARC conversion). We can try to tweak it to some factor of number of CPUs since there is some IO going on.

marc_export/src/main/java/whelk/export/marc/ProfileExport.java
```diff
@@ -430,4 +460,206 @@ private PreparedStatement getAllChangedByStatement(String id, Timestamp from, Ti
...
+
+    /**
+     * Prepare and convert documents to MARC in parallel.
+     *
+     * Work is performed by a shared pool handling all requests. (Using more threads will just slow us down.) 
+     * Serializing and writing the converted records to the output stream is done by a separate single thread per request. 
+     * 
+     * Getting ids to be exported from the database will always be faster than exporting them.
+     * So we need to block the calling thread when too much work is queued up.
+     * Alas, there is no clean way of having a ThreadPoolExecutor with a bounded queue where the caller blocks
```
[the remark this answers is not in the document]
olovy: I don't agree it's simpler. And since this thread is waiting >99% of the time it would mostly be scheduled to run without anything to do.

ThreadPool isn't really a thread pool as it doesn't reuse threads but starts a new one for each task. Maybe not a big problem here since tasks are relatively long running. 

I'd rather replace it instead.

jannistsiroyannis:
> > it seems to be based on the misconception that threads will be given timeslices to waste while they're blocked. But that's not how it works. The kernel will never wake up a thread while it is waiting on a read() or write(). It will wake them up only when they can actually run.
> 
> No. I forgot to add that this is based on the fact that the worker threads are CPU bound (MARC conversion). We can try to tweak it to some factor of number of CPUs since there is some IO going on.

Yes, and being CPU bound is what we want! It means all cores are busy doing useful work!

I do realize that one large request could affect the response time of other smaller ones, if all cores have work. But if many threads want to run, time will be divided among them, no thread will be "starved" unless you add thread priorities into the mix (which we won't). The overhead of creating (rather then reusing) a thread is real of course, but if you batch records, so that each thread gets more than one, that overhead shrinks to ~nothing.

In the end, this is a question of throughput not latency, and if all cores are doing usefull work, throughput will be maximized.

I guess we'll agree to disagree? :) In the end, you did the work and I did not, so you get to decide of course!

olovy:
I did a quick comparison
```
psql -h localhost -U postgres whelk_dev -c 'truncate lddb__embellished' &&  time curl -v -XPOST "http://localhost:8580/marc_export/?from=2020-01-01T00:00:00Z&until=2021-01-01T00:00:00Z" --data-binary @./profil.properties -o /dev/null
```

After one warm-up round

Current implementation
```
real	5m21,009s
```

Naive use of `ThreadPool`, without any batching 
```
real	6m14,979s
```
But I think we don't want to batch ids since auth record tasks can be so much larger. 
Also, passing errors from workers (assuming that's what we want to do) will be more complicated.

I think I want to stick to the current implementation. I will clean it up though :)

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
olovy: Yup. 

Fixed here https://github.com/brettwooldridge/HikariCP/pull/1265

PR #1068, opened 2022-01-04
API documentation: minor fix to example

Abbe98:


olovy:
Thanks!

Hmm, the example still doesn't work for me if I don't url-encode the `|`character (as `%7C`).

Perhaps this is clearer:?
```
curl -XGET -v -s -G -H 'Accept: application/ld+json' 'https://libris-qa.kb.se/find' --data-urlencode 'q=tove (jansson|lindgren)' --data '_limit=2'
```

PR #1069, opened 2022-01-07
Fix clean up after deleted documents

olovy:
Fix the following two issues related to document deletion when the deleted document has "`@reverse` relations" to other documents.

If e.g. "<sao/Digitala möten>" which has `broader` "<sao/Möten>" is deleted
* "<sao/Möten>" is not reindexed
* A (broken) link to "<sao/Digitala möten>" is still included under `@reverse/broader` in embellished "<sao/Möten>"


olovy:
FYI

PR #1071, opened 2022-01-13
Workaround for marc_export.jar output garbled by log4j warning

olovy:
We should fix this everywhere but let's start with marc_export.jar.
Maybe it's time to get rid of log4j completely...


olovy:
I guess we could just set `Multi-Release: true` in all fat-jars to handle any dependency that depends on this.

niklasl (commented):
:shipit: !

PR #1070, opened 2022-01-12
Add script for linking lcgft/lcsh terms in saogf terms

lrosenstrom:
Labels mapped by quering https://id.loc.gov/authorities/subjectheadings/label/<label> in a bash script.

- 112 voyager form nodes (see https://jira.kb.se/browse/LXL-3739) will be linked
- 18 on the "new" form

TBD
*Change closeMatch -> exactMatch?
*There is a mix of https://id.loc.gov/authorities/genreForms, https://id.loc.gov/authorities/subjects, https://id.loc.gov/authorities/childrensSubjects etc in the label to URI mapping. Do we care?

Corrected manually (spelling mistakes etc):

 Darkwawe (music) -> Darkwave (Music) => https://id.loc.gov/authorities/genreForms/gf2014026759.html (pm14bfp71ldshtn)
 Taraab -> Taraab (Music) => https://id.loc.gov/authorities/genreForms/gf2014027123.html (wt7bjpmf1s02t5n)
 Occational verse -> Occasional verse => https://id.loc.gov/authorities/genreForms/gf2014026460.html (khw087h30db2179)
 Alternative histories (fiction) -> Alternative histories (Fiction) => https://id.loc.gov/authorities/genreForms/gf2014026220.html (0xbdhl4j3g28rlz)
 Straight-edge -> Straight-edge (Music) => https://id.loc.gov/authorities/subjects/sh2003002058.html (42gkrt7n4r3cqv8)
 Indians Music -> Indians--Music => https://id.loc.gov/authorities/subjects/sh85065058.html (86lpwfvs37m8dj2)
 Minutes (records) -> Minutes (Records) => https://id.loc.gov/authorities/genreForms/gf2014026128.html (0xbfnmfj240rck1)

PR #1074, opened 2022-01-20
Add script for linking lcsh terms in sao

lrosenstrom:
See https://jira.kb.se/browse/LXL-3795 for more info. 
Similar to https://github.com/libris/librisxl/pull/1070, but for sao. This links >17000 lcsh terms. About 850 left for manual handling.


olovy (approved):
LGTM!

We should evaluate `closeMatch` vs `exactMatch` for id.loc.gov terms.

PR #1076, opened 2022-01-25
Feature/lxl 3590 see also

olovy:


klngwll (approved):
LGTM!

whelktool/scripts/cleanups/2022/01/lxl-3590-see-also.groovy
```diff
@@ -0,0 +1,58 @@
...
+                }
+            }
+        }
+    }
+}
+
+String find(Map seeAlso) {
+    def query = [
+            'q' : [seeAlso.values().join(" ")]
+    ]
```
kwahlin, line 47: Clever! I'm stealing this solution for https://jira.kb.se/browse/LXL-3589

whelktool/scripts/cleanups/2022/01/lxl-3590-see-also.groovy
```diff
@@ -0,0 +1,58 @@
...
+
+    AtomicReference<String> id = new AtomicReference<>() 
+    selectByIds(queryIds(query).collect()) { candidate -> 
+        def (record, candidateThing) = candidate.graph
+        
+        if (seeAlso.every { key, value -> candidateThing[key] == value}) {
+            id.set(candidateThing.'@id')
+        }
+    }
+    return id.get()
```
kwahlin, line 57: Can't there be more than one matching agent? Also, maybe more could be matched if variants were compared too?

kwahlin (commented):
Nice solution. See comments for possible improvement.

whelktool/scripts/cleanups/2022/01/lxl-3590-see-also.groovy
olovy: > Can't there be more than one matching agent? 

I think there _shouldn't_ be but there of course will be in the actual data . Good catch. 

> Also, maybe more could be matched if variants were compared too?

Will try compare them with some extra guards.

PR #1080, opened 2022-02-02
Feature/merge ignore some links

jannistsiroyannis:
Create exceptions to the rule that "objects containing links may never be replaced in the merge process".

With this change, a few select sorts of objects may be replaced even if they contain links.

### DIFF
--- a/batchimport/src/main/java/whelk/importer/Merge.java
+++ b/batchimport/src/main/java/whelk/importer/Merge.java
@@ -147,9 +147,22 @@ public class Merge {
                 }
             }
 
+            // Determine if the subtree that is to potentially be replaced contains links.
+            // If it does, it should generally speaking not be replaced. But there is a list
+            // of exceptions to this rule:
+            boolean containsSanctifiedLinks = subtreeContainsLinks(base);
+            // instanceOf,language may be overwritten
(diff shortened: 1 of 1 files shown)


klngwll (commented):
I figure these are considered special because they are always linked from codes in 008. However there is a fair amount of postprocessing done and not all end up as publication depending on the kinds of dates that are connected to these codes (production, distribution, marc:primaryProvisionActivity is examples on top of my mind). Whether or not this is of relevance to any sources that need this merge-functionality I cannot say :)

jannistsiroyannis:
This is more a question of wanting this properties overwritten in the export even if they were (automatically) linked to something before!
I'll merge it as is if there's no technical argument against I think!

niklasl:
Just to verify assumptions: Since a linked work is has no language "visible" here (in the canonical, normalized source record), that language will not be replaced. I would expect that to be correct if the _linked_ work is considered correct and verified here (i..e "this is the print of this Swedish translation").

_But_ if the merged data is considered of higher quality and being more trustworthy than "the action of the linking of the work" (in theory, I do not know if this assumption ever holds), that would mean that this merging could indicate a challenge ("this is _not_ a print of the Swedish translation, it is a print of the Finnish translation").

Again, I would assume this is not a practical case during merging, but I want to check this assumption. (Ping @poacherone .)

PR #1089, opened 2022-02-23
Digital Reproduction API: Make it possible to pass data to holdings

olovy:


rest/src/main/groovy/se/kb/libris/digi/DigitalReproductionAPI.groovy
```diff
@@ -261,19 +272,24 @@ class ReproductionService {
...
                         '@type' : 'Item',
                         'itemOf' : ['@id' : thingId],
                         'heldBy': ['@id' : heldById],
-                        'hasComponent' : [
-                                [
-                                        '@type' : 'Item',
-                                        'heldBy': ['@id' : heldById],
-                                ]
-                        ]
+                        'hasComponent' : components
```
niklasl, line 292: Only add hasComponent if non-empty? (RDF-wise an empty set means no arc, but that also makes it superfluous here.)

niklasl (approved):
LGTM! (See comment about superfluous empty components.)

rest/src/main/groovy/se/kb/libris/digi/DigitalReproductionAPI.groovy
olovy: Had a feeling I did this a bit too "clever" =P
There is always at least one hasComponent. Since this is Libris standard practice.

`(asList(item.hasComponent) ?: [[:]])` <-- Line 279, a single empty component is added if missing

niklasl: Ah, I see it now! 😉 

(Alas, that standard practise is bad; but also terribly hard to get at (as the code has been built up around that).)

PR #1090, opened 2022-02-24
Feature/lxl 3589 link agents 4

kwahlin:
Last changes now! See spec https://jira.kb.se/browse/LXL-3589

### DIFF
--- a/whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
+++ b/whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
@@ -37,6 +37,12 @@ selectByCollection('bib') { data ->
             s.termComponentList.each { tc ->
                 modified |= tryLinkAndReport(tc, id)
             }
+        } else if (s.'@type' == 'Work') {
+            asList(s.contribution).each { c ->
+                asList(c.agent).each { a ->
+                    modified |= tryLinkAndReport(a, id)
+                }
+            }
         } else {
             modified |= tryLinkAndReport(s, id)
         }
@@ -95,6 +101,7 @@ Map loadAgents() {
 }
 
 void normalize(Map agent) {
+    agent.remove('@type')
     DocumentUtil.traverse(agent) { value, path ->
         if (value in String) {
             def normValue = path.last() in Integer ? removeTrailingPeriod(value) : asList(removeTrailingPeriod(value))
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
```diff
@@ -94,14 +96,20 @@ Map loadAgents() {
                 agents[chip] << id
             else
                 agents[chip] = [id] as Set
+
+            chip = chip.findAll { it.key != '@type' }
```
lrosenstrom, line 100: Maybe do `chip.remove('@type')` for clarity?

whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
```diff
@@ -94,14 +96,20 @@ Map loadAgents() {
...
+            if (agents.containsKey(chip))
+                agents[chip] << id
+            else
+                agents[chip] = [id] as Set
+
         }
```
lrosenstrom, line 106: Do you really want to store one agent with '@type' and one without?

whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
kwahlin: Would prefer that too but it'll mutate `chip` so we wouldn't get two distinct map entries (one with `@type` and one without).

whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
kwahlin: One could probably come up with a more clever solution, I just implemented the first somewhat (time) efficient solution that came to mind :)

lrosenstrom: Sure, if it works the way you planned it's perfectly OK!

lrosenstrom (approved):
:ship:

whelktool/scripts/cleanups/2022/01/lxl-3589-link-agents.groovy
lrosenstrom: Ah, of course :+1:

PR #1095, opened 2022-03-08
CRUD: Enable all JSON-LD validation for POST

olovy:
New documents should always be valid even though we haven't cleaned up all existing data yet.

Might break "create new from file" with old documents.


lrosenstrom (approved):
Ship it!

klngwll:
Om den med "all" även kör `validateRepeatability(key, value, validation)` så kommer vi nog åka på ganska många stoppers...

PR #1096, opened 2022-03-08
Remove imageBitDepth in hold-data

klngwll:
Remove validation errors `Unknown term. Missing definition.`: Old datatype form and only `---` values in 94 hold-records.


kwahlin (approved):
:ship:

PR #1099, opened 2022-03-10
Add distinct ignore on auth fixed fields

klngwll:
Because imports...

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -22408,7 +22408,7 @@
       },
       "_spec": [
         {
-          "name": "Fixed defaults with ignored (x in originalLeader)",
+          "name": "Fixed defaults with ignored (x in source)",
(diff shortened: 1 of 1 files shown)


niklasl:
Ah, so we get unhandled values for incoming data that isn't equal to the `fixedDefault`? Do we still get that `fixedDefault` on output? (A `_spec` for just one of these would be nice to secure the behaviour if so.)

klngwll:
> Ah, so we get unhandled values for incoming data that isn't equal to the `fixedDefault`? Do we still get that `fixedDefault` on output? (A `_spec` for just one of these would be nice to secure the behaviour if so.)

It is mostly a preemptive move so we don't have to unnecessary clean up bad sourcedata since there is talk about importing external MARC-auths etc. Good call on the specs, I realized 000/6 "recordtype" behaved awfully weird and injected postprocessing from hold etc so better let that one produce a failedFixedField if it ever would occur.

niklasl (approved):
Great preemptive protections! :ship:

PR #1101, opened 2022-03-17
Search: Use dfs_query_then_fetch 

olovy:
Two things
* Use dfs_query_then_fetch 
* Fix a bug where search_after would query all indices instead of a specific one

Use normalized index statistics: dfs_query_then_fetch [1]

Fix search relevancy issue. prefLabel is boosted alot for concepts.
Since each term in preflabel is pretty rare across all indexed
documents the idf value for the term might vary alot across shards.

Specifically, "n, number of documents containing term", seems to get
skewed for a while after a document has been updated alot?
For instance when reverseLinks.totalItems is updated.

Example
prefLabel:sverige

sao/Sverige--Östergötland
idf: n=74, N=12116
```
{
  "value": 5.0915656,
  "description": "idf, computed as log(1 + (N - n + 0.5) / (n + 0.5)) from:",
  "details": [
    {
      "value": 74,
      "description": "n, number of documents containing term",
      "details": []
    },
    {
      "value": 12116,
      "description": "N, total number of documents with field",
      "details": []
    }
  ]
}
```

sao/Sverige
idf: n=315, N=12019
```
{
  "value": 3.6401684,
  "description": "idf, computed as log(1 + (N - n + 0.5) / (n + 0.5)) from:",
  "details": [
    {
      "value": 315,
      "description": "n, number of documents containing term",
      "details": []
    },
    {
      "value": 12019,
      "description": "N, total number of documents with field",
      "details": []
    }
  ]
},

```

See LXL-3837.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -494,8 +494,12 @@ class ElasticSearch {
         }
     }
 
-    private String getQueryUrl(filterPath = []) {
-        def url = "/${indexName}/_search?search_type=$SEARCH_TYPE"
+    private String getQueryUrlWithoutIndex(filterPath = []) {
+        getQueryUrl(filterPath, null)
+    }
+    
+    private String getQueryUrl(filterPath = [], index = indexName) {
+        def url = (index ? "/${index}" : '') + "/_search?search_type=$SEARCH_TYPE"
         if (filterPath) {
             url += "&filter_path=${filterPath.join(',')}"
         }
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -600,7 +605,7 @@ class ElasticSearch {
 
         @Override
         String queryPath() {
-            "/${indexName}/_search?filter_path=$filterPath"
+            "/${indexName}/_search?search_type=$SEARCH_TYPE&filter_path=$filterPath"
```
niklasl, line 608: Can this just call `getQueryUrl` now? ("Eliminate moving parts" or are they "accidentally equal" right now?)

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -657,7 +662,9 @@ class ElasticSearch {
 
         @Override
         String queryPath() {
-            "/_search?filter_path=$filterPath"
+            isPitApiAvailable // point in time is created on index and then index cannot be specified here 
+                    ? "/_search?search_type=$SEARCH_TYPE&filter_path=$filterPath" 
+                    : "/${indexName}/_search?search_type=$SEARCH_TYPE&filter_path=$filterPath"
```
niklasl, line 667: Same question: call `queryPath()` here? :thinking:

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
olovy: Let's say they're unaccidentally equal

niklasl (approved):
LGTM! :ship: (Then we'll watch QA for changes in query scores *and* if speed is unexpectedly affected.)

PR #1103, opened 2022-03-17
Refactor Crud::getFormattedResponseBody

olovy:
Minor restructuring. Don't modify response inside `getFormattedResponseBody`.
Fix 400 Bad Request on missing profile.


niklasl (approved):
Much better! :+1:

PR #1105, opened 2022-03-22
Use empty prefix for KBV

kwahlin:
Set `:` as the prefix for KBV in Virtuoso


niklasl (approved):
Good catch! I think we should keep `kbv:` as well though, to let the user decide which level of explicit to use in queries.

kwahlin:
Fair enough!

PR #1106, opened 2022-03-23
DELETE instead of TRUNCATE embellish cache.

jannistsiroyannis:
Truncating a table, while faster, is beleived to be problematic
because an exclusive lock on the table is necesseray. Obtaining
this lock while other transations are active may be the reason for
DB lock ups while trying to copy data to another environment
(this is however unproven).

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -134,7 +134,7 @@ class PostgreSQLComponent {
     private static final String GET_TABLE_SIZE_BYTES =
             "SELECT pg_total_relation_size(?)"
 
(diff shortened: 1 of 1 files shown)


niklasl:
We need to test this properly, right? Does a `DELETE` also obtain a lock, or is it "one by one"?

andersju:
"TRUNCATE acquires an ACCESS EXCLUSIVE lock on each table it operates on, which blocks all other concurrent operations on the table. ... If concurrent access to a table is required, then the DELETE command should be used instead." (https://www.postgresql.org/docs/9.6/sql-truncate.html#AEN93272)

niklasl (approved):
LGTM! This won't help us during _this_ release, but presuming it has no unexpected drawbacks I suppose we'd like to get it into this release to have an easier time reading from PROD from thereon on. If so, we need to cherry-pick this into the release branch.

PR #1109, opened 2022-04-19
Add rate limiting to CRUD and search APIs

olovy:


rest/src/main/java/whelk/util/RateLimiter.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class RateLimiter {
+    private final int UPDATE_PERIOD_MS = 100;
+    
+    // Make the calculations work with really low max rates
+    private final long SCALE_FACTOR = 1000;
```
olovy, line 11: a bit ugly...

PR #1111, opened 2022-04-28
Add the possibilty of having a priority list with add-if-none merge r…

jannistsiroyannis:
…ules.

If there is such a list, the priority of the owner of the entity _to which
we're adding_ something is compared to the priority of whatever source
we're currently importing from.

If there is no priority list, the add-if-none rule will work as before.
That is, any source may add data at the specified path.


olovy (approved):
:ship:

PR #1112, opened 2022-05-02
Add gtin-14 Identifier normalization

klngwll:
Normalize whitespace?


kwahlin (approved):
LGTM

kwahlin:
> Normalize whitespace?

Sounds like a good idea. At least that's how all the gtin-14 values we have so far looks like.

niklasl (approved):
LGTM

PR #1119, opened 2022-05-16
Copy more details about original to digitization

niklasl:
Includes handling scalar values in CopyOnRevertStep.


olovy (approved):
:ship:

klngwll (changes requested):
Remove `hasNote` from copyIfMissing. Decision has been made to not copy notes between Instances. Otherwise LGTM.

PR #1121, opened 2022-06-01
Add a cleanup script.

jannistsiroyannis:


whelktool/scripts/cleanups/2022/06/lxl-3859-barn-ungdom.groovy
```diff
@@ -0,0 +1,51 @@
...
+        if (subject["@type"] == "ComplexSubject" && subject.inScheme?.containsKey("@id") && subject.inScheme["@id"] == "https://id.kb.se/term/barn") {
+            if (subject.termComponentList.removeAll { termComponent ->
+                return termComponent["@type"] == "GenreSubdivision" && termComponent.prefLabel == "barn- och ungdomslitteratur"
+                }) {
+                changed = true
+            }
+
+            // Only the main term still remaining? - Remove the whole subject
+            if (subject.termComponentList.size() == 1) {
+                subjectToPreserve = subject.termComponentList[0]
```
niklasl, line 29: Shouldn't all `subjectsToPreserve` (plural) be collected here (in a `Set<Map>`)? Otherwise, it seems that all but the last subject in e.g. https://libris.kb.se/7rkl60dk2q1r7tl#it would be dropped?

jannistsiroyannis: Yepp! Good catch! :)

niklasl (approved):
LGTM!

PR #1122, opened 2022-06-02
Feature/extended dataset importer

niklasl:
Make DatasetImporter capable of importing both "pre-cooked" json.lines and regular RDF (currently only Turtle, *real* TriG dataset support is now easily added if we need it). This replaces the separate `LinkedDataImporter` sublass and is thus capable of  creating records with predictably minted URI:s from simple resource descriptions (one record per resource). 

Details include:

* Merge LinkedDataImporter into DatasetImporter.
* Instantiate a DatasetImporter and use instance methods for processing.
* Add DatasetInfo to encapsulate Dataset specifics and its URI minting.
* Use sourceUrl to determine whether JSON lines or "one record per resource" mode is to be used.
* LD-normalize all non "pre-cooked" jsonld lines imports (this expands and re-compacts using the system context via TargetVocabMapper).
* Either load dataset description from file, or expect it to be the first description in the source input dataset.
* Use CacheRecord if sourceUrl is an HTTP(S) resource.
* Add flag support to ImporterMain CLI for additional control.


jannistsiroyannis (approved):
I've looked it over (quickly) and nothing jumped out! :)
:ship:

importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
```diff
@@ -62,24 +111,167 @@ class DatasetImporter {
...
+    }
+
+    protected Document completeRecord(Map data, String recordType) {
+        if (GRAPH !in data) {
+            data = makeSystemRecordData(data, recordType)
+        }
+
+        if (REVERSE in data) {
+            Map revs = data.remove(REVERSE)
+        }
```
olovy, line 158: This doesn't do anything since data is always `@graph`?

`@reverse` is always removed here anyway
https://github.com/libris/librisxl/blob/cba147b8218c7ddd8d28daf59a934339773da6a4/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy#L2322

niklasl: Ah, right, I've cobbled up what `data` is here. Also certainly redundant; I'll remove this.

olovy (approved):
LGTM! 

Let's add some documentation to the xlimporter commands.

PR #1123, opened 2022-06-07
Invalidate dependency cache of all whelk instances on write

olovy:
Send a notification to all other whelk instances to invalidate affected parts of DependencyCache when a document is saved. PostgreSQLs LISTEN/NOTIFY is used as a simple IPC mechanism.

This uses one more physical database connection per whelk instance.

Here is a previous iteration that instead sends the ids of the previous and current document versions and lets each listener calculate which links have been added and removed https://github.com/libris/librisxl/compare/feature/clear-deps-cache

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -2663,11 +2663,11 @@ class PostgreSQLComponent {
 
                 def (oldVersion, newVersion) = [docs[0], docs[1]]
                 if (oldVersion && newVersion) {
-                    log.info("New doc version {}, invalidating dependencyCache.", newVersion.shortId)
+                    log.info("Document {} updated, invalidating dependencyCache.", newVersion.shortId)
                     dependencyCache.invalidate(oldVersion, newVersion)
                 }
                 else if (newVersion) {
-                    log.info("New doc version {}, invalidating dependencyCache.", newVersion.shortId)
+                    log.info("Document {} created, invalidating dependencyCache.", newVersion.shortId)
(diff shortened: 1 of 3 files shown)


olovy:
Ping @lrosenstrom @niklasl @andersju

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2554,6 +2568,98 @@ class PostgreSQLComponent {
             }
         }
     }
+    
+    enum NotificationType {
+        dependency_cache_invalidate // LISTEN silently fails to register if upper case(???)
```
andersju, line 2573: Given https://stackoverflow.com/a/5173993 I'm guessing the channel name here is _not_ forcibly lowercased
```
try (PreparedStatement statement = getMyConnection().prepareStatement('SELECT pg_notify(?, ?)')) {
    statement.setString(1, type.toString())
```
while here it _is_
```
statement.execute("LISTEN ${t.toString()}")
```
?

andersju (approved):
Nice! LGTM. Let's try it on dev!

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
olovy: Ah!

I'll make that clear.

olovy: Forgot to push it here....
On `develop` now https://github.com/libris/librisxl/commit/8a047738c4ae3ac9cbfd75d209009a6cd697fef8

PR #1125, opened 2022-06-09
Limit depth of frame/embed

olovy:
For very connected datasets like Queerlit where many Concepts are `related` to many other Concepts in the same embellish graph there is a combinatorial explosion leading to huge framed documents (which are later thrown away).

We still want e.g. the queerlit concept scheme definition to be embedded in all places in the result so just add a limit on the number of levels where we start with a new "embedChain".

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLd.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLd.groovy
@@ -1179,23 +1179,23 @@ class JsonLd {
         }
     }
 
-    private static Map embed(String mainId, Map mainItem, Map idMap, Set embedChain) {
+    private static Map embed(String mainId, Map mainItem, Map idMap, Set embedChain, int depth = 0) {
         embedChain.add(mainId)
         Map newItem = [:]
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -1209,7 +1209,9 @@ class JsonLd {
                 obj = fullObj ? [:] + fullObj : null
             }
             if (obj) {
-                return embed(oId, obj, idMap, new HashSet<String>(embedChain))
+                return depth < 2
+                        ? embed(oId, obj, idMap, new HashSet<String>(embedChain), depth + 1)
```
niklasl, line 1213: This counts all depths equally, right? It doesn't take e.g. `:QualifiedRole`- or `:StructuredValue`-based entities into account, nor `:category :integral` on the properties? I suspect `2` may be too low in general, since we do have chains like `relatedTo.instanceOf.hasTitle.mainTitle` (and various `instanceOf`-indirected important parts (`subject` et al))... :thinking:

PR #1126, opened 2022-06-10
Add force option to PostgreSQLComponent.remove

niklasl:
This is used via the `--force-delete` flag to DatasetImporter to delete records in a dataset even if there are links to them. This is sometimes necessary to break (cyclic) cross-references within a dataset.

(Currently only needed to refresh qlit in QA. We may want to cherry-pick this into a hotfix if we require dataset-imports to run from the master branch.)

### DIFF
--- a/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
+++ b/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy
@@ -29,12 +29,15 @@ class DatasetImporter {
     // verify that id:s are served by the system; else use:
     static REPLACE_MAIN_IDS = 'replace-main-ids' // replace id with XL-id (move current to sameAs)
 
+    static FORCE_DELETE = 'force-delete'
+
     Whelk whelk
     String datasetUri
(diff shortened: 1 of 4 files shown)


olovy:
LGTM. 
We just have to be aware that this also breaks the "integrity" of any links from outside the dataset.

Could change `force` to take a list of URI prefixes but I think that would complicate the code unnecessarily.

olovy:
Another option would be that the importer delete step first removes all links to the same dataset from the records to be deleted in an update.

niklasl:
I agree that there may be a integrity-preserving variant such as you suggest, but I suppose right now it boils down to a  "I know what I'm doing when using --force-delete". With the "two pass" option you suggest (which I like) a "force" oughtn't be needed anyway. It should already work like that. So I presume we'll chalk it up as a future improvement ([the code does mention this](https://github.com/libris/librisxl/blob/95775b6676731e1e5b075c43405433c0887f66e9/importers/src/main/groovy/whelk/importer/DatasetImporter.groovy#L291)) and merge this as is.

PR #1131, opened 2022-06-24
Pull request for NatLibFi changes

zds-kansalliskirjasto:
I profiled the batchimport to figure out why it is slow with our data, and found out it almost busy-loops on Thread.yield() which causes a lot of system calls. It also created a new OS-level thread for each task, which again causes system calls and is somewhat expensive.

I changed the import to utilize the thread pool implementation that comes with Java standard libraries and it reduced amount of time spent in system calls by factor of seven. Total execution time didn't drop much as this test was run on otherwise idle developer laptop, but on busy system system calls can become a bottleneck and slow the whole system down.

Here are timings run without profiler:

Original:
real	3m57.395s
user	5m17.832s
sys	2m27.186s

ExecutorService:
real	5m6.852s
user	4m44.039s
sys	0m21.946s

The other changes are more minor, ignoring Eclipse files, fixing half-broken Log4J file logger setup and making it easier to see from the logs where a NPE coming from wrong config originates from.

Screenshot at 2022-06-17 15-26-52_crop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,12 @@ Session.vim
 /*/build
 /*/catalina.base_IS_UNDEFINED
 /*/logs
+
+# Eclipse meta files
 /eclipse-classes/
+.classpath
+.project
+.settings
+
+# Local overrides to Log4J configuration etc.
 /local-overrides/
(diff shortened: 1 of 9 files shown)


klngwll:
Thanks Jari for letting us know about your performance-findings. we probably can´t vet this properly until after vacations but I have added the backenders as reviewers on this.

niklasl (approved):
Looks good to me! (I'll let the others chime in after vacations.)

PR #1132, opened 2022-06-29
Feature/working core unit tests

niklasl:
Apparently these stopped running quite some time ago (presumably during a Gradle upgrade after which JUnit has to be explicitly enabled for these to run).

Thus some subsequent changes caused their specs to fail. Mostly innocent ones (with simple spec fixes), apart from two (of my doing), where I had to fix the implementations (tiny fixes, but still).

So not only do tests never verify the _absence_ of bugs, not noticing they're no longer running won't enable the verifying of their _presence_ either.... :grimacing: 

(To save face regarding my blunders I _could_ claim that had the tests been run those blunders would have been smoked out. But not diligently beginning with negative tests would have smoked out the non-running of the tests, so there ain't much to save.)


lrosenstrom (approved):
Good work! :ship:

PR #1134, opened 2022-07-07
Partition marcframe JSON into separate files

niklasl:
Also factor out serving of said marcframe data into a servlet specifically for serving this data in full (with included files).


olovy:
LGTM!

ETag header is not set anymore. I that intended?

PR #1135, opened 2022-07-21
Feature/summary to instance

kwahlin:
This script moves summaries not containing language down to instance for _all_ Text works.

### DIFF
--- a/whelktool/scripts/cleanups/2022/07/lxl-3303-summary-to-instance.groovy
+++ b/whelktool/scripts/cleanups/2022/07/lxl-3303-summary-to-instance.groovy
@@ -7,18 +7,14 @@ def where = """
 
 def badShape = ['@type', 'label', 'language'] as Set
 
-selectByCollection('bib') { bib ->
+selectBySqlWhere(where) { bib ->
     def instance = bib.graph[1]
     def work = instance.instanceOf
-
-    if (!work || work instanceof List)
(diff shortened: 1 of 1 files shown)


olovy (approved):
LGTM!

niklasl:
Has the marcframe logic been adapted to this requirement?

klngwll:
> Has the marcframe logic been adapted to this requirement?

Yes default for conversion was made Instance a year back or so, but with a fallback entity for works. We have just lagged in the actual moving of them.

niklasl:
> > Has the marcframe logic been adapted to this requirement?
> 
> Yes default for conversion was made Instance a year back or so, but with a fallback entity for works. We have just lagged in the actual moving of them.

Ah,  great; then all is in sync! (I faintly recalled "something amiss"; which is what this PR addresses!)

klngwll:
Did we decide to move all other summaries than the bad shapes. I thought we were only moving if they also were from certain import channels? My memory need a refresh here :)

kwahlin:
> Did we decide to move all other summaries than the bad shapes. I thought we were only moving if they also were from certain import channels? My memory need a refresh here :)

Yes, we discussed this in a format team meeting (22/6) and decided to move all the summaries. But we can discuss it again in the next meeting to fully establish the decision before we go through with any changes.

PR #1136, opened 2022-08-09
Use persistent user id as key for user data instead of email

olovy:
...address which can now be changed

Depends on https://github.com/libris/xl_auth/pull/250
Related https://github.com/libris/lxlviewer/pull/809

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
@@ -40,9 +40,9 @@ class UserDataAPI extends HttpServlet {
         Map userInfo = request.getAttribute("user")
         if (!isValidUserWithPermission(request, response, userInfo))
             return
-
-        String id = userInfo.email.digest(ID_HASH_FUNCTION)
-        String data = whelk.getUserData(id) ?: "{}"
+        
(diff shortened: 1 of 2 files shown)


rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
```diff
@@ -112,13 +132,13 @@ class UserDataAPI extends HttpServlet {
...
-            log.info("User check failed: 'email' missing in user")
-            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Key 'email' missing in user info")
+        if (!userInfo.containsKey("id")) {
+            log.info("User check failed: 'id' missing in user")
+            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Key 'id' missing in user info")
             return false
         }
-
-        String id = userInfo.email.digest(ID_HASH_FUNCTION)
+2
```
andersju, line 140: :point_up: :two:

andersju (approved):
LGTM! (note the stray '2')

rest/src/main/groovy/whelk/rest/api/UserDataAPI.groovy
olovy:
```suggestion
```

PR #1137, opened 2022-08-09
Validate @id in JsonLdValidator

andersju:
Ensures @id is a valid IRI. Uses Apache Jena's IRI validation (which we're already using in JsonLdToTrigSerializer).


olovy (approved):
LGTM!

PR #1138, opened 2022-08-17
Put scripts in separate files to run normally with whelktool

kwahlin:
- Methods common to several scripts are put in the Util class
- The scripts are now grouped together as a package. Maybe not a correct solution but I think they should be grouped together somehow.
- Not sure if the variables in Util are correctly declared


whelktool/src/main/groovy/datatool/scripts/mergeworks/Util.groovy
```diff
@@ -4,9 +4,13 @@ import org.apache.commons.lang3.StringUtils
 import whelk.Whelk
 import whelk.util.Unicode
 
-import java.util.regex.Pattern
-
 class Util {
+    static def whelk = Whelk.createLoadedCoreWhelk()
```
olovy, line 8: This creates an extra whelk instance which is unnecessary.

whelktool/src/main/groovy/datatool/scripts/mergeworks/Util.groovy
```diff
@@ -4,9 +4,13 @@ import org.apache.commons.lang3.StringUtils
...
+
+    static def clusters = 'clusters'
+
+    static def contributionPath = ['@graph', 1, 'instanceOf', 'contribution']
```
olovy, line 12: I'm not sure think we need these as constants for the scripts.
(And they should be uppercase in that case, and not follow my bad example :wink:)

olovy (approved):
See comments

PR #1139, opened 2022-08-17
Search: Improve queries on nested documents

olovy:
* Get which fields are "nested" from ES mappings instead of hardcoding them
* Make some queries work
  * Repeated parameters [(LXL-265)](https://jira.kb.se/browse/LXL-265)
    * `?identifiedBy.@type=ISBN&identifiedBy.value=1234&identifiedBy.@type=LCCN&identifiedBy.value=5678`
    * `?identifiedBy.@type=ISBN&identifiedBy.@type=LCCN&identifiedBy.@type=NBN`
  * More than two properties inside nested document
    *  `?hasTitle.@type=Title&hasTitle.mainTitle=städer&hasTitle.subtitle=storstadspolitik`

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -499,13 +499,11 @@ class ESQuery {
         handledParameters.each {queryParametersCopy.remove(it)}
         filters.addAll(rangeFilters)
         
-        // TODO This is copied from the old code and should be rewritten.
-        // We don't have that many nested mappings, so this is way to general.
         Map groups = queryParametersCopy.groupBy { p -> getPrefixIfExists(p.key) }
         Map nested = getNestedParams(groups)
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
```diff
@@ -570,28 +571,40 @@ class ESQuery {
     }
 
     @CompileStatic(TypeCheckingMode.SKIP)
-    private Map getNestedParams(Map groups) {
-        // TODO We hardcode `identifiedBy` here, since that's currently the only
-        // type of search in the client where we need nesting.
```
olovy, line 575: `libris_config.json` actually also defined `hasTitle` as nested ¯\_(ツ)_/¯

lrosenstrom (approved):
Looks useful! IMO we need to document this somehow and make it clear that only fields that are defined as 'nested' internally are searchable in this way. (Preferably without using that word, which only makes sense with prior knowledge of Elasticsearch).

PR #1140, opened 2022-08-18
Add script for fetching and moving 

kwahlin:
With some minor changes from the original script https://github.com/libris/librisxl/blob/feature/merge-works/whelktool/scripts/cleanups/2021/09/lxl-3376-elib-cover-designer.groovy


olovy (approved):
:ship:

PR #1144, opened 2022-08-22
Return 409 Conflict (instead of 400) on ConflictingHoldException

andersju:
Makes it possible for lxlviewer to show a nicer error message.


olovy (approved):
:ship:

PR #1147, opened 2022-08-30
Feature/handle linked works

kwahlin:
This makes it possible to run "verkskörningen" repeatedly and create new works as well as link instances to already existing works.

[The very first step](https://github.com/libris/librisxl/blob/feature/merge-works/whelktool/scripts/analysis/find-work-clusters.groovy) is still the same, however after the first run we should only use newly created/modified docs as a starting point to find clusters.

The "title cluster" step is also pretty much the same, the only difference is that duplicate linked works are filtered out from the cluster before doing the partitioning, i.e. only one instance/doc per linked work is saved.

Then we partition title clusters into "work clusters" (cluster = same work). This is also done the same way as before, but since we now have linked works in the mix we get three different possibilities for each cluster:
-  It contains only local works -> Create new linked work (merge local works) and link the instances to this.
-  It contains one linked work -> Merge the local works with the linked one and link the instances to the updated work. We could also skip the merging and just link the instances to the existing work, that is if we don't want to "enrich" existing works with data from new/updated local works.
-  It contains more than one linked work -> Log and handle manually? Should be very unusual that two linked works are considered to be the same but it _can_ happen if either is modified by e.g. a cataloguer.

### DIFF
--- a/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
+++ b/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
@@ -14,6 +14,7 @@ import whelk.util.DocumentUtil.Remove
 import static whelk.JsonLd.GRAPH_KEY
 import static whelk.JsonLd.ID_KEY
 import static whelk.JsonLd.TYPE_KEY
+import static whelk.JsonLd.WORK_KEY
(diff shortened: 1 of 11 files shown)


olovy (approved):
:ship:

PR #1148, opened 2022-08-30
Fix a list index corruption when generating history.

jannistsiroyannis:
When comparing two lists, all elements that are (identical) in both lists
are filtered out so that the remaining different elements can be compared
in the correct order.

However, doing the comparisons of the elements that remain, their respective
list indexes have changed (in the above mentioned filtering). So in order to
report differences at _the correct path_ we must pass on the _original_ list
indexes rather than the modified temporary indexes.


lrosenstrom (approved):
:ship: !

PR #1152, opened 2022-09-14
Fix and improvement to the ExecutorService parallelization

zds-kansalliskirjasto:
I recently changed Whelk ThreadPool to utilize Java standard ExecutorService.

However, Whelk uses some unconventional messaging paths for code flow control, and I missed one such item.

Whelk.batchJobThread() checks for thread group name and the exact name affects the path of execution. However, as string literal is used and not constant, unless you accidentally run into this one line method you can easily miss it.

I now changed the literals to a constant and changed the ExecutorService implementation to set the name.

I also added a command line switch -t for the reindex operation to set the level of parallelizatrion.

Then I tested on a beefy server on various values for -t to see that my fix improves the perfornamce, and indeed it does.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ Session.vim
 
 # Local overrides to Log4J configuration etc.
 /local-overrides/
+
+# Logs
+/logs/
--- a/apix_server/.gitignore
+++ b/apix_server/.gitignore
@@ -1,0 +1,1 @@
+/eclipse-classes/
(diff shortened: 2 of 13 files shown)


olovy (changes requested):
:+1: 

I guess you intended to change this line to use the constant as well:
https://github.com/libris/librisxl/blob/70487c504b5ba42809a34e3df7315dfd047b18cd/whelk-core/src/main/groovy/whelk/Whelk.groovy#L476

I suggest changing the constant to for example 
```
public static final String BATCH_THREAD_GROUP = "whelk-batch";
```
since it is now used for more things than whelktool

PR #1153, opened 2022-09-14
Add script for unlinking some bad agents.

jannistsiroyannis:


niklasl:
Is LXL-3919 the correct issue for this?

jannistsiroyannis:
> Is LXL-3919 the correct issue for this?
Urgh.. nope :P It's LXL-3913

jannistsiroyannis:
I also have a sneaking suspicion that maybe we shouldn't just remove them, but replace them with local entities?

niklasl:
Ah, check. Yes, its only the linkage that's to be removed; "some" local entity data should be kept (presumably only names and possibly lifeSpan (albeit that might be wrong if added on the linked entity after original match)). Ask for clarification/spec of fields.

whelktool/scripts/cleanups/2022/09/lxl-3919-unlink-agents.groovy
```diff
@@ -0,0 +1,113 @@
...
+    if (changed) {
+        //System.err.println("result:\n" +initial + "\nchanged into:\n"+ data.doc.getDataAsString() + "\n\n")
+        data.scheduleSave()
+    }
+}
+
+boolean isABadLink(String candidate) {
+    String s = candidate.substring(0, candidate.length()-3) // Trim off the #it
+    System.err.println(candidate + " -> " + s)
+    return s.endsWith("53hlt8kp5swj700") ||
```
niklasl, line 68: Nit-pick: safer (and slightly faster) to put these in a `Set` and use that here and to build the select string.

jannistsiroyannis: To use a Set I'd instead have to shave of the https://libris{-qa,-stg}/ (which varies in length), which could go wrong. I think I'll keep this as is!

niklasl: Or use `candidate.substring(candidate.lastIndexOf('/') + 1, candidate.lastIndexOf('#'))`?

kwahlin: `candidate.split('[#/]')[-2]` should work too!

niklasl: Yes, that is certainly much more succinct (albeit somewhat less performant, as it would parse a regexp (unless cached under the hood), splitting on all `/`:s and building a new list of those substrings; though I bet any performance loss is dwarfed by the I/O going on here).

PR #1154, opened 2022-09-15
Move "pliktex" information for Li holdings

poacherone:
Move information from hasNote to immediateAcquisition, for Li holdings that contain the string "Pliktex". For more information see LGC-76.


kwahlin (approved):
Looks good!

PR #1155, opened 2022-09-16
Add `asList()` and `getAtPath()` to whelktool API

olovy:
They are used (copy-pasted) in a lot of scripts.

Existing scripts continue to work as before since "bindings" are shadowed by methods defined in script.


niklasl (approved):
LGTM! :shipit:

PR #1157, opened 2022-09-26
Check for identical versions _after_ final adjustments of records hav…

jannistsiroyannis:
…e been done.

This fixes a bug where an incoming record would be considered not-identical to the existing record (being updated) because the comparisons were done _before_ the final fixup, specifically before the normalizeDocumentForStorage() call. This meant that for example incoming records with local entities about to be linked (during the normalize) call would be overwritten with a new identical version.

This moves the comparison to _after_ the normalizeDocumentForStorage() call, and makes the behaviour controllable through the "writeIdenticalVersions" boolean. As a side effect, the  behaviour (not overwriting) now becomes an availale option to other code trying to save records and in one place (APIX) that option will now be exercised, as that is more appropriate behaviour even though overwriting was tolerated before.


lrosenstrom (approved):
Looks good to me!

PR #1159, opened 2022-10-07
Make batchimport integTest work without bundled secret.properties

olovy:
And on environments where psql requires a username.

### DIFF
--- a/batchimport/integtest/test.sh
+++ b/batchimport/integtest/test.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -eux
+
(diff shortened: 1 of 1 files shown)


batchimport/integtest/test.sh
```diff
@@ -1,14 +1,79 @@
 #!/bin/bash
```
niklasl, line 2: `set +ex`? (Could imagine even a trap cleanup, but then, if _that_ fails... :arrows_counterclockwise: )

niklasl (approved):
LGTM! (I'm _assuming_ here that noone will accidentally think it's wise to run these using prod secrets... :grimacing: )

PR #1163, opened 2022-10-12
MX-11 

zds-kansalliskirjasto:
Currently only Groovy side logs via the logging framework and Java does not. This means informal DEBUG level messages are persisted whereas fatal errors are epheremal. Here I'm changing Main.java log uncaught fatal problems via the same framework as Groovy already does.


olovy:
LGTM!

PR #1167, opened 2022-10-20
Include alternateProperties within showProperties

niklasl:
Used for calculating boost from a class lens.

If this works well on QA we'll cherry-pick it into a hotfix.


andersju (approved):
:rocket: LGTM, let's try it!

PR #1168, opened 2022-10-26
Make it possible to sync datasets without refreshing dependers on upd…

jannistsiroyannis:
…ated records.

This is necessary in order to sync some datasets that see extremely heavy use, like "languages".

Doing this has drawbacks however. If you sync a dataset this way, it's up to you to make sure the changed data should not have affected the dependency table, and also to reindex the dataset (and all it's dependers!) if and when that is necessary.

All of this was equally true for the old definitions importer as well.

Additionally: Fix the counters in the dataset importer (updated/created/deleted). These had rotted with other changes to the core system.

Additionally: Make it an explicit rule that changes brought by syncing of "definitions" (meaning any datasets!) do not ever trigger MARC21 or OAIPMH exports under any circumstances.

Finally this STILL does not make it possible to completely replace the old definitions importer, because the dataset loader relies on the normal write-record code paths, which in turn _need_ some definitions data (vocab?) to function. This puts us in a chicken-and-egg type situation.

The dataset importer could now however be used to _update_ all definitions data.

### DIFF
--- a/apix_server/src/main/java/whelk/apixserver/ApixCatServlet.java
+++ b/apix_server/src/main/java/whelk/apixserver/ApixCatServlet.java
@@ -208,7 +208,7 @@ public class ApixCatServlet extends HttpServlet
                 id = Utils.s_whelk.getStorage().getSystemIdByIri("http://libris.kb.se/" + collection + "/" + id);
 
             incomingDocument.deepReplaceId( Document.getBASE_URI().resolve(id).toString() );
-            Utils.s_whelk.storeAtomicUpdate(id, false, false, Utils.APIX_SYSTEM_CODE, request.getRemoteUser(),
+            Utils.s_whelk.storeAtomicUpdate(id, false, false, true, Utils.APIX_SYSTEM_CODE, request.getRemoteUser(),
                     (Document doc) ->
                     {
                         doc.data = incomingDocument.data;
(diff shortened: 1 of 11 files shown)


niklasl:
Good progress! Just a quick confirmation: the chicken-egg problem is real, and anticipated. We will divide "definitions" into a bunch of regular datasets, some of which are still crucial for daily operation, such as languages. These will all be loaded by this improved DatasetsImporter.

The remaining definitions dataset should then be the "vocab" core dataset (containing vocab, context and display), which is the initial "domain seed". An empty whelk *must* be started with that dataset (we may even want to reference its data file in `secret.properties` and load it upon initial startup, and update it upon system upgrade).

niklasl:
@jannistsiroyannis Haven't checked in practise yet, just lazily asking: is it the case that even with `--skip-dependers`, the updated (and new) records themselves reindexed (i.e. only their dependents are skipped)? If i read correctly that is so, and that is great! Examples of why include:

* We update the languages dataset and want to index some detail (e.g. a new altLabel for an existing language, or a new language which is to be findable by its labels). It would be good not to have to reindex them separately, but of course not have all their dependents reindexed (there will be no depdentents on new languages of course).

* With that in effect we can reload bibdb-data using `--skip-dependers`. That should speed it up by some magnitudes!

jannistsiroyannis:
> @jannistsiroyannis Haven't checked in practise yet, just lazily asking: is it the case that even with `--skip-dependers`, the updated (and new) records themselves reindexed (i.e. only their dependents are skipped)?
Yes! That is correct! :)

PR #1169, opened 2022-10-27
Do all elastic indexing asynchronously

olovy:
(except for batch jobs, like before)

HTTP requests to elastic sometimes hang. These hanging requests would
cause POST or PUT to our CRUD API to block for 60s (SO_TIMEOUT/READ_TIMEOUT
in ElasticClient). These would timeout in API clients or the proxy even if
everything was finally successful. Leading some clients to retry and creating
duplicates in case of POST.

Before this change the document being created/updated/removed was
indexed synchronously and other affected documents asynchonously.
Do all elastic indexing in the background instead.

This also fixes a similar problem when the main URI of a resource is
changed.

Also reduce SO_TIMEOUT for indexing of single docs to 15s so that we retry
hanged requests sooner.

---- 
Also:
Fix an old problem where trying to update the reverse link counter on a deleted document would be retried indefinitely.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -195,10 +195,9 @@ class Whelk {
                 .collectEntries { id, doc -> [(idMap.getOrDefault(id, id)) : doc]}
     }
 
-    private void reindex(Document updated, Document preUpdateDoc) {
-        if (elastic && !skipIndex) {
+    private void reindexUpdated(Document updated, Document preUpdateDoc) {
+        indexAsyncOrSync {
             elastic.index(updated, this)
-
             if (hasChangedMainEntityId(updated, preUpdateDoc)) {
                 reindexAllLinks(updated.shortId)
             } else {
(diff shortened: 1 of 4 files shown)


niklasl (approved):
LGTM! (Did the crud tests pass? We should thoroughly test this on QA too.)

(Also, nothing is specifically CRUD-related in the code, but I presume that is the only process where `isBatchJobThread()` is false?)

Finally, I believe this has precedence (and must be cherry-picked into a hotfix now); so we'll have to adapt #1168 when this lands.

jannistsiroyannis:
This looks fine, but can we please merge it _after_ https://github.com/libris/librisxl/pull/1168 ?
Because resolving conflict between the two looks like it might be a mess, and I'd rather merge these smaller changes onto the larger ones than the other way around.

olovy:
> LGTM! (Did the crud tests pass? We should thoroughly test this on QA too.)
Had to [add a small delay to them](https://github.com/libris/lxl_api_tests/commit/44b98a8c3cbbdb66d0a802eadeaddfb235ba553c) :see_no_evil:. Ran them 50 on my machine :see_no_evil:, seems to works fine.

> (Also, nothing is specifically CRUD-related in the code, but I presume that is the only process where `isBatchJobThread()` is false?)
That was a sloppy way to put it, will update the description.

> Finally, I believe this has precedence (and must be cherry-picked into a hotfix now); so we'll have to adapt #1168 when this lands.
I will fix the conflicts in #1168

niklasl (approved):
Still looks good to go (to me)!

PR #1172, opened 2022-11-10
Feature/detailed history again

jannistsiroyannis:
This PR clears up what was previously a confusion around the double meaning of "path" in the context of change sets.

From now on, it is unambigiously the case that removedPaths strictly refer to the path of the removed object *in the previous version*, while addedPaths strictly refer to the path of the added object *in the new/current version*.

These were previously confused, which did not matter for 99% of cases, but made difficult cases where different things were being removed and added within the same list impossible.

This WILL require frontend changes to compensate for, but hopefully they will be minor @lrosenstrom

### DIFF
--- a/whelk-core/src/main/groovy/whelk/history/History.java
+++ b/whelk-core/src/main/groovy/whelk/history/History.java
@@ -131,7 +131,7 @@ public class History {
         if (m_lastVersion == null) {
             m_pathOwnership.put( new ArrayList<>(), new Ownership(version, null) );
         } else {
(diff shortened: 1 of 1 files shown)


andersju (approved):
Nice, LGTM! (Possibly we need to open a new issue for the frontend changes, and merge this one + frontend fixes in tandem? If it's not a super quick fix? @lrosenstrom )

niklasl (commented):
Are there any unit tests in place (to see how it works, and to facilitate maintenance)?

PR #1174, opened 2022-11-10
Feature/extend language linker

kwahlin:
New features:

- Concatenated language labels, e.g. "Svenska & engelska" will now be linked.
- Option to arbitrarily specify links to be used for disambiguation. When ambiguous, the linker will check for a linked sibling as before but if none can be found it will now go on and check the specified links.
- More label mappings through `hiddenLabelByLang`/`hiddenLabel`.
- Linker objects accessible from whelk instance, e.g. `whelk.normalizer.normalizers.find { it.getLinker() instanceof LanguageLinker }`. This can easily be generalized to any normalizer object, not just linkers, if needed in the future.
- Keep track of which normalizers have already been loaded to avoid multiple instances of identical normalizers (so far only to avoid Language normalizer being loaded by `heuristicLinkers` method).

### DIFF
--- a/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
+++ b/whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
@@ -35,21 +35,20 @@ class Normalizers {
     private static Set<String> loadedTypes = []
 
     static DocumentNormalizer nullRemover() {
-        return { Document doc ->
+        return new Normalizer({ Document doc ->
             traverse(doc.data, { value, path ->
                 if (value == null) {
                     new Remove()
                 }
             })
-        }
+        })
     }
 
     static DocumentNormalizer language(Whelk whelk) {
         LanguageLinker linker = new LanguageLinker()
         loadDefinitions(linker, whelk)
-        loadedTypes.addAll(linker.getTypes())
 
-        return new LinkerNormalizer(linker, { LanguageLinker l, Document doc ->
+        return new Normalizer(linker, { Document doc, LanguageLinker l = linker ->
             l.linkAll(doc.data, 'associatedLanguage')
             l.linkAll(doc.data, 'language')
         })
@@ -58,7 +57,7 @@ class Normalizers {
     static DocumentNormalizer romanizer(Whelk whelk) {
         def langAliases = whelk.jsonld.langContainerAlias.values() as Set
 
-        return { Document doc ->
+        return new Normalizer({ Document doc ->
             traverse(doc.data, { value, path ->
                 if (value instanceof Map && path && path.last() instanceof String && path.last() in langAliases) {
                     def byLang = value
@@ -70,7 +69,7 @@ class Normalizers {
                 }
                 DocumentUtil.NOP
             })
-        }
+        })
     }
 
     /**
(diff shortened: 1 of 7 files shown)


whelk-core/src/test/groovy/whelk/filter/LanguageLinkerSpec.groovy
```diff
@@ -264,16 +280,80 @@ class LanguageLinkerSpec extends Specification {
...
+                        '@id'          : 'http://id/lat',
+                        code           : 'lat',
+                        prefLabelByLang: [
+                                'sv': 'Latin',
+                                'en': 'Latin'
+                        ]
+                ]
+        ].each(linker.&addDefinition)
+        linker.addMapping('grekiska', 'http://id/gre')
+        linker.addMapping('grekiska', 'http://id/grc')
```
olovy, line 341: Put these as `hiddenLabelByLang` in the test data?

whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
```diff
@@ -33,65 +32,67 @@ example:
 
 @Log
 class Normalizers {
-    
+    private static Set<String> loadedTypes = []
```
olovy, line 35: This is shaky because it depends on which order the normalizers below are created

olovy (changes requested):
The implicit dependency between langueLinker and heuristicLinker which is only upheld by the current order the methods are called needs to be fixed.

Maybe something like
```
    private void initDocumentNormalizers() {
        LanguageLinker languageLinker = new LanguageLinker()  // ***
        Normalizers.loadDefinitions(languageLinker, this)  // ***
        normalizer = new NormalizerChain(
                [
                        Normalizers.nullRemover(),
                        //FIXME: This is KBV specific stuff
                        Normalizers.workPosition(jsonld),
                        Normalizers.typeSingularity(jsonld),
                        Normalizers.language(languageLinker),  // ***
                        Normalizers.identifiedBy(),
                        Normalizers.romanizer(this),
                ] + Normalizers.heuristicLinkers(this, languageLinker.getTypes())  // ***
        )

        // Make languageLinker avaliable outside this context ***
    }
```
```
static Collection<DocumentNormalizer> heuristicLinkers(Whelk whelk, Collection<String> skipTypes) {
```

whelk-core/src/test/groovy/whelk/filter/LanguageLinkerSpec.groovy
kwahlin: :+1: See [eafb332](https://github.com/libris/librisxl/pull/1174/commits/eafb332a4f8429ad189c9935ec10220c3fc00cf5)

whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
kwahlin: Makes perfect sense. Fixed in [0bf743a](https://github.com/libris/librisxl/pull/1174/commits/0bf743a192678ffe42afbd676478f666cd86946d)

olovy (approved):
LGTM!
:ship:

PR #1175, opened 2022-11-15
EOD Change controlnumber in otherPhysicalFormat

poacherone:
Fix for faulty controlnumbers in otherPhysicalFormat for some EOD records.


niklasl:
If the controlnumbers are from real records, change the local (blank) `otherPhysicalFormat` to a link instead. (Or `reproductionOf` if that is the case here?)

poacherone:
Changed to adding reproductionOf. @niklasl

whelktool/scripts/syslib/change-otherphysicalformat-controlnumber.groovy
```diff
@@ -0,0 +1,32 @@
+File bibids = new File(scriptDir, "eod_som_ska_rattas_i_libris_2022_11.txt")
+List<String> ProgramLines = bibids.readLines()
+
+for (String operation : ProgramLines) {
+    String[] part = operation.split('\t')
+    String ctrlnr = part[0].trim()
+    String oldCtrlnr = part[1].trim()
+    String newCtrlnr = part[2].trim()
+
+    where = """id in
```
niklasl, line 10: Needs to be local (to avoid being overwritten by another thread):
```suggestion
    def where = """id in
```

whelktool/scripts/syslib/change-otherphysicalformat-controlnumber.groovy
```diff
@@ -0,0 +1,32 @@
...
+    where lb.collection = 'bib' and
+    lb.data#>>'{@graph,0,controlNumber}' = '${ctrlnr}'
+)"""
+
+    selectBySqlWhere(where, loud: true, { bib ->
+        def oPF = bib.graph[1]["otherPhysicalFormat"][0]["describedBy"][0]["controlNumber"]
+        boolean modified
+
+        if (oPF == oldCtrlnr) {
+            bib.graph[1].remove('otherPhysicalFormat')
```
niklasl, line 22: Might not be the case here, but in general, check if there is more than one item in `otherPhysicalFormat` before dropping (unless any others are to be discarded).

whelktool/scripts/syslib/change-otherphysicalformat-controlnumber.groovy
```diff
@@ -0,0 +1,32 @@
...
+            bib.graph[1]["reproductionOf"] =  []
+            bib.graph[1]["reproductionOf"] <<
+                    ["@id": "https://libris-qa.kb.se/$newCtrlnr#it"]
```
niklasl, line 25: Tied to the QA environment, won't work in Prod. Use `findCanonicalId(newCtrlnr)` to get the main id from the record id.

niklasl (changes requested):
Some small fixes (and another test run, perhaps in STG) should be all to make this work!

poacherone:
Added new script to fix reproductionOf (on the correct records!).

niklasl (approved):
I believe my concerns have been addressed. (Alas, I don't have much time to verify, so do make another test run or ask for a second glance in case of doubt though.)

*NOTE* I do think making the checks before dropping `otherPhysicalFormat` are prudent (e.g. using `assert` which will fail if the assumptions don't hold).

(And once run in QA or STG, check that the links work and that, from these, the expected bib 773 MARC21 is present.)

poacherone:
I have made a test run on QA and will do another one on STG to make sure all the links work correct. 776 (as is the case here) is present in the conversion to MARC21.

PR #1178, opened 2022-11-25
Fix a regression that prevented CLI exports from terminating.

jannistsiroyannis:
A Java process does not terminate until all (normal) threads have finished executing. The ExecutorService, by (bad) design leaves it's threads alive and spinning, which means that process termination cannot occurr unless the ExecutorService is shutdown().


marc_export/src/main/java/whelk/export/marc/ProfileExport.java
```diff
@@ -495,6 +495,7 @@ public void submit(ResultSet resultSet) throws SQLException, IOException {
         public void awaitCompletion() throws IOException {
             workQueue.awaitAll();
             out.close();
+            pool.shutdown();
```
niklasl, line 498:
```suggestion
            try {
                out.close();
            } finally {
                pool.shutdown();
            }
```
(Maybe include `workQueue.awaitAll()` in the `try`, unless that failing would  make any futher recovery futile?)

jannistsiroyannis: :+1: did it as a separate commit.

olovy (changes requested):
This breaks the HTTP export since a new pool is created for every request.
If we make `ProfileExport` create and own the pool instead. And pass it to the constructor of `ParallelExporter`.

olovy (approved):
LGTM!

marc_export/src/main/java/whelk/export/marc/ProfileExport.java
poxelman: Reply

PR #1179, opened 2022-11-28
Add test for /_changesets; 'changesets' => '_changesets' in changeset @id

andersju:
Hardly PR-worthy but note that in the changeset `@id`, `changesets` is now `_changesets` to reflect the actual URL we use (previously the `@id` would 404). I assume `_changesets` is what we want. (i.e. https://libris-qa.kb.se/9slgbxnm1x552vm/changesets => https://libris-qa.kb.se/9slgbxnm1x552vm/_changesets)

I realized that we did already have some History tests, so I just re-enabled the one that was ignored (#1172 made it work again). Plus added a basic test in CrudSpec.

### DIFF
--- a/rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
+++ b/rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
@@ -1,6 +1,6 @@
 package whelk.rest.api
 
-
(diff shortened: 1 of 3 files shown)


jannistsiroyannis (approved):
:ship:

PR #1180, opened 2022-11-28
Fix MARC21 bib 008 generation for digital reproductions

niklasl:
See commits for details. This should(...) address LXL-3942.


kwahlin (approved):
LGTM!

klngwll (approved):
Especially for Manuscripts there seems to be no particular coherence in production/publication/marc:primaryProvisionActivity and it's corresponding "Primary" class, just to note there are still cases we are not covering when only injecting from publication (Not suggesting to fix this within this issue, only for awareness of sprawling dataforms that will not work in 008).

PR #1181, opened 2022-11-29
Add IgnoreOnRevert to STRN in linkfields with spec

klngwll:
Add ignoreOnRevert for STRN identifiers in linkfield pattern ($u). This will mute occurences in any linked publication.

LXL-3964


niklasl (approved):
LGTM! :ship:

PR #1182, opened 2022-12-01
Allow URI:s to be used as "collections"

niklasl:
This allows URI:s to be used as the `collection` argument to `storage.loadData`, in which case we search for records linking to the URI using `inDataset` instead of the lddb collection column (one step closer to deprecation).

This makes it possible to e.g. reindex based on dataset membership.


jannistsiroyannis (approved):
:ship:

olovy (approved):
:ship:

PR #1187, opened 2022-12-13
Fix RDF/XML serialization

niklasl:
Broke when trying to frame certain payloads. This was unnecessary anyway, so removed that step and simplified the code a bit.

Example:
```sh
$ curl -s -L -HAccept:application/rdf+xml "http://libris.kblocalhost.kb.se/find?q=*"
```

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/JsonLD2RdfXml.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/JsonLD2RdfXml.groovy
@@ -1,45 +1,45 @@
 package whelk.converter
 
+import groovy.transform.CompileStatic
+
 import org.apache.commons.io.IOUtils
-import org.apache.jena.rdf.model.Model
 import org.apache.jena.rdf.model.ModelFactory
-import org.apache.jena.rdf.model.RDFWriter
 import whelk.Document
 import whelk.JsonLd
 import whelk.Whelk
-import whelk.component.PostgreSQLComponent
-import whelk.util.PropertyLoader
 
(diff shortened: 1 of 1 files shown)


whelk-core/src/main/groovy/whelk/converter/JsonLD2RdfXml.groovy
```diff
@@ -1,45 +1,45 @@
...
+        srcData.putAll(originaldata)
+        if (context) {
+            srcData[JsonLd.CONTEXT_KEY] = context
+        }
 
-        InputStream input = IOUtils.toInputStream(framedString)
-        Model model = ModelFactory.createDefaultModel()
-        ByteArrayOutputStream baos = new ByteArrayOutputStream()
+        var jsonldStr = mapper.writeValueAsString(srcData)
+        var input = IOUtils.toInputStream(jsonldStr)
```
olovy, line 32:
```suggestion
        var input = IOUtils.toInputStream(jsonldStr, StandardCharsets.UTF_8)
```

This seems to be the code that's used on the Jena side:
https://github.com/jsonld-java/jsonld-java/blob/da156eebe0fdd2cee34a8bb603d64979a281d389/core/src/main/java/com/github/jsonldjava/utils/JsonUtils.java#L68

Newer version of same code:
https://github.com/jsonld-java/jsonld-java/blob/f2f1cf22dc8a9c30c47ab570c7aaa8243101ad02/core/src/main/java/com/github/jsonldjava/utils/JsonUtils.java#L102

olovy (approved):
See suggestion.

LGTM

PR #1191, opened 2023-01-05
Search: Make filtering on multi-language properties work 

olovy:
### Make filtering on multi-language properties work 

Make filtering on for example `prefLabel=foo` also match `prefLabelByLang.sv`, `prefLabelByLang.en` etc.

Filtering on `prefLabelByLang.sv` explicitly still works.

Add an extra field to the ES index that contains both the language tagged and plain strings for a property.

For example:
```
{
  "foo": "FOO",
  "fooByLang": { "en": "EN", "sv": "SV" }
}
```
is indexed as
```
{
  "foo": "FOO",
  "fooByLang": { "en": "EN", "sv": "SV" },
  "__foo": ["FOO", "EN", "SV"]
}
```

Queries are adjusted accordingly to filter on this field instead.
Internal fields starting with '_' are already removed from the results.

This makes searching on mainTitle in the cataloging client work for instances with romanized + original script titles.

----

### Keep transformed + original strings in embellishments 
When indexing, for things added by embellish we remove language tagged string that are not in the whelk locale in order to keep
the index size down. With this change we also keep transformed strings and their original counterpart (i.e. for now romanized + original script). For example title of Instance linked from Item.

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -239,7 +239,7 @@
                 "include_in_parent": true,
                 "properties": {
                     "@type": {"type": "keyword"},
-                    "mainTitle": {
+                    "__mainTitle": {
                         "type": "text",
                         "analyzer": "softmatcher",
                         "copy_to": "_all",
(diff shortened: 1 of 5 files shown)


lrosenstrom (approved):
LGTM. Seems to work as intended when testing on my local setup.

PR #1192, opened 2023-01-09
Search: Boost Records more than CacheRecords

olovy:
Per the suggestion in LXL-3919.
So that for instance sao is ranked higher than external term collections like Queerlit.

It is not possible to specify negative boost values so just boost the base `Record` type instead.

1000 seems like a number that works...

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
@@ -1,13 +1,12 @@
 package whelk.search
 
-import groovy.transform.CompileDynamic
+
 import groovy.transform.CompileStatic
 import groovy.transform.PackageScope
 import groovy.transform.TypeCheckingMode
 import groovy.util.logging.Log4j2 as Log
 import whelk.JsonLd
 import whelk.Whelk
-import whelk.component.ElasticSearch
 import whelk.exception.InvalidQueryException
 import whelk.util.DocumentUtil
 import whelk.util.Unicode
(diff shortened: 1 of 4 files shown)


whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
```diff
@@ -30,6 +30,13 @@ class ESQuery {
     private static final String OR_PREFIX = 'or-'
     private static final String EXISTS_PREFIX = 'exists-'
 
+    private static final Map recordsOverCacheRecordsBoost = [
+            'constant_score': [
+                    'filter': [ 'term': [ (JsonLd.RECORD_KEY + '.' + JsonLd.TYPE_KEY) : JsonLd.BASE_RECORD_TYPE ]],
+                    'boost': 1000.0
```
niklasl, line 36: I'm assuming this doesn't affect scoring in other ways (e.g. reaching a maximum making other scores moot)? (Otherwise we could perhaps "punish" use of `CacheRecord`?)

niklasl (approved):
Nice! LGTM! (See question comment.)

(As follow-up we could do a sweep for other hardcoded uses of `Record`.)

olovy:
> (As follow-up we could do a sweep for other hardcoded uses of `Record`.)

Added on use that I had missed.

whelk-core/src/main/groovy/whelk/search/ESQuery.groovy
olovy: It is not possible to specify negative boost values.

PR #1193, opened 2023-01-16
Fix dependers handling in DatasetImporter

olovy:
Summary
-------
In DatasetImporter
* Make `--skip-dependers` flag only affect indexing
* Handle circular dependencies among deleted docs in dataset

Changes
-------
Remove the option to not update `lddb__dependencies` when loading a
dataset (`--skip-dependers`) since using it puts the database in an
inconsistent state. Updating `lddb_dependencies` is cheap because it is
only the _outgoing_ links for a modified doc that are updated.

`--skip-dependers` now only affects reindexing. Reindexing dependers
might be expensive when reloading core datasets. The flag can be
removed if we improve the logic for calculating which documents
need to be reindex for an update.

Handle circular dependencies among deleted docs in dataset. If no other
docs depend on the they can be deleted safely.

### DIFF
--- a/apix_server/src/main/java/whelk/apixserver/ApixCatServlet.java
+++ b/apix_server/src/main/java/whelk/apixserver/ApixCatServlet.java
@@ -5,7 +5,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import whelk.Document;
-import whelk.util.LegacyIntegrationTools;
 
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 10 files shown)


niklasl (approved):
LGTM! :1st_place_medal: for less flags and other code clarifications!

niklasl:
Note: We need to conditionally adapt the devops scripts (`whelk.py` and `bibdb-import.sh`) so they use this new flag when they use a "develop" version of new `xlimporter.jar`.

olovy:
> Note: We need to conditionally adapt the devops scripts (`whelk.py` and `bibdb-import.sh`) so they use this new flag when they use a "develop" version of new `xlimporter.jar`.

Maybe we just keep the old name for the flag.

olovy:
Keep `skip-dependers` instead of a new `skip-index-dependers`

PR #1195, opened 2023-01-18
Prevent additional bindings during record processing

niklasl:
Complain and stop if a whelktool groovy script adds new global bindings during its processing. This is done by monitoring existing bindings for each processBatch call, and reacting if anything was added.

This has been a long-standing risk, since a script might accidentally bind a *global* variable deep down in the code (by forgetting to declare it using `def`, `var` or a type). This will then be overwritten by another thread running the same parts of the code, causing all kinds of unexpected behaviour (commonly data from one thread being used or written by another).

There are other ways of handling this (the most proper being to create fully separate bindings for each thread), but this was the simplest tweak I could think of.


olovy (approved):
Nice!
:ship:

PR #1196, opened 2023-01-24
Move Li donation notes to custodialHistory

poacherone:
Move donation data from notes (in various fields/entities) to custodialHistory, removing empty objects. If hasComponent or marc:hasTextualHoldingsBasicBibliographicUnit only has certain values left (type, heldBy or marc:groupid in hasComponent for example, i.e. are left "empty") they are removed as well.


kwahlin (approved):
Looks correct!

PR #1197, opened 2023-01-25
Add `and-` and `not-` prefixes to /find in REST API

andersju:
Seems to work (but needs more testing!).


niklasl (approved):
LGTM! Let's get it to QA and try it out thoroughly!

(IIRC we had some integration tests in the making recently for testing search behaviour more knowingly against a known dataset. That's a good place to start.)

PR #1198, opened 2023-01-26
Search: Allow multiple facets on the same property

olovy:
Now that we have `and-` we can do it.

TODO: Client needs to be fixed to check `predicate` and not `variable` when rendering facet pills.
TODO: Generated `predicate` should reflect `not-` somehow?

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
@@ -165,18 +165,18 @@ class SearchUtils {
                 return item
             }
         }
-
-        // Don't return facets on fields that already have an active facet.
-        // Now multiple conditions on the same field are always ORed, which means facets will behave weird.
-        // This can be changed if we add a way to have x=a AND x=b in the API. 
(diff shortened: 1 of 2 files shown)


andersju (approved):
LGTM!

"Generated predicate should reflect not-" -- not sure what is meant or needed here (and I haven't really looked at how this stuff is generated) but the predicate's `label` and `_key` will be prefixed with `and-` or `not-`, right? And the first id in propertyChainAxiom would be e.g. https://id.kb.se/vocab/not-instanceOf (just like the predicate for `exists-` would be e.g. https://id.kb.se/vocab/exists-mediaType).

olovy:
> "Generated predicate should reflect not-" -- not sure what is meant or needed here (and I haven't really looked at how this stuff is generated) but the predicate's `label` and `_key` will be prefixed with `and-` or `not-`, right? And the first id in propertyChainAxiom would be e.g. https://id.kb.se/vocab/not-instanceOf (just like the predicate for `exists-` would be e.g. https://id.kb.se/vocab/exists-mediaType).

We need to revisit and document the whole `search.mapping` construct.

PR #1199, opened 2023-01-26
Add availability to Ylm holdings

poacherone:
All holdings for sigel Ylm are not for ILL, this script adds availability to all items (that doesn't already have this).

### DIFF
--- a/whelktool/scripts/syslib/Ylm-add-availability.groovy
+++ b/whelktool/scripts/syslib/Ylm-add-availability.groovy
@@ -1,0 +1,76 @@
+String where = """id in (
+   SELECT lh.id FROM lddb lh WHERE
+   lh.data #>>'{@graph,1,heldBy,@id}' = 'https://libris.kb.se/library/Ylm'
+   AND lh.deleted = 'false')
+"""
+
+selectBySqlWhere(where, silent: false, { hold ->
+
+    def holds = hold.graph[1]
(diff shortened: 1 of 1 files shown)


whelktool/scripts/syslib/Ylm-add-availability.groovy
```diff
@@ -0,0 +1,76 @@
...
+        holds["hasComponent"].each { it ->
+            if (it["availability"]) {
+                asList(it["availability"]).each { availability ->
+
+                    if (availability["label"] && availability["label"] instanceof String) {
+                        if (!availability["label"].contains("Ej fjärrlån")) {
+                            shouldChange = true
+                        }
+                    }
+                }
```
olovy, line 24: This should be reversed?
Now `shouldChange` is true if there is _any_ availability that is _not_ "Ej fjärrlån"
It should be false if there is _any_ availability that _is_ "Ej fjärrrlån"

olovy (commented):
The check for existing availability needs to look for `"@id": "https://id.kb.se/term/enum/NotForILL"? 
Now it only handles local `ItemAvailability`. I'm not sure if there any of these.

poacherone:
There is no linked availability in these records, only a few local ones.

olovy (approved):
See comment about backwards condition
:ship:

poacherone:
Fixed condition.

PR #1200, opened 2023-01-27
Add isPrimaryTopicOf for LiU

poacherone:
Adding isPrimaryTopicOf (uri and public note) for DiVA records from LiU. No records currently has isPrimaryTopicOf, checked separately.


olovy (approved):
LGTM!

PR #1201, opened 2023-01-31
whelktool: Add --num-threads parameter

olovy:
When running whelktool on a machines with many cores (xlbuild) against an underpowered ES (dev2) the HttpClient connection pool starts timing out: `whelk.exception.ElasticIOException: Timeout waiting for connection from pool`.

A better solution would be to use a different timeout for whelktool.
An even better solution would be to do bulk indexing for whelktool jobs,


kwahlin (approved):
LGTM

PR #1205, opened 2023-02-06
OAI-PMH: Don't export bibs/holds on insignificant auth changes

olovy:
When requesting bib or hold sets, minor changes in linked auth records should not trigger exports. If the changes are outside the auth card they cannot affect the contents of the converted bib/hold record.

### DIFF
--- a/oaipmh/src/main/java/whelk/export/servlet/Helpers.java
+++ b/oaipmh/src/main/java/whelk/export/servlet/Helpers.java
@@ -103,7 +103,7 @@ public class Helpers
                 if (requestedCollection.equals("auth") &&
                         (type == null || !OaiPmh.workDerivativeTypes.contains(type)))
                     queueDocument(updated);
-                if (includeDependenciesInTimeInterval && (requestedCollection.equals("bib") || requestedCollection.equals("hold")))
+                if (includeDependenciesInTimeInterval && (requestedCollection.equals("bib") || requestedCollection.equals("hold")) && hasCardChanged(updated))
                 {
                     List<Tuple2<String, String>> dependers = OaiPmh.s_whelk.getStorage().followDependers(updated.getShortId(), JsonLd.getNON_DEPENDANT_RELATIONS());
                     for (Tuple2<String, String> depender : dependers)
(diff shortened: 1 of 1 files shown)


jannistsiroyannis:
Great!

Probably, the same logic should also be in effect somewhere around here:

https://github.com/libris/librisxl/blob/d64f221b349b6e6db801de3477ec49ce30c9c6b2/marc_export/src/main/java/whelk/export/marc/ProfileExport.java#L241

oaipmh/src/main/java/whelk/export/servlet/Helpers.java
```diff
@@ -151,6 +151,18 @@ else if (updatedCollection.equals("hold"))
...
+        private boolean hasCardChanged(Document updated) {
+            Document previousVersion = OaiPmh.s_whelk.getStorage().load(updated.getShortId(), "-1");
+            if (previousVersion == null) {
+                return true;
+            }
+            
+            var jsonLd = OaiPmh.s_whelk.getJsonld();
+            var oldCard = jsonLd.toCard(previousVersion.data);
+            var newCard = jsonLd.toCard(updated.data);
+            return !oldCard.equals(newCard);
```
niklasl, line 163: This _ought_ to be stable enough (and is certainly simple enough). It hinges on whether the `equals` of the List and Map implementations are "deep equals", which is should be the case for [AbstractList](https://docs.oracle.com/javase/1.5.0/docs/api/java/util/AbstractList.html#equals(java.lang.Object)) and Map (entry sets) see [AbstractSet](https://docs.oracle.com/javase/1.5.0/docs/api/java/util/AbstractSet.html#equals(java.lang.Object)) (since at least 1.5.0). It effectively depends on implementation though; so even if wee're good now, _if_  the JSON parser would change, we would need some test cases to ensure this contract holds?

niklasl (approved):
LGTM! (See comment about spec:ing expectations.)

oaipmh/src/main/java/whelk/export/servlet/Helpers.java
olovy: It's part of the contract of the List and Map interfaces. So unless we change to something that maps JSON arrays and objects to something different I thing we are fine.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#equals(java.lang.Object)
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html#equals(java.lang.Object)

olovy: Switching to our `DocumentComparator.isEqual` would treat arrays/lists as unordered where they are.

PR #1210, opened 2023-02-09
Fixes for Azerbaijani

olovy:
- Add characters missing in LC's table (see https://github.com/lcnetdev/scriptshifter/pull/4)
- Add case handling (not supported by ScriptShifter)


lrosenstrom (approved):
LGTM!

PR #1213, opened 2023-02-13
Fixes for Armenian

olovy:
- Add missing chars (see https://github.com/lcnetdev/scriptshifter/pull/5)
  - Պ P
  - պ p
  - Ձ Dz
  - ձ dz
- Handle "Note 6"
- Handle lower/upper case
- Handle punctuation
- Generate composed (NFC) Latin chars
- Add spec


lrosenstrom (approved):
Great! Let's get this out for testing :ship:

PR #1214, opened 2023-02-17
marcexport: Don't export bibs on insignificant auth changes

olovy:
Minor changes in linked auth records should not trigger exports. If the changes are outside the auth card they cannot affect the contents of the converted bib record.


jannistsiroyannis (approved):
:ship:

niklasl (approved):
:shipit:

PR #1215, opened 2023-02-17
Fixes for Amharic

olovy:
* Handle sixth order characters where possible
* Add specs


lrosenstrom (approved):
LGTM! Instead of transforming በታች to batāče it now becomes batāč, which seems to follow the rules mentioned at https://learnamharic.wixsite.com/learnamharic/amharic-script-1.

lrosenstrom:
LGTM! Instead of transforming በታች to batāče it now becomes batāč, which seems to follow the rules mentioned at https://learnamharic.wixsite.com/learnamharic/amharic-script-1.

PR #1216, opened 2023-02-21
Add binding to field 340 conversion

klngwll:
Add binding to 340$l as per LC convspec 2.3 and MARC -> https://www.loc.gov/marc/bibliographic/bd340.html

- [x] vocab change https://github.com/libris/definitions/pull/416 (not required).


niklasl (approved):
LGTM

PR #1217, opened 2023-02-23
Add support for search-chips

olovy:
Search-chips work in the same way as search-cards. They are indexed but not displayed. Embellish documents for indexing with search-chips instead of regular chips. Note that a search-chip has to be a subset of the card for the same type. Because they are made by shrinking cards.


niklasl (approved):
LGTM!

PR #1219, opened 2023-03-01
Strip trailing line breaks in strings on doc save

olovy:
Reported as a real world problem: trailing newline in uri property is treated as a part of the uri in presentation.


niklasl (approved):
LGTM!

PR #1220, opened 2023-03-01
Add support for multiple entries with identical language tags

lrosenstrom:


whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy
```diff
@@ -185,7 +185,7 @@ class JsonLdValidator {
                     handleError(new Error(Error.Type.OBJECT_TYPE_EXPECTED, key, value), validation)
                 }
             }
-            else if (value !instanceof String) {
+            else if (value !instanceof String && value !instanceof List) {
```
olovy, line 188: Should check that all members in the list are strings?

PR #1221, opened 2023-03-02
Add script for linking CC licenses given in `uri`

niklasl:
This should be combined with informing the sources of that data that they can now properly link up these licenses using `@id` only.

We may want to check incoming uri values in general, either in the cataloguing interface and/or on save by extending the LinkFinder-based normalization step with such logic.


jannistsiroyannis:
This (as it turns out) has been implicitly achieved by another script already.

PR #1222, opened 2023-03-03
Add and-matches- prefix to /find in REST API

andersju:
Continuation of #1197.

Might be a nicer way to do this, I just can't think of any at the moment.

Also added an example of a use of `matches-` that was previously undocumented (intentionally? :thinking:).


olovy (approved):
:ship:

PR #1225, opened 2023-03-13
Handle range queries properly again

andersju:
After #1222 invalid range queries (e.g. https://libris-qa.kb.se/find.jsonld?max-meta.created=2001-W05&min-meta.created=2000) would erroneously return 200 OK rather than an error. This should fix it.


niklasl (approved):
LGTM! (It's a bit hard to follow without a unit test, but I assume this fixed the failing integration test?)

PR #1227, opened 2023-03-17
Search: Fix sorting on lang taggable property, e.g. hasTitle.mainTitle

olovy:
which are now indexed as e.g. `hasTitle.__mainTitle`


lrosenstrom (approved):
:ship:

PR #1228, opened 2023-03-17
Add seriesMembership path

klngwll:
Add seriesMembership path to marc:version > version script

### DIFF
--- a/whelktool/scripts/2023/02/lxl-1556-version-vs-marc-version/2-marc-version-to-version.groovy
+++ b/whelktool/scripts/2023/02/lxl-1556-version-vs-marc-version/2-marc-version-to-version.groovy
@@ -20,7 +20,7 @@ MARC_VERSION_BIB_PATHS =
                 ['instanceOf', 'hasPart'],
                 ['instanceOf', 'relationship', '*', 'entity'],
                 ['instanceOf', 'subject'],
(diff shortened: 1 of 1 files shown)


whelktool/scripts/2023/02/lxl-1556-version-vs-marc-version/2-marc-version-to-version.groovy
```diff
@@ -19,7 +19,8 @@ MARC_VERSION_BIB_PATHS =
                 ['instanceOf', 'expressionOf'],
                 ['instanceOf', 'hasPart'],
                 ['instanceOf', 'relationship', '*', 'entity'],
-                ['instanceOf', 'subject']
+                ['instanceOf', 'subject'],
+                ['instanceOf', 'seriesMembership', 'inSeries', 'instanceOf']
```
kwahlin, line 23:
```suggestion
                ['seriesMembership', '*', 'inSeries', 'instanceOf']
```

kwahlin: `seriesMembership` is on the instance right? Since `seriesMembership` is an array we also need the asterisk to get all objects from that array.

kwahlin (changes requested):

kwahlin (approved):
LGTM!

PR #1229, opened 2023-03-24
Attempt to run whelk-dependent integration tests

niklasl:
Some MarcFramePostProcStep:s are resource-dependent and need a backing whelk to load those. This change adds a flag on those steps, which is used to let MarcFrameConverterSpec conditionally run the specs for these if it is able to instantiate a whelk upon startup (commonly when integTest is run).

(This is conditional to make it possible to run the non-resource-dependent steps locally without having a whelk up.)

Remaining:
* [ ] Make failing Romanization specs pass

To consider:
* [ ] Can we untangle the dependency on `RomanizationStep.LanguageResources` from `MarcframeConverter`  (general > specific)  by abstracting something? (Like a "resource provider" from the whole `Whelk` API... :thinking: )

When merged:
* [ ] Add `local(f"../gradlew -Dxl.secret.properties={secrets.create_whelk_secret_properties()} integTest")`  to `app.whelk.run_core_integration_tests` in devops repo.

### DIFF
--- a/whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
+++ b/whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
@@ -7,9 +7,17 @@ import spock.lang.*
 @Unroll
 class MarcFrameConverterSpec extends Specification {
 
-    static converter = new MarcFrameConverter() {
-        def config
+    static Whelk whelk = null
+    static {
+        try {
+            whelk = Whelk.createLoadedSearchWhelk()
+        } catch (Exception e) {
+            System.err.println("Unable to instantiate whelk: $e")
+        }
+    }
 
+    static converter = new MarcFrameConverter(null, whelk?.jsonld, whelk?.languageResources) {
+        def config
         void initialize(Map config) {
             super.initialize(config)
             this.config = config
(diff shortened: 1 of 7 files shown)


kwahlin (approved):
Nice! Looks good.

I think we need to rework `RomanizationStep` in general, along with proper test cases.

whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
```diff
@@ -7,9 +7,17 @@ import spock.lang.*
...
 class MarcFrameConverterSpec extends Specification {
 
-    static converter = new MarcFrameConverter() {
-        def config
+    static Whelk whelk = null
+    static {
+        try {
+            whelk = Whelk.createLoadedSearchWhelk()
+        } catch (Exception e) {
+            System.err.println("Unable to instantiate whelk: $e")
```
kwahlin, line 15: Do we also need to warn that some tests may not be carried out?

olovy (approved):
LGTM!

I'll try to remove the resource dependencies from RomanizationStep

whelk-core/src/integTest/groovy/MarcFrameConverterSpec.groovy
niklasl: Good suggestion; addressed in 97b471a618bfd25099cdc6235659223d21484966.

PR #1230, opened 2023-03-24
Feature/remove nationality from topics

klngwll:
Remove nationality property from Topics


kwahlin (approved):
LGTM :united_nations:

PR #1233, opened 2023-03-29
Feature/fix linkage in 880 revert

kwahlin:
This addresses a few known issues with the 880 revert conversion.
- Missing fieldref in 240 and mismatched fieldrefs for repeated fields. Fieldrefs are now placed out in a reliable way, see 3619cc3.
- Missing 130 for romanized work titles (after "titeljonglering"). `RomanizationStep` and `NormalizeWorkTitlesStep` should no longer interfere with each other, see 32d4a7d.

Added a few test cases for some of the (before) problematic cases that I'm aware of. Feel free to add more.

- [x] integTest

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
@@ -166,7 +166,7 @@ class RomanizationStep extends MarcFramePostProcStepBase {
 
         def fieldRefs = []
 
-        DocumentUtil.findKey(thingCopy, FIELDREF) { value, path ->
+        DocumentUtil.findKey(thingCopy, FIELD_REFS) { value, path ->
             def romanizedField = romanizedFieldsByTmpRef[value]
             if (romanizedField) {
                 def fieldNumber = romanizedField.keySet()[0]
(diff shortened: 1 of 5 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
```diff
@@ -149,7 +145,7 @@ class RomanizationStep extends MarcFramePostProcStepBase {
         def uniqueTLangs = findUniqueTLangs(thing, byLangPaths)
 
         // TODO: Can multiple transform rules be applied to the same record? What happens then with non-repeatable fields?
-                
+
         uniqueTLangs.each { tLang ->
             unmodifyTLang(thing, tLang, byLangPaths, record)
         }
```
olovy, line 151: I suggest to sort the 880s on their corresponding field number so that they appear in a logical order in the final record. Even though the fields-refs may still be in another order. 
```suggestion
        }
        
        if (thing[HAS_BIB880]) {
            ((List) thing[HAS_BIB880]).sort { bib880 -> bib880[PART_LIST]?[0]?[FIELDREF] ?: '' }
        }
```

olovy (approved):
LGTM!

See suggestion about sorting the 880s.

PR #1234, opened 2023-04-03
Add source for uniform titles

kwahlin:


niklasl (commented):
We define source with `@container` `@set` in our JSON-LD context, so it is expected to be an array. Other than that, looks good!

niklasl (approved):
LGTM! :ship:

PR #1235, opened 2023-04-03
History: More precise diffs

olovy:
* Fix an issue when modifying objects inside arrays. Adding a property to an object inside an array would report the whole object as removed in addition to the property being added. And vice versa when removing properties.
* Return more helpful diffs when objects are completely replaced. Before each property inside the object would be listed as removed/added. Now only the object is removed + added.
* Always include `addedPaths` and `removedPaths` in the response even if they are empty.
* Add more tests


lrosenstrom (approved):
Nice! :ship:

jannistsiroyannis (approved):
:ship:

PR #1241, opened 2023-04-17
Feature/translation of into has part on convert

kwahlin:
We (me + Elisabet and Katja) came to the conclusion that it's better to move some than nothing at all (less manual work). 
On import this will move translationOf into each part in hasPart given that:
- There is exactly one instanceOf language
- There is exactly one instanceOf.translationOf language
- Every part has a title, no translationOf and exactly one language which the same as the instanceOf language.

Should be safe to do this!


olovy (approved):
:ship:

PR #1243, opened 2023-04-19
Always reassign bib880s to marc:hasBib880

kwahlin:
Realized that if _none_ of the `bib880`'s could be handled they would be lost.


niklasl (approved):
LGTM! :sweat_smile: :shipit:

PR #1244, opened 2023-04-20
Handle translationOf containing more than @type and language

kwahlin:
See for example https://libris.kb.se/katalogisering/1kcgv3jc0m8kc75.
The translationOf title should not be moved into hasPart.


olovy (commented):
:ship:

PR #1246, opened 2023-04-20
On revert always move original title to instanceOf, regardless of languages

kwahlin:
I can't see why instanceOf.language or translationOf.language would matter here. If the languages are not correctly described then that's a data quality problem, and we should still produce 240/130 $a, even if $l is incorrect or missing.


olovy (approved):
LGTM!

PR #1247, opened 2023-04-21
Fix MARC post processing steps randomly not running

olovy:
There is one shared MarcFrameConverter instance in Whelk. MarcFrameConverter has a single MarcConversion instance reused for all conversions. Turning off the doPostProcessing flag to prevent recursive calls causes post processing to not be run at random in multi-threaded contexts e.g. the MARC export API.

(Could of course also be fixed with a parameter to runConvert/runRevert instead)


andersju (approved):
:ship:

PR #1248, opened 2023-04-24
Move only translated titles recursively

kwahlin:
Fixes LXL-4106. The problem was that `expressionOf` was constructed recursively for any work associated with the `instanceOf` property. 830 is not mapped to `expressionOf` in marcframe so the conversion fails when the title ends up in `seriesMembership.inSeries.instanceOf.expressionOf`. Now only the move from/to `translationOf` is done recursively.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/NormalizeWorkTitlesStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/NormalizeWorkTitlesStep.groovy
@@ -24,10 +24,6 @@ class NormalizeWorkTitlesStep extends MarcFramePostProcStepBase {
     LanguageLinker langLinker
 
     void modify(Map record, Map thing) {
-        _modify(record, thing)
-    }
-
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/NormalizeWorkTitlesStep.groovy
```diff
@@ -22,37 +24,39 @@ class NormalizeWorkTitlesStep extends MarcFramePostProcStepBase {
     LanguageLinker langLinker
 
     void modify(Map record, Map thing) {
-        traverse(thing, doModify)
+        _modify(record, thing)
     }
 
-    void unmodify(Map record, Map thing) {
-        traverse(thing, doUnmodify)
+    void _modify(Map record, Map thing) {
```
niklasl, line 30: Maybe merge these two (and similarly for `unmodify`?

niklasl (approved):
LGTM!  (See suggestions for possible simplification.)

PR #1249, opened 2023-04-25
Feature/safer bib880 conversion

kwahlin:
By resetting `thing` we don't risk losing any data should an exception occur between where we remove `marc:hasBib880` (line 95) and where we put it back in place (line 168). This _may_ solve https://jira.kb.se/browse/LXL-4112 (needs more investigation). For the better regardless.

Should also fix the following (from whelk log):
```
ERROR whelk.converter.marc.RomanizationStep - Failed to convert 880: groovy.lang.MissingMethodException: No signature of method: java.util.ArrayList.
keySet() is applicable for argument types: () values: []
Possible solutions: toSet(), toSet(), set(int, java.lang.Object), set(int, java.lang.Object), get(int), get(int)
groovy.lang.MissingMethodException: No signature of method: java.util.ArrayList.keySet() is applicable for argument types: () values: []
Possible solutions: toSet(), toSet(), set(int, java.lang.Object), set(int, java.lang.Object), get(int), get(int)
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:70) ~[groovy-3.0.9.jar:3.0.9]
        at org.codehaus.groovy.runtime.callsite.PojoMetaClassSite.call(PojoMetaClassSite.java:46) ~[groovy-3.0.9.jar:3.0.9]
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47) ~[groovy-3.0.9.jar:3.0.9]
        at java_util_Map$keySet$1.call(Unknown Source) ~[?:?]
        at whelk.converter.marc.RomanizationStep$_findByLangPaths_closure26.doCall(RomanizationStep.groovy:528) ~[whelk-core.jar:?
```

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
@@ -525,11 +525,11 @@ class RomanizationStep extends MarcFramePostProcStepBase {
         List paths = []
 
         DocumentUtil.findKey(thing, byLangToBase.keySet()) { value, path ->
-            if (value instanceof Map && value.keySet().any { it =~ '-t-' }) {
-                paths.add(path.collect())
-            }
+            paths.add(path.collect())
             return
         }
+
+        paths = paths.findAll { DocumentUtil.getAtPath(thing, it).keySet().any { it =~ '-t-' } }
 
         return paths
     }
--- a/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
@@ -62,9 +62,12 @@ class RomanizationStep extends MarcFramePostProcStepBase {
     List FIELD_REFS = [FIELDREF, BIB250_REF]
 
     void modify(Map record, Map thing) {
+        def thingCopy = deepCopy(thing)
         try {
             _modify(record, thing)
         } catch (Exception e) {
+            thing.clear()
+            thing.putAll(thingCopy)
             log.error("Failed to convert 880: $e", e)
         }
     }
--- a/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
@@ -62,12 +62,13 @@ class RomanizationStep extends MarcFramePostProcStepBase {
     List FIELD_REFS = [FIELDREF, BIB250_REF]
 
     void modify(Map record, Map thing) {
-        def thingCopy = deepCopy(thing)
+        def bib880 = thing[HAS_BIB880]
         try {
             _modify(record, thing)
         } catch (Exception e) {
-            thing.clear()
-            thing.putAll(thingCopy)
+            if (bib880) {
+                thing[HAS_BIB880] = bib880
+            }
             log.error("Failed to convert 880: $e", e)
         }
     }
--- a/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
@@ -524,11 +524,11 @@ class RomanizationStep extends MarcFramePostProcStepBase {
         List paths = []
 
         DocumentUtil.findKey(thing, byLangToBase.keySet()) { value, path ->
-            paths.add(path.collect())
+            if (value instanceof Map && value.keySet().any { it =~ '-t-' }) {
+                paths.add(path.collect())
+            }
             return
         }
-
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
```diff
@@ -62,9 +62,13 @@ class RomanizationStep extends MarcFramePostProcStepBase {
...
 
     void modify(Map record, Map thing) {
+        def bib880 = thing[HAS_BIB880]
         try {
             _modify(record, thing)
         } catch (Exception e) {
+            if (bib880) {
+                thing[HAS_BIB880] = bib880
+            }
             log.error("Failed to convert 880: $e", e)
```
olovy, line 72:
```suggestion
            log.error("Failed to convert 880 for record ${record[ID]}: $e", e)
```

olovy:
Must have been a whelktool script or MARC conversion that caused the bad data.
That language containers have to be maps is covered by JsonLdValidator.
https://github.com/libris/librisxl/blob/04169f3e71fafc0be7874676b57b32444ef67f0a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy#L184

whelk-core/src/main/groovy/whelk/converter/marc/RomanizationStep.groovy
kwahlin: Wasn't sure if we have a record id at this stage...do we?

kwahlin:
> The bad data must have been created by a whelktool script or the MARC conversion. JsonLdValidator checks that language containers are maps.
> 
> https://github.com/libris/librisxl/blob/04169f3e71fafc0be7874676b57b32444ef67f0a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy#L184

I couldn't find any language container that was an array either! So why did we get this error? :thinking:

kwahlin:
> > The bad data must have been created by a whelktool script or the MARC conversion. JsonLdValidator checks that language containers are maps.
> > https://github.com/libris/librisxl/blob/04169f3e71fafc0be7874676b57b32444ef67f0a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy#L184
> 
> I couldn't find any language container that was an array either! So why did we get this error? thinking

Maybe omit [794730d](https://github.com/libris/librisxl/pull/1249/commits/794730dc9b34cd449a6bcde33c95c21656434176) for now and keep an eye on the log the next few days to find these records and see what's going on. Given that record id can be found in the error message ([b94eb00](https://github.com/libris/librisxl/pull/1249/commits./b94eb00ca4d7bdca8dace7bc30e0a5f680e43ae6)).

@olovy

olovy:
Sounds good to 

> > > The bad data must have been created by a whelktool script or the MARC conversion. JsonLdValidator checks that language containers are maps.
> > > https://github.com/libris/librisxl/blob/04169f3e71fafc0be7874676b57b32444ef67f0a/whelk-core/src/main/groovy/whelk/JsonLdValidator.groovy#L184
> > 
> > 
> > I couldn't find any language container that was an array either! So why did we get this error? thinking
> 
> Maybe omit [794730d](https://github.com/libris/librisxl/pull/1249/commits/794730dc9b34cd449a6bcde33c95c21656434176) for now and keep an eye on the log the next few days to find these records and see what's going on. Given that record id can be found in the error message ([b94eb00](https://github.com/libris/librisxl/pull/1249/commits./b94eb00ca4d7bdca8dace7bc30e0a5f680e43ae6)).
> 
> @olovy

Sounds good to me

kwahlin:
Now definitely addresses https://jira.kb.se/browse/LXL-4112, see [6c4053d](https://github.com/libris/librisxl/pull/1249/commits/6c4053d9b1237774bfa3229fe98139e494e4a659). The problem was that we have 
`Whelk.converter.conversion.flatLinkedForm = true` 
vs 
`MarcFrameConverterSpec.converter.conversion.flatLinkedForm = false`. 

With `flatLinkedForm` set to `true`, `converted.get()['mainEntity']` evaluates to `null` causing the `bib880`'s to be incorrectly marked as handled and thus dropped.
When `flatLinkedForm` is off however, `converted.get()['mainEntity']` gives us the converted thing as expected, explaining why the tests still passed.

Also brought in another fix here, [6c4053d](https://github.com/libris/librisxl/pull/1249/commits/6c4053d9b1237774bfa3229fe98139e494e4a659), stopping unwanted `expressionOf` from being created. Realized that after conversion from marc it's not uncommon that language objects contain more than `@id`, e.g.
```
{
              "@id": "https://id.kb.se/language/eng",
              "code": "eng"
}
```
This was not taken into account before so the above entity was considered different to `{"@id": "https://id.kb.se/language/eng"}` which stops the expressionOf -> instanceOf normalization.

olovy (approved):
LGTM!

PR #1253, opened 2023-05-10
Feature/cxz

jannistsiroyannis:
On a basic level the idea is to periodically scan recent changes and compare them to what users have requested to be notified of. The notifications themselves are stored in separate table (referencing userdata and record versions). The notificiations can have the "handled" property flipped by POSTing.

Step by step it is used like this:

1. A user requests some notification by writing their user-data for example like so:
{
  .. other settings ..
  "requestedNotifications": [
    {
      "library": "https://libris.kb.se/library/Utb1",
      "triggers": [
        "https://id.kb.se/notificationtriggers/worktitle",
        "https://id.kb.se/notificationtriggers/primarycontribution"
      ]
    }
  ]
}
This example should be read as:
"Give me notifications when instances held by Utb1 either have their work titles or primary contributions modified"

2. Somebody (else?) changes a record that matches the above request

3. Get my notifications:
curl 'https://kblocalhost.kb.se:5000/_notifications' -H 'Authorization: Bearer NNNNNNNNNNNNNNNNNNNNNNNNN'
The bearer token is used to determine whos notifications are to be fetched and the result would look something like so:
{
  "notifications": [
    {
      "notificationID": 16,
      "mainEntityID": "http://kblocalhost.kb.se:5000/ljx0tr142tk2vrj#it",
      "triggers": {
        "triggered": [
          "https://id.kb.se/notificationtriggers/primarycontribution"
        ]
      },
      "handled": false
    }
  ]
}

4. When you are done with the notification you can mark it handled by doing:
curl -X POST 'http://kblocalhost.kb.se:5000/_notifications/2' -H 'Authorization: Bearer NNNNNNNNNNNNNNNNNNNNNNNNNNNNNN'

In order to determine if a change matches a request, two historic knowledge graphs (embellished documents) are constructed, each rooted with the same ID, but different in that they represent the graph around that ID at different points in time. These two graphs can then be compared and the differences may or may not result in some requested rule being triggered.

In order to perform this scanning periodically (but only in one place!) the "housekeeping" module is introduced. It is (very) rudimentary, but provides for periodic scanning and status/error reporting.

This is a WIP, and only a few rules are implemented so far.


klngwll:
> Give me notifications when instances held by Utb1 either have their work titles or primary contributions modified"

Still think we should be more precise in what we mean in the model (the predicate if you will) is it specifically the entity which is the ’recipient’ of the notification or the particular library which has the holdings would be the existing `heldBy`.

```
"requestedNotifications": [
{
"library": "https://libris.kb.se/library/Utb1",
```

PR #1255, opened 2023-05-11
Index blank works

olovy:
Add a new feature flag INDEX_BLANK_WORKS.
When enabled blank/local works are indexed as separate documents in ES so that the can be searched in the same way as linked stand-alone works.

If this works well and the need arises we could rewrite it in a more general fashion. Using vocab to decide what to index.


olovy:
Rebased on develop

olovy:
Rebased on develop

PR #1262, opened 2023-05-24
Bin duplicate gf terms

kwahlin:
Apparently some of the terms we link already exist (see https://jira.kb.se/browse/LXL-3167)


niklasl (approved):
LGTM

PR #1267, opened 2023-05-29
Add script for moving https://id.kb.se/term/barngf/Pocketb%C3%B6cker …

kwahlin:
…from work to instance


niklasl (approved):
LGTM! (Could move the g/f URI to a variable to prep for a reusable script (or function); but this works.)

Note that (at least) SBI needs to be informed of this change before we run it in PROD, to adapt their cataloguing practise.

PR #1269, opened 2023-06-02
Change uri for MTM holdings

poacherone:
All MTM holdings have (exactly) one uri in AssociatedMedia. This script changes it to their new uri.


kwahlin (approved):
LGTM!

PR #1270, opened 2023-06-08
Reorder postprocessing which is reversed on revert

niklasl:
The copying of reproductionOf data has to happen before 008 specific post-processing is run. Since the steps run in reverse order on revert, these steps are now put *after* the depending step.


olovy (approved):
:ship:

PR #1272, opened 2023-06-14
Add cleanup script.

jannistsiroyannis:


kwahlin:
Great! Have you checked that `uri` is always a valid Libris id (with #it)? Looks fine otherwise.

jannistsiroyannis:
> Great! Have you checked that `uri` is always a valid Libris id (with #it)? Looks fine otherwise.

I have now :P

kwahlin (approved):
:ship:

PR #1274, opened 2023-06-15
Search: Make matches- work with vocab terms

olovy:
For example `matches-domain.%40id=https%3A%2F%2Fid.kb.se%2Fvocab%2FEmbodiment`


niklasl (approved):
LGTM! :shipit:

PR #1275, opened 2023-06-16
Add representation of search results as Atom feeds

niklasl:


olovy (approved):
LGTM!

We can adjust the content format later.

niklasl:
3 years later; rebased on develop and adapted to the java rewrite. `SearchFeed` still in groovy, with some CompileStatic, but the main method generates the feed portion using `groovy.xml.StreamingMarkupBuilder`...

olovy (approved):
LGTM!

Let's merge this and then adapt rendering to use FresnelUtil machinery.

PR #1278, opened 2023-06-19
Don't limit MARC export trigger from works to changes in card

olovy:
Since :instanceOf is an :integral relation, parts of the linked work description outside the card might theoretically be included in the instance description.


kwahlin:
:ship:

PR #1279, opened 2023-06-21
Elastic: Add missing outer layer of retries when deleting from index

olovy:
These were only retried by the resilience4j layer. Handle in the same way as other requests.


lrosenstrom (approved):
Fix typo, otherwise LGTM!

PR #1281, opened 2023-07-06
Feature/lxl 4150 deduplicate contribution

kwahlin:
See https://jira.kb.se/browse/LXL-4150 for more info.


whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
```diff
@@ -179,6 +179,33 @@ class Normalizers {
...
+            findKey(thing, 'contribution') { value, _ ->
+                List<Map> contribution = asList(value)
+                def duplicates = contribution.countBy { asList(it.agent) }.findResults { it.value > 1 ? it.key : null }
+                duplicates.each { d ->
+                    def firstIdx = contribution.findIndexOf { asList(it.agent) == d }
+                    def first = contribution[firstIdx]
+                    def roles = []
+                    contribution.removeAll {
+                        if (asList(it.agent) == d) {
+                            roles += it.role
```
niklasl, line 194:
```suggestion
                            roles += asList(it.role)
```
I think?

kwahlin: Exactly! See variants of the `plus()` method: https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/List.html#plus(java.lang.Object).

whelk-core/src/main/groovy/se/kb/libris/Normalizers.groovy
```diff
@@ -179,6 +179,33 @@ class Normalizers {
         })
     }
 
+    static DocumentNormalizer contribution() {
```
niklasl, line 182: Rename to `mergeContributionsByAgent`?

kwahlin:
> We should revise this to _only_ apply a `mergeContributionOnAgent` in the whelktool script for now. See the Jira issue for detailed reasons.

Ok! Let's limit this to a one-off normalization on svsk records.

niklasl (approved):
LGTM! :+1:

PR #1284, opened 2023-08-09
Use .localhost instead of .kblocalhost.kb.se

andersju:
Using kblocalhost.kb.se and id.kblocalhost.kb.se as domains for local development is an (increasingly) bad idea because in modern browsers an increasing number of features are only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) -- essentially https://, http://127.0.0.1, http://localhost and http://*.localhost.

Among [restricted features](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts) is, for example, the Web Crypto API, so we have a [workaround](https://github.com/libris/lxlviewer/blob/70cb543a0b067e14318c81957bd752e0bd030ee9/lxljs/string.js#L14-L27) for local development.

Let's just switch to .localhost. (Later on we might _also_ want to use local self-signed certificates.)

To "migrate" from an old setup:

* Stop the REST API, nuxt-app, vue-client
* Use the corresponding branches in lxlviewer and devops unless they've already been merged
* Change `id.kblocalhost.kb.se` => `id.localhost` and `kblocalhost.kb.se` => `libris.localhost`  in the following files:
    * `/etc/hosts` 
    * `librisxl/secret.properties`
    * `lxlviewer/nuxt-app/.env`
    * `lxlviewer/vue-client/.env.development`
    * `/etc/apache2/sites-enabled/something.conf` (or equivalent)
* Import everything again: `fab xl_local app.whelk.import_work_example_data`
* Restart Apache
* In `lxlviewer/vue-client/.end.development`, change `VUE_APP_CLIENT_ID` (see chat or ask for new one)
* Start REST API, nuxt-app, vue-client

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,8 @@ cp secret.properties.in secret.properties
 # In secret.properties, set:
 # - elasticCluster to whatever you set cluster.name to in the Elasticsearch configuration above.
 vim secret.properties
-# Make sure kblocalhost.kb.se points to 127.0.0.1
-echo '127.0.0.1 kblocalhost.kb.se' | sudo tee -a /etc/hosts
+# Make sure libris.localhost points to 127.0.0.1
+echo '127.0.0.1 libris.localhost' | sudo tee -a /etc/hosts
 ```
 
 ### Importing test data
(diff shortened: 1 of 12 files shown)


jannistsiroyannis (approved):
Nånstans kommer detta att smälla, men det får vi nog ta när vi märker det! Bra att det blir fixat!

olovy (approved):
Agree, nice to have this fixed.

LGTM!

PR #1285, opened 2023-08-14
Clean up manage-whelk-storage.sh

olovy:
* [Remove obsolete --nuke-definitions](https://github.com/libris/librisxl/commit/3fcf971d551b6972384c34a709efda4abeb6ae26)
* [Don't create elasticsearch index](https://github.com/libris/librisxl/pull/1285/commits/7f1f14488609e5cf47294e0c959355a000e5712e)
    * Because fab tasks always use numbered indices + aliases now. This caused the initial docker-compose setup to fail after copying data.


olovy:
FYI

PR #1286, opened 2023-08-17
Fix intellij autocomplete etc inside whelktool scripts

olovy:
This makes intellij recognize whelktool and whelk-core classes in whelktool scripts.
It does not make the "scripts" directory part of whelktool.jar.
Stuff made available to the scripts through `javax.script.Bindings` like `selectBySqlWhere` is not recognized.

After importing the gradle project a manual step has to be performed for it to work 100%. Se comment.


kwahlin (approved):
:ok_hand:

PR #1287, opened 2023-08-17
Add IntelliJ GroovyDSL script for whelktool

olovy:
* Getting the correct type `DocumentItem` for the object passed to the `process` Closure doesn't work


andersju (approved):
:sunglasses: :ship:

PR #1288, opened 2023-08-18
Feature/merge works clean

kwahlin:
- All the "work code" has been moved a separate module `librisworks`
- A special tool (`WorkTool`) is no longer needed, instead the regular `WhelkTool` is used for document loading, saving, iterating etc., hence each step in the clustring/merging process is just a groovy script that runs with `WhelkTool`.
- Those scripts are wrapped in `run.sh`, which takes as arguments the environment to run against and (optionally) flags to pass on to WhelkTool. 
Example: `./run.sh qa --dry-run` 
I have run this locally to make sure that the functionality is maintained, i.e. getting the same result as when running from the old branch https://github.com/libris/librisxl/tree/feature/merge-works. Obviously we still need automatic tests as well.
- All files that are not actively used at this point have been excluded. They can still be found in the old branch should we need to pick up some of it again.


olovy (approved):
Nice!

Let's merge this and take it from there.

:ship:

kwahlin:
Replaced by https://github.com/libris/librisxl/pull/1290

PR #1291, opened 2023-08-22
Don't collapse the system with a db connection leak when starting wit…

jannistsiroyannis:
…hout ES available.


kaipoykio:
Verkar rimligt ;)

olovy (approved):
:ship:

kaipoykio (approved):

PR #1297, opened 2023-08-25
Feature/lxl 4243 move out content type from has part

kwahlin:
- [x] On import, automatically move solitary contentType out from hasPart and put it directly on the work instead.
- [x] Script for executing the above retroactively on current data.
- [x] ~Export? Multiple values in instanceOf.contentType generate a _single_ 336 with _repeated subfields_ whereas single-valued contentType in hasPart generate repeated 336.~ **Update:** No need to change export according to MNB.

See LXL-4243 for more info.


whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+    void modify(Map record, Map thing) {
+        moveContentType(thing)
+    }
+
+    boolean moveContentType(Map thing) {
+        def work = thing.instanceOf
+        def moved = work?.hasPart?.removeAll { p ->
+            if (p.keySet() == shape) {
+                work[contentType] = ld.asList(work[contentType]) + p[contentType]
```
olovy, line 15:
```suggestion
                work[contentType] = ld.asList(work[contentType]) as Set + p[contentType]
```

niklasl, line 15:
```suggestion
                work[contentType] = (ld.asList(work[contentType]) + p[contentType]).unique()
```

whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
```diff
@@ -0,0 +1,27 @@
...
+    def shape = [TYPE, contentType] as Set
+
+    void modify(Map record, Map thing) {
+        moveContentType(thing)
+    }
+
+    boolean moveContentType(Map thing) {
+        def work = thing.instanceOf
+        def moved = work?.hasPart?.removeAll { p ->
+            if (p.keySet() == shape) {
```
niklasl, line 14: In theory, someone could add a part with a more specific type than `Work`, which if different from the parent work could indicate that this is a proper part. (We should check this assumption though.)

whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
```diff
@@ -0,0 +1,27 @@
...
+                return true
+            }
+        }
+        if (work?.hasPart?.isEmpty()) {
+            work.remove('hasPart')
+        }
+        return moved
+    }
+
+    void unmodify(Map record, Map thing) {
```
niklasl, line 25: _If_ we need to have the same repeated 336:s on revert, we can add back repeated contentType values as parts here.

niklasl (approved):
Clean and clear, and with good specs! It would be wise to discuss, and if agreed upon address, comments below before merging.

whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
kwahlin: I found only one example where the type is not `Work` (https://libris.kb.se/katalogisering/w89srsd1t3wbnhzp). It looks like the part was added manually, however I can't tell why that needs to be a separate part. I guess no need to address this now since we have no deviating records from import?

whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
kwahlin: See updated description.

whelk-core/src/main/groovy/whelk/converter/marc/NormalizeContentTypeStep.groovy
niklasl: Yes, of course, you're absolutely right! This is only done for data coming from MARC, so at this point we know the type is only `Work`! Thus this is perfectly fine!

niklasl (approved):
LGTM! :ship:

PR #1298, opened 2023-08-31
Feature/fmt 18 illustrators to instance

kwahlin:
Updated criteria for when to keep illustrators on work according to https://jira.kb.se/browse/FMT-18. Let's try this and evaluate after next QA iteration.


klngwll (approved):
LGTM!

niklasl (approved):
LGTM! (Same thought regarding specific/reusable, but it's pretty clear that this is not an either/or, so let's just keep it in mind for "Works part II".)

PR #1299, opened 2023-08-31
Do not use PrimaryContribution on instances

niklasl:
When moving contributions to the instance, set them to type Contribution, to avoid there being multiple primary (100) fields when reverting to MARC.


kwahlin (approved):
LGTM!

PR #1300, opened 2023-09-05
Feature/lxl 4183 correct illustrators vs translators

kwahlin:
- When an agent X is stated as either illustrator or translator in responsibilityStatement, remove the other from the Contribution holding agent X since it's likely to be incorrect (of course nothing is removed should both illustrator and translator be stated in responsibilityStatement). All these removals are reported in ill-vs-trl.tsv for manual verification.
- Never add translator/illustrator role from another work in same cluster when either is already present.
- Match agents on acronym, e.g. match Anne-Marie Constant with A.M. Constant.
 
Plus some more minor improvements.

Don't expect a thorough review since the whole script is very messy due to new details being added continually. This is more of an FYI :)

### DIFF
--- a/librisworks/scripts/add-missing-contribution-data.groovy
+++ b/librisworks/scripts/add-missing-contribution-data.groovy
@@ -138,6 +138,20 @@ selectByIds(clusters.flatten()) { bib ->
     // if still no match, add constructed local Contribution with agent + roles extracted from responsibilityStatement
     modified |= addRemainingContributionsFromRespStatement(contribution, contributionsInRespStatement, normalizedNameToName, respStatement, id)
 
+    if (modified) {
+        bib.scheduleSave()
+    }
+}
+
+selectByIds(clusters.flatten()) { bib ->
+    def id = bib.doc.shortId
+    def work = bib.graph[1].instanceOf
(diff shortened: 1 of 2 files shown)


klngwll (approved):
As you mentioned a bit hard to follow the code :) but let's merge and test it thoroughly in QA.

niklasl (approved):
Thorough job! It looks good; but as you say, testing is the way to make sure. Do we want to mark these "gnarly" parts now so they're easier to utangle (or toggle) from the ideally uniform and reusable scripts in for other work-sets? (Perhaps the name is hint enough, or we could name it e.g. `svsk-specific/add-missing-conribution-data.groovy`?)

kwahlin:
> Thorough job! It looks good; but as you say, testing is the way to make sure. Do we want to mark these "gnarly" parts now so they're easier to utangle (or toggle) from the ideally uniform and reusable scripts in for other work-sets? (Perhaps the name is hint enough, or we could name it e.g. `svsk-specific/add-missing-conribution-data.groovy`?)

Yep, some parts are definitely very svsk specific and others may be reusable. Let's see further down the line what can be extracted and reused!

PR #1303, opened 2023-09-08
Add partNumber from instance title to work title

kwahlin:
When reusing an already existing work title on the merged (linked) work, pick up partNumber from instance title(s) if present.


olovy (approved):
LGTM!

As discussed, some specs for these would be nice to have.

PR #1305, opened 2023-09-11
Add cleanup script

kwahlin:


whelktool/scripts/cleanups/2023/09/lxl-4263-missing-description-modifier.groovy
```diff
@@ -0,0 +1,12 @@
+def where = """
+    changedIn = 'batch import'
+    and modified::date = '2023-09-08'
```
andersju, line 3:
```suggestion
    and modified > '2023-09-07' and modified < '2023-09-09'
```
I believe `::date` makes this query not use an index for modified and thus very very slow

andersju (changes requested):
See comment, otherwise LGTM! :)

andersju (approved):
:ship:

PR #1307, opened 2023-09-15
Fix revert for DDC broadMatch and narrowMatch

klngwll:
When adding DDC classification with broadMatch these should also revert to 083 instead of 065. Added bonus: narrowMatch can also be used removing an old TODO.

See LXL-3533 for details.


niklasl (approved):
LGTM!

PR #1309, opened 2023-09-15
Use Elasticsearch 8

andersju:
Changes necessary for XL to work with Elasticsearch 8. Main change is that ES 8 has security (and TLS) enabled by default, so we now have to user+pass over HTTPS when talking to the ES cluster.

The devops part has already been merged.

How to tune things for search speed is a separate issue that we can look at more once we have QA running in the new ES cluster. The increased `number_of_shards` / `number_of_replicas` is just a starting point based on our current usage in the main cluster and current recommendations (and anyway we should make `number_of_shards` and `number_of_replicas` configurable, but that's a devops thing).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Related external repositories:
 
 ## Dependencies
 
-The instructions below assume an Ubuntu 20.04 system (Debian should be identical), but should work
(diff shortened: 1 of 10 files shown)


olovy (approved):
LGTM

PR #1310, opened 2023-09-21
Parameterize frame depth limit

niklasl:
Use depth limit 2 by default (for ES), and limit 3 for MARC conversion.


kwahlin (approved):
Looks good for now.

PR #1313, opened 2023-09-28
Feature/lxl 4266 fix roles not moved on import

kwahlin:
Not the neatest solution but should fix https://jira.kb.se/browse/LXL-4266. We should review the whole normalizer vs postprocessing machinery at some point.

- [x]  I have run integTest with required resources.


niklasl (approved):
LGTM! :ship:

PR #1314, opened 2023-10-04
Only log successful ES queries if they took more than a certain amount of time

andersju:
Currently the API whelk log in production has few million lines like this every day:
```
INFO whelk.component.ElasticSearch - ES query took 48 (46 server-side)
```
This is a) not particularly useful and b) makes it more annoying to find other things in the log. Let's write a log entry only if the query exceeded a certain amount of time.


niklasl (approved):
:100: :ship:

kwahlin (approved):
:ship:

PR #1315, opened 2023-10-06
Allow some relations to link to deleted resources

kwahlin:
See https://jira.kb.se/browse/LXL-4209

We may want to allow certain properties regardless of where they appear, but let's be specific to begin with.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLd.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLd.groovy
@@ -48,7 +48,10 @@ class JsonLd {
     static final List<String> NS_SEPARATORS = ['#', '/', ':']
 
     static final List<String> NON_DEPENDANT_RELATIONS = ['narrower', 'broader', 'expressionOf', 'related', 'derivedFrom']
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -48,6 +48,7 @@ class JsonLd {
     static final List<String> NS_SEPARATORS = ['#', '/', ':']
 
     static final List<String> NON_DEPENDANT_RELATIONS = ['narrower', 'broader', 'expressionOf', 'related', 'derivedFrom']
+    static final List<String> ALLOW_LINK_TO_DELETED = ['meta.derivedFrom', 'hasTitle.source', 'translationOf.hasTitle.source', 'instanceOf.hasTitle.source', 'instanceOf.translationOf.hasTitle.source']
```
niklasl, line 51:
```suggestion
    static final List<String> ALLOW_LINK_TO_DELETED = [
        'meta.derivedFrom', 'hasTitle.source',
        /* following are combinations only needed while there are local unlinked works */
         'translationOf.hasTitle.source', 'instanceOf.hasTitle.source', 'instanceOf.translationOf.hasTitle.source']
```

niklasl (approved):
LGTM!

(See suggested comment.)

PR #1316, opened 2023-10-12
Feature/cxz

jannistsiroyannis:
The email part of the CXZ feature.


housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+            Map<String, List<String>> notificationsByUser = new HashMap<>()
+            while (resultSet.next()) {
+                String id = resultSet.getString("id")
+                generateNotificationsForChangedID(id, heldByToUserSettings, from.toInstant(),
+                        until.toInstant(), affectedInstanceIDs, notificationsByUser)
+            }
+
+            notificationsByUser.keySet().each { email ->
+                List<String> notifications = notificationsByUser.get(email)
+                sendEmail(senderAddress, email, "Förändingsmeddelande", notifications.join("\n"))
```
andersju, line 117:
```suggestion
                sendEmail(senderAddress, email, "Förändringsmeddelande", notifications.join("\n"))
```
Hard coding the subject feels a bit icky. Using java.util.Locale (+ java.util.ResourceBundle) is - maybe? - a bit overkill for just a few lines given that we (AFAIK!) don't have a need for it anywhere else. So, it's just a few lines in total, maybe put them in secret.properties? :shrug:

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+                }
+            }
+        }
+    }
+
+    private String generateEmailBody(String changedInstanceId, List<String> triggeredCategories) {
+        Document changed = whelk.getStorage().load(changedInstanceId)
+        String mainTitle = Document._get(["@graph", 1, "hasTitle", 0, "mainTitle"], changed.data)
+        if (mainTitle == null)
+            mainTitle = "Ingen huvudtitel"
```
andersju, line 216: See previous comment about strings

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+            if (swedishLabel) {
+                changeCategories += swedishLabel + "(" + categoryUri + ") "
+            } else {
+                changeCategories += categoryUri + " "
+            }
+        }
+
+        return "Instansbeskrivning har ändrats\n\tInstans: " + Document.BASE_URI.resolve(changedInstanceId) +
+        "(" + mainTitle + ")\n\t" +
+        "Ändrinskatergorier: " + changeCategories + "\n"
```
andersju, line 231:
```suggestion
        "Ändringskategorier: " + changeCategories + "\n"
```
But see previous comment about strings

housekeeping/src/main/groovy/whelk/housekeeping/WebInterface.groovy
```diff
@@ -0,0 +1,74 @@
...
+    public abstract void trigger()
+
+    public ZonedDateTime lastFailAt = null
+    public ZonedDateTime lastRunAt = null
+}
+
+@CompileStatic
+@Log
+public class WebInterface extends HttpServlet {
+    private static final long PERIODIC_TRIGGER_MS = 10 * 1000
```
andersju, line 25: Should maybe be configurable through secret.properties?

andersju (changes requested):
Looks reasonable overall! With a disclaimer that I haven't been part of the discussions around this, so another pair of eyeballs would be good :) See comments. Also, maybe add a basic spec?

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+
+    public void trigger() {
+        // Build a multi-map of library -> list of settings objects for that library's users
+        Map<String, List<Map>> heldByToUserSettings = new HashMap<>();
+        {
+            List<Map> allUserSettingStrings = whelk.getStorage().getAllUserData()
+            for (Map settings : allUserSettingStrings) {
+                if (!settings["notificationEmail"])
+                    continue
+                settings?.requestedNotifications.each { request ->
```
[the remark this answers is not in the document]
jannistsiroyannis: :+1:

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+        return status
+    }
+
+    public void trigger() {
+        // Build a multi-map of library -> list of settings objects for that library's users
+        Map<String, List<Map>> heldByToUserSettings = new HashMap<>();
+        {
+            List<Map> allUserSettingStrings = whelk.getStorage().getAllUserData()
+            for (Map settings : allUserSettingStrings) {
+                if (!settings["notificationEmail"])
```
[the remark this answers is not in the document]
jannistsiroyannis: One way or another we must allow the user to choose/use an email address. If the address is incorrect (badly formed), it will not pass our smtp (and so no harm done). I'm not sure what else we could validate.

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
jannistsiroyannis: I agree, it feels icky!
But ResourceBundle (or secret.properties) would be far worse.
I still think this is the lesser evil.

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,429 @@
...
+                    if (!request instanceof Map)
+                        return
+                    if (!request["heldBy"])
+                        return
+
+                    String heldBy = request["heldBy"]
+                    if (!heldByToUserSettings.containsKey(heldBy))
+                        heldByToUserSettings.put(heldBy, [])
+                    List userSettingsForThisHeldBy = heldByToUserSettings[heldBy]
+                    userSettingsForThisHeldBy.add(settings)
```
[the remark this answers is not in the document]
andersju: I think the existing code _is_ wrong? Otherwise what's the point of
```
List userSettingsForThisHeldBy = heldByToUserSettings[heldBy]
userSettingsForThisHeldBy.add(settings)
```
? `userSettingsForThisHeldBy` is not used anywhere else in the code.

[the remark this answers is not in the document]
jannistsiroyannis: You can (as I did) get a reference to a list, and modify that list, without invalidating other references to that list (its changed in place).
The existing code, and your suggestion do the exact same thing, only yours is shorter and more readable.

Or are we talking about different things?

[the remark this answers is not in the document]
andersju: No, you're right. Sorry, my brain is still a bit fried. Nevermind :D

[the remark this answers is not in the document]
jannistsiroyannis: :+1:

andersju (approved):

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,428 @@
...
+                    String heldBy = request["heldBy"]
+                    if (!heldByToUserSettings.containsKey(heldBy))
+                        heldByToUserSettings.put(heldBy, [])
+                    heldByToUserSettings[heldBy].add(settings)
+                }
+            }
+        }
+
+        // Determine the time interval of changes for which to generate notifications.
+        Timestamp from = Timestamp.from(Instant.now().minus(1, ChronoUnit.DAYS)) // Default to last 24h if first time.
```
olovy, line 83: Why default to 24h? Better to default to a 0 interval, i.e. first run does nothing?

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
niklasl: Can there be no middleground; e.g. an `se.kb.libris`-namespaced class with constants?

housekeeping/src/main/groovy/whelk/housekeeping/NotificationSender.groovy
```diff
@@ -0,0 +1,428 @@
...
+            case "https://id.kb.se/changenote/ddcclassification":
+            case "https://id.kb.se/changenote/sabclassification":
+                historicEmbellish(instanceAfterChange, ["mainEntity", "classification"], after, changeNotes)
+                break
+            case "https://id.kb.se/changenote/serialrelation":
+                historicEmbellish(instanceAfterChange, ["mainEntity", "precededBy", "succeededBy"], after, changeNotes)
+                break
+            case "https://id.kb.se/changenote/primarycontribution": {
+                historicEmbellish(instanceAfterChange, ["mainEntity", "instanceOf", "contribution"], after, changeNotes)
+
```
niklasl, line 304: Lots of hardcoded domain terms here. Cf. https://github.com/libris/definitions/blob/develop/source/changenotes.ttl and https://github.com/libris/lxlviewer/blob/develop/vue-client/src/utils/changecategories.js for "less obscure" hardcoding? (Normally when data is repeated more than once in code, it's time to actually move it to the data.)

PR #1318, opened 2023-10-18
Reindexing: Wait until all threads done until logging "Done!"

olovy:
Also
* use logger instead of println
* only one decimal place in docs/second


andersju (approved):
LGTM!

PR #1319, opened 2023-10-23
Check exclude list also when actually about to remove doc

kwahlin:
Follow up on https://github.com/libris/librisxl/pull/1315 which allows the document we wish to delete to pass the dependency check at CRUD level, however it still got stopped at the next dependency check, before actually deleting from postgres.


olovy (approved):
LGTM

PR #1321, opened 2023-10-27
Add proper adminMetadata to new linked works

kwahlin:
See https://jira.kb.se/browse/FMT-20. Note that this does not include any changes to the display of the derivedFrom links since there is still a question mark hanging over that.


niklasl (approved):
LGTM! :ship:

klngwll (approved):
LGTM!

PR #1322, opened 2023-11-06
Add more fine-grained merge rules for SAB classification

kwahlin:
Rules explained in https://jira.kb.se/browse/FMT-204

### DIFF
--- a/librisworks/src/main/groovy/se/kb/libris/mergeworks/compare/Classification.groovy
+++ b/librisworks/src/main/groovy/se/kb/libris/mergeworks/compare/Classification.groovy
@@ -1,13 +1,7 @@
 package se.kb.libris.mergeworks.compare
 
 class Classification extends StuffSet {
-    // Terms that will be merged (values precede keys)
-    private static def norm = [
-            'uHc'                                                        : ['Hc,u'],
-            'uHce'                                                       : ['Hce,u'],
-            'Hc'                                                         : ['Hc.01', 'Hc.02', 'Hc.03'],
-            'Hc,u'                                                       : ['Hcf', 'Hcg']
-    ]
+    private static def sabPrecedenceRules = loadSabPrecedenceRules()
 
     @Override
     Object merge(Object a, Object b) {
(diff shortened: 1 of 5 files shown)


andand79 (approved):
Très bien!

librisworks/src/main/groovy/se/kb/libris/mergeworks/compare/Classification.groovy
```diff
@@ -79,7 +73,94 @@ class Classification extends StuffSet {
...
+            return b
+        }
+        return null
+    }
+
+    static sabPrecedes(String a, String b) {
+        def (equal, startsWith) = sabPrecedenceRules
+        def startsWithExceptions = ~/^Hcb|^Hdab/
+        def preferred = equal[b] ?: startsWith.find { b.startsWith(it.key) }?.value
+        if (preferred && !(a =~ startsWithExceptions)) {
```
olovy, line 99: `startsWithExceptions` also apply to "equals?"
In that case the naming is a bit confusing.

olovy (changes requested):
I think this needs some tests
Also: see comment about startsWithExceptions.

I don't suggest changing the implementation but the code is a bit hard to follow. It could be made clearer by for instance by making it more explicit what the sabPrecedenceRules data structure contains.

kwahlin:
> I think this needs some tests
> 
> I don't suggest changing the implementation but the code is a bit hard to follow. It could be made clearer by for instance by making it more explicit what the sabPrecedenceRules data structure contains.

I agree. I was a bit eager to get this in quickly.

olovy:
> > I think this needs some tests
> > I don't suggest changing the implementation but the code is a bit hard to follow. It could be made clearer by for instance by making it more explicit what the sabPrecedenceRules data structure contains.
> 
> I agree. I was a bit eager to get this in quickly.

Just a couple of tests on `tryMergeSabCodes` or `merge` is fine

olovy (approved):
LGTM

See comment about `startsWithExceptions`

PR #1324, opened 2023-11-08
Terminate script immediately (with explicit exit code) on error

kwahlin:
- Before, if a script contained several calls to `select` and one failed, the subsequent ones would still be run.
- The exit code is useful e.g. for controlling the course of events when running Whelktool scripts from within a shell script.

### DIFF
--- a/whelktool/src/main/groovy/datatool/WhelkTool.groovy
+++ b/whelktool/src/main/groovy/datatool/WhelkTool.groovy
@@ -693,7 +693,7 @@ class WhelkTool {
         }
         if (errorDetected) {
             log "Script terminated due to an error, see $reportsDir/ERRORS.txt for more info"
-            System.exit(2)
+            System.exit(1)
         }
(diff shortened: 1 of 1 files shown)


whelktool/src/main/groovy/datatool/WhelkTool.groovy
```diff
@@ -687,6 +691,10 @@ class WhelkTool {
             it.flush()
             it.close()
         }
+        if (errorDetected) {
+            log "Script terminated due to an error, see $reportsDir/ERRORS.txt for more info"
+            System.exit(2)
```
andersju, line 696:
```suggestion
            System.exit(1)
```
1 is conventionally the generic "an error occurred" exit code (in bash, exit code 2 means "Misuse of shell builtins")

andersju (commented):
See comment, otherwise LGTM!

andersju (approved):

PR #1325, opened 2023-11-09
Feature/lxl 4279

kwahlin:
Should cover https://jira.kb.se/browse/LXL-4279. See individual commits.

Regarding a4c8bba and 6dd287a: All scripts don't necessarily have to be run _during_ the release, I just think it's good to keep all the "preparation" scripts in one place and we can figure out exactly how to roll them out when we plan the release.


librisworks/src/main/groovy/se/kb/libris/mergeworks/Util.groovy
```diff
@@ -78,20 +82,30 @@ class Util {
...
+            if (copy['subtitle'] || copy['titleRemainder']) {
+                DocumentUtil.traverse(copy) { value, path ->
+                    if (('subtitle' in path || 'titleRemainder' in path) && value instanceof String) {
+                        if (genericSubtitle(value)) {
+                            new DocumentUtil.Remove()
+                        } else {
+                            ((List) value.split(':')).with {
+                                if (it.size() > 1 && genericSubtitle(it.last().trim())) {
+                                    new DocumentUtil.Replace(value.replaceFirst(~/\s*:.+$/, ''))
+                                }
```
olovy: Can you add a a comment/example explaining why we do this?

olovy (approved):
LGTM!

librisworks/src/main/groovy/se/kb/libris/mergeworks/Util.groovy
kwahlin: See [c86d9ee](https://github.com/libris/librisxl/pull/1325/commits/c86d9ee16d438f059bcf4449ad18bfe9ca8b5b50).
When there's time I'll go through the whole module and document more of the code.

PR #1328, opened 2023-11-14
Do not suppress work titles with source on translations

niklasl:
The rules rely on NormalizeWorkTitles doing its juggling in the background, and cover a bit more than what is currently done by that post-processing.

(Eventually we might come to remove this "source on title blocks bib 240" altogether.)

### DIFF
--- a/whelk-core/src/main/resources/ext/marcframe.json
+++ b/whelk-core/src/main/resources/ext/marcframe.json
@@ -7469,15 +7469,8 @@
     },
 
     "240": {
+      "IMPORTANT": "Actual results are subsequently modified by the NormalizeWorkTitles postProcessing step!",
       "match": [
-        {
-          "when": {
-            "NOTE": "Ignore title with source",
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
LGTM!

niklasl:
> :star for documentation! Maybe add the Primary Contribution to the 240 examples and a 130 example just to make sure they both work?

The specs run before all post-processing; so the 130/240 juggling isn't part of the equation. This is a bit obscure, as any advanced post-processing cannot be reflected in the specs (in theory throughout all of marcframe, not just for these two fields).

Not sure what kind of comment is needed (there are many specs in 240 which aren't "real" anymore due to the title "juggling"); this PR already adds:
```json
"IMPORTANT": "Actual results are subsequently modified by the NormalizeWorkTitles postProcessing step!",
```
which may be enough?

klngwll (approved):
LGTM!

PR #1330, opened 2023-11-16
Feature/lxl 4209 allow outgoing links to deleted

kwahlin:
See https://jira.kb.se/browse/LXL-4209. Third try!

### DIFF
--- a/whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
+++ b/whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
@@ -119,6 +119,7 @@ class LinkFinder {
 
         String id = data.get("@id")
         if (id != null && data.keySet().size() == 1) {
+            // Path to same form as in lddb__dependencies.relation
             String normalizedPath = (path.take(2) == Document.recordPath
                     ? [JsonLd.RECORD_KEY] + path.drop(2)
                     : (path.take(2) == Document.thingPath ? path.drop(2) : path)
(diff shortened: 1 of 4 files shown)


whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
```diff
@@ -114,11 +114,19 @@ class LinkFinder {
...
         if (id != null && data.keySet().size() == 1) {
-            String primaryId = lookupPrimaryId(id)
+            // Path to same form as in lddb__dependencies.relation
+            String normalizedPath = (path.take(2) == Document.recordPath
+                    ? [JsonLd.RECORD_KEY] + path.drop(2)
+                    : (path.take(2) == Document.thingPath ? path.drop(2) : path)
+            )
+                    .findAll { it instanceof String }
+                    .join('.')
+            String primaryId = lookupPrimaryId(id, normalizedPath)
```
niklasl, line 129: Just so I can grasp the context; what was failing here before?

kwahlin: Nothing was failing before, however `lookupPrimaryId()` would throw `LinkValidationException` regardless of which relation links to a deleted record.

My previous attempts at https://jira.kb.se/browse/LXL-4209 only fixed the opposite case, i.e. enabling deletion of records with _incoming_ links via certain relations. This new fix is for allowing to delete/save a record with _outgoing_ links to deleted records via these relations.

kwahlin: Example:
`<WorkRecord> :derivedFrom <InstanceRecord1>, <InstanceRecord2>`
If <InstanceRecord2> is deleted, then you wouldn't be able to delete or resave <WorkRecord> due to the broken link.

niklasl: Ah; perfect; now I see it! Good work!

niklasl (approved):
LGTM! :ship:

PR #1332, opened 2023-11-16
Guarantee meaningful closeMatch links by matching primary contributors

kwahlin:
As per earlier discussion, add `closeMatch` link to another work from same cluster only if the other work has the same agent in `PrimaryContribution`.


niklasl (approved):
LGTM! :ship:

PR #1335, opened 2023-11-20
Update TRLD code

niklasl:
Update trld-java sources and adapt wrappers. Includes improvements to the Turtle/TriG serialization.

(Also used the option to exclude source line numbers from transpiled code to make these updates less noisy.)


niklasl:
Ah, yes, I fixed that and just pushed the change.

olovy (approved):
LGTM!

PR #1336, opened 2023-11-20
Include sameAs URIs when loading instance relators

kwahlin:
...because that's the given URI form in the batch import.


klngwll (approved):
LGTM

niklasl (approved):
Right; the linkfinder matches on the sameAs! LGTM!

PR #1337, opened 2023-11-20
Filter out works with certain genre/form terms from SVSK selection

kwahlin:
Since there's a risk that they won't apply to all instances of the linked work in the end.


klngwll (approved):
LGTM!

PR #1341, opened 2023-11-22
Feature/lxl 4279 keep ill and 9pu together

kwahlin:
Last minute fix. Supposed to be done already (included in https://jira.kb.se/browse/LXL-4279) but apparently the previous fix didn't work (see https://libris-qa.kb.se/katalogisering/z7hlzkdswv29r5tq).


niklasl (approved):
LGTM!

PR #1342, opened 2023-11-23
Return true when terms are equal

kwahlin:
You obviously expect **x** to imply **x**, for example when used here:
https://github.com/libris/librisxl/blob/b940941197682c84638e07c99d799a651e4494bb/librisworks/src/main/groovy/se/kb/libris/mergeworks/Doc.groovy#L273


olovy (approved):
:ship:

PR #1344, opened 2023-11-29
Do linkFinder.normalizeIdentifiers in RemoteSearchAPI

olovy:
After conversion from MARC the documents might contain entities with both `@id` and data

```
"language": { "@id": "https://id.kb.se/language/eng", "code": "eng" }
```

This trips up Embellisher which doesn't see it as a link (`@id` and size == 1). 
So https://id.kb.se/language/eng is not added to embellish data.

The same thing then happens in the cataloging interface, so the definition for English is not fetched there by `addMissingLinksToQuoted` either. This leads to a chip with the URI instead of the label in the form for the imported record.

bild

`linkFinder.normalizeIdentifiers` runs `clearReferenceAmbiguities` which clears out `code` in this example.
I guess we could use a cheaper version which doesn't do any id checks against the DB in this case.


kwahlin (approved):
LGTM

niklasl (approved):
LGTM!

PR #1345, opened 2023-11-30
Add script for moving age from genreForm to intendedAudience

kwahlin:
See https://jira.kb.se/browse/LXL-4379

### DIFF
--- a/whelktool/scripts/cleanups/2023/11/lxl-4379-gf-to-intendedAudience.groovy
+++ b/whelktool/scripts/cleanups/2023/11/lxl-4379-gf-to-intendedAudience.groovy
@@ -1,0 +1,28 @@
+Set iaLabels = ['0-3 år', '3-6 år', '6-9 år', '9-12 år', '12-15 år', '15']
+
+def where = """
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2023/11/lxl-4379-gf-to-intendedAudience.groovy
```diff
@@ -0,0 +1,28 @@
...
+and data #>> '{@graph,1,instanceOf,genreForm}' is not null
+"""
+
+selectBySqlWhere(where) { bib ->
+    def work = bib.graph[1].instanceOf
+    def toIntendedAudience = []
+    work.genreForm.removeAll { gf ->
+        if (gf.prefLabel in iaLabels) {
+            toIntendedAudience.add(gf.prefLabel)
+        }
```
jannistsiroyannis, line 15: So the if "returns" as true here? You are just too clever for me, but I'll assume it works :)

jannistsiroyannis (approved):

PR #1349, opened 2023-12-07
Search: Fix sorting on hasTitle.mainTitle

olovy:
Broke since the the way to specify sorting on nested docs in ES has changed.

nested_path och nested_filter were deprecated in ES6 and removed in ES8. 
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/sort-search-results.html#nested-sorting


andersju (approved):
LGTM!

PR #1350, opened 2023-12-08
Feature/speed up merge works

kwahlin:
Not doing a selectBy for each document individually did the trick. Comes with a slight cost though as we instead need to keep lots of data in memory. So had to make some adjustments to compensate for the higher memory usage.

Tested in dev2 where the merging and linking part took only a couple hours. Reporting is slower but that's not important.

Planning to rehearse the whole thing one last time in QA on Monday. See https://jira.kb.se/browse/FMT-205 for more info.

### DIFF
--- a/librisworks/scripts/merge-works.groovy
+++ b/librisworks/scripts/merge-works.groovy
@@ -6,7 +6,6 @@ import static se.kb.libris.mergeworks.Util.workClusters
 
 maybeDuplicates = getReportWriter("maybe-duplicate-linked-works.tsv")
 multiWorkReport = getReportWriter("multi-work-clusters.html")
-multiWorkReport.print(Html.START)
 
 enum WorkStatus {
     NEW('new'),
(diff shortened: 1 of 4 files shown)


librisworks/scripts/merge-works.groovy
```diff
@@ -23,12 +22,17 @@ WorkStatus.values().each {
...
+def changedBy = 'SEK'
+def generationProcess = 'https://libris.kb.se/sys/merge-works'
+
+def newWorks = []
+def idToUpdatedWorkData = [:]
+def instanceIdToWorkId = [:]
+def dataForReports = []
+def dataForMultiWorkReport = []
 
+def clusters = new File(System.getProperty('clusters')).collect {it.split(/[\t ]+/) as List }
```
niklasl, line 34: These are local to each script thread, since they're `def`:ed, right?

kwahlin: These are used read-only inside selects (where threads are used) so should be safe.

niklasl (approved):
LGTM!

PR #1352, opened 2023-12-11
Do not drop contributions without roles

niklasl:
Fixes bug in ContributionByRoleStep, which dropped contributions without roles if there was at least one contribution with an instance-level role. The reason was that if some contributions where moved,  only those with explicit work-level roles where kept in the set of work contributions.

**CAUTION:** Only approve on github but do not merge this here; use `git flow hotfix finish` (I'll do that once approved unless :bus: ).


kwahlin (approved):
LGTM!

PR #1353, opened 2023-12-14
Feature/lxl 4418 instance gfs

kwahlin:
See https://jira.kb.se/browse/LXL-4418

### DIFF
--- a/whelktool/scripts/cleanups/2023/12/move-bearer-like-gfs-from-linked-work-to-instance.groovy
+++ b/whelktool/scripts/cleanups/2023/12/move-bearer-like-gfs-from-linked-work-to-instance.groovy
@@ -1,0 +1,97 @@
+/**
+ * LXL-4418
+ * Move bearer-like genre/form terms from work to instance like in
+ * ../../2020/08/lxl-3294-move-bearer-like-gfs-from-work-to-instance.groovy,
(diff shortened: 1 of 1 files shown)


jannistsiroyannis (approved):

whelktool/scripts/cleanups/2023/12/move-bearer-like-gfs-from-linked-work-to-instance.groovy
```diff
@@ -0,0 +1,97 @@
...
+        "Video dvd",
+        "Video vhs",
+        "Historiska faksimil",
+        "Deckare & spänning",
+        "Fantasy & SF",
+        "Idrott & friluftsaktiviteter",
+        "Hus & hem",
+        "Medicin & hälsa",
+        "Resor & geografi",
+        "Samhälle & politik"
```
niklasl, line 34: More like "uncontrolled" than "carrier-specific"; but we're currently moving those "out of sight" (not really logical (the content is still about or not about these), but that's another debate).

kwahlin: True, "bearer-like" doesn't really apply no longer since those were added.

PR #1355, opened 2023-12-15
Add script for deleting unwanted duplicate records

kwahlin:
Removing these ~500 records that was created by mistake the other day (Dec 13). 
The ids of the "original" records that were already known so the duplicates could be retrieved by matching on Mimer identifier. See Sparql query: https://api.triplydb.com/s/OpB6bJzdH (Tip: fold the `values` clause)


jannistsiroyannis (approved):

andersju (approved):
LGTM!

PR #1360, opened 2023-12-20
fixes corrupt uri

kaipoykio:


kwahlin (approved):
LGTM assuming that you've already verified that `isPrimaryTopicOf` and `uri` are always arrays.

kaipoykio:
Well, a testrun setting the values didn't break so I assume it is so... But no explicit checks are made for arrays as it is not contraindicated by testing, but maybe it should be done just for safety ;)

kwahlin:
All good then!

kaipoykio:
But you're right, should also check the existence of attributes... I'll be back ;)

kwahlin (approved):
:+1:

PR #1361, opened 2024-01-02
Don't generate ChangeObservation for minor title changes

olovy:
Edit distance cutoff maybe needs to be revised. Let's start like this.
Should changes in title prefix always lead to ChangeObservation?


lrosenstrom (approved):
LGTM!

lrosenstrom (approved):
:ship:

PR #1362, opened 2024-01-03
Feature/fmt 210 artist cartographer

kwahlin:
See description in https://jira.kb.se/browse/FMT-210

### DIFF
--- a/whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
+++ b/whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
@@ -5,8 +5,7 @@
  * See FMT-210 for details.
  */
 
-unhandled = getReportWriter('unhandled.txt')
-addedRole = getReportWriter('added-role.tsv')
+unhandled = getReportWriter('unhandled.tsv')
 addedPrimary = getReportWriter('added-primary.tsv')
 
 def roles = [
@@ -47,6 +46,8 @@ selectBySqlWhere("collection = 'bib' and deleted = false and data#>>'{@graph,1,c
 
     // Reaching here means that one of the concerned roles was moved from PrimaryContribution on work to "just" Contribution on instance by lxl-2512.
 
+    def currentPrimary = work['contribution']?.find { it['@type'] == 'PrimaryContribution' && it['agent'] == primaryBefore['agent'] }
+
     // Decide which role should be in PrimaryContribution based on work type.
     def newPrimaryRole = work['@type'] == 'StillImage'
             ? ['@id': 'https://id.kb.se/relator/artist']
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
```diff
@@ -0,0 +1,77 @@
...
+    if (!work || work['@id']) return
+
+    def newestToOldestVersion
+    try {
+        newestToOldestVersion = bib.getVersions().reverse()
+    } catch (Exception e) {
+        println(e.getStackTrace())
+        return
+    }
+    def versionTouchedByScriptIdx = newestToOldestVersion.findIndexOf { it.data['@graph'][0]['generationProcess']?['@id'] == scriptId }
```
olovy, line 37: Shouldn't this lookup be in the reverse order? "oldestToNewest"
The oldest version with generationProcess == scriptId is the one that was touched by the script.

whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
```diff
@@ -0,0 +1,77 @@
...
+            // The current role(s) is neither artist nor cartographer, report.
+            unhandled.println([id, work['@type'], asList(currentPrimary['role']).collect(roleShort)].join('\t'))
+        }
+    } else {
+        // The agent is no longer present in work PrimaryContribution, readd it with relevant role.
+        def newPrimary = ['@type': 'PrimaryContribution', 'agent': primaryBefore['agent'], 'role': [newPrimaryRole]]
+        work['contribution'] = [newPrimary] + asList(work['contribution'])
+        addedPrimary.println([id, newPrimary['agent'], newPrimary['role'].collect(roleShort)].join('\t'))
+        bib.scheduleSave()
+    }
```
olovy, line 76: Don't we have to check if there is currently _another_ `PrimaryContribution` agent on the work?

olovy (commented):
See comments. Otherwise LGTM.

whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
kwahlin: You are right, I'll fix that.

whelktool/scripts/cleanups/2023/12/fmt-210-restore-primary-contributors.groovy
kwahlin: I don't think that's necessary since if there is another PrimaryContribution it also has different agent (see line 49). Maybe there shouldn't be two PrimaryContribution ideally but since that was the case before I think it's ok to revert to that state.

kwahlin: I don't think that's necessary since if there is another PrimaryContribution it also has different agent (see line 49). Maybe there shouldn't be two PrimaryContribution ideally but since that was the case before I think it's ok to revert to that state.

olovy (approved):
LGTM!

PR #1364, opened 2024-01-16
(CXZ) Check lifeSpan in greater detail.

jannistsiroyannis:
It is not enough to check if the property lifeSpan is added/changed as a whole. Instead, the individual dates (birth and death) must be compared.


andersju (approved):
I only had a quick look but LGTM!

PR #1370, opened 2024-01-26
Fix sorting on fields with numeric_extractor, e.g. publication.year

olovy:
Some fields are indexed as text with the "numeric_extractor" analyzer. See [libris_config.json](https://github.com/libris/librisxl/blob/develop/librisxl-tools/elasticsearch/libris_config.json). This keeps four-digit years and throws away everything else. For example "℗2006" -> "2006", "uuuu" -> null. They are also indexed as keyword fields under <property>.keyword which preserves the raw value.

For sorting we want to use the analyzed value, i.e the four-digit year. By default elastic puts missing values last so this works both ascending and descending.

Fixes sorting on publication.year, publication.startYear, publication.endYear.

This also fixes filtering on years like "nnnn". Before they showed up in facets but didn't work.
https://libris-qa.kb.se/katalogisering/search/libris?q=*&@type=Instance&_limit=20&and-publication.year=nnnn


niklasl (approved):
:100: LGTM

PR #1375, opened 2024-02-21
Use embedded jetty instead of tomcat

olovy:
Package all XL servlet applications as uber-jars with an embedded jetty server. (Instead of as WAR files).

Background: The gretty plugin that we used to run wars locally pulled in the gradle version of groovy (3.0.18) which currently blocks us from upgrading to Java 21.

* Upgrade to Servlet API 4 / EE8 as a first step. We can try moving to Servlet 6 / EE10 after this.
* Currently all configuration is done programmatically. If we want, we could use jetty.xml and friends for some parts.

Running with gradle task example
```
cd librisxl/rest
../gradlew -Dxl.secret.properties=$HOME/secret.properties-local appRun
```

Building and running jar example
```
cd librisxl/rest
../gradlew jar
java -Dxl.secret.properties=$HOME/secret.properties-local -jar build/libs/rest.jar
```

New configuration parameters. Set as system properties. 
(The `appRun` gradle task sets different ports for the different apps)
* `xl.http.port` HTTP port. Default 8180
* `xl.http.static` Directory for static files served under `/static/*`. Default `/srv/static`
* `xl.http.maxConnections` Max HTTP connections. Default 500
* `xl.logRoot`. Directory for whelk and access logs. Default `./logs`

Example 
```
java -Dxl.secret.properties=$HOME/secret.properties-local -Dxl.logRoot=/var/log/xl/rest -Dxl.http.port=8080 -jar build/libs/rest.jar
```

Depends on corresponding devops changes in: https://github.com/libris/devops/pull/206


lrosenstrom (approved):
LGTM! Let's try it out!

PR #1376, opened 2024-02-26
Feature/upgrade gradle and groovy

kwahlin:
Required for Java 21 (https://jira.kb.se/browse/LXL-4437)

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLd.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLd.groovy
@@ -579,15 +579,15 @@ class JsonLd {
         s && (s.startsWith('https://') || s.startsWith('http://'))
     }
 
-    static List<List<String>> findPaths(Map obj, String key, String value) {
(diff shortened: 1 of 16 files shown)


whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -584,10 +584,10 @@ class JsonLd {
     }
 
     static List<List<String>> findPaths(Map obj, String key, Set<String> values) {
-        List paths = []
-        new DFS().search(obj, { List path, v ->
+        List<List<String>> paths = []
```
olovy, line 587: paths are not `List<String>`?
for example `['publication', 0, 'hasPart' , 0, 'place', '@type']`

olovy (approved):
See comment.

Other than that: LGTM

whelk-core/src/main/groovy/whelk/JsonLd.groovy
kwahlin: Right, good catch!

PR #1379, opened 2024-03-04
Allow title with source to be reverted to bib 240

niklasl:
Adresses [LXL-4442](https://jira.kb.se/browse/LXL-4442).


katnai1:
LGTM!

PR #1384, opened 2024-03-11
Feature/touch new works

lrosenstrom:
Adds script for touching works in batches. This script should be set to run once a day until all works have been resaved. Currently the batch size is set to 2000 works per run of the script.

Ids of touched works are saved to a file specified by the `touched-ids-file` flag, i.e. something like `java -Dtouched-ids-file=<path-to-file> ...`.   See devops repo PR: https://github.com/libris/devops/pull/210

The list of all works to resave is a copy of `/home/devops/globalchanges-PROD/librisxl/librisworks/reports/merge-works/prod-20231212/merged-works/CREATED.txt`.  (Maybe this needs to be fine-tuned?)


whelktool/scripts/syslib/new-works-touch.groovy
```diff
@@ -0,0 +1,20 @@
+// MUST be run with --allow-loud
+
+BATCH_SIZE = 2000
+File allWorksToResave = new File(scriptDir, "new-works-touch-ids.txt")
+String touchedIds = System.getProperty("touched-ids-file", '')
+File touchedIdsFile = new File(touchedIds)
+List touched = touchedIdsFile.readLines()
+def count = 0
```
kwahlin, line 8: Is this thread safe?

whelktool/scripts/syslib/new-works-touch.groovy
```diff
@@ -0,0 +1,20 @@
+// MUST be run with --allow-loud
+
+BATCH_SIZE = 2000
+File allWorksToResave = new File(scriptDir, "new-works-touch-ids.txt")
+String touchedIds = System.getProperty("touched-ids-file", '')
+File touchedIdsFile = new File(touchedIds)
+List touched = touchedIdsFile.readLines()
```
kwahlin, line 7: Better with Set instead of List?

kwahlin (approved):
See comment, otherwise LGTM

whelktool/scripts/syslib/new-works-touch.groovy
olovy: Nope, need to synchronize or use an `AtomicInteger` (`incrementAndGet`)

olovy:
We want to export 5k _Instances_ per day, not _Works_? Can be fixed by just replacing the id list.

olovy:
Could be a bit more efficient by removing already handled ids from the set/list and picking BATCH_SIZE ids before calling `selectByIds`. Now it will always fetch all documents from the original list. (doesn't really matter with this number of docs)

olovy:
> Could be a bit more efficient by removing already handled ids from the set/list and picking BATCH_SIZE ids before calling `selectByIds`. Now it will always fetch all documents from the original list. (doesn't really matter with this number of docs)

i.e.
```
def all = allWorksToResave.readLines() as Set
def touched = touchedIdsFile.readLines() as Set
def ids = (all - touched).take(BATCH_SIZE)
selectByIds( allWorksToResave.readLines() ) { ...
```

lrosenstrom:
> We want to export 5k _Instances_ per day, not _Works_? Can be fixed by just replacing the id list.

But we don't have a list of only instances. Or is that all MODIFIED - all CREATED?

lrosenstrom:
> > Could be a bit more efficient by removing already handled ids from the set/list and picking BATCH_SIZE ids before calling `selectByIds`. Now it will always fetch all documents from the original list. (doesn't really matter with this number of docs)
> 
> i.e.
> 
> ```
> def all = allWorksToResave.readLines() as Set
> def touched = touchedIdsFile.readLines() as Set
> def ids = (all - touched).take(BATCH_SIZE)
> selectByIds( ids ) { ...
> ```



> > Could be a bit more efficient by removing already handled ids from the set/list and picking BATCH_SIZE ids before calling `selectByIds`. Now it will always fetch all documents from the original list. (doesn't really matter with this number of docs)
> 
> i.e.
> 
> ```
> def all = allWorksToResave.readLines() as Set
> def touched = touchedIdsFile.readLines() as Set
> def ids = (all - touched).take(BATCH_SIZE)
> selectByIds( ids ) { ...
> ```

Sure, that looks more elegant.

lrosenstrom:
> > Could be a bit more efficient by removing already handled ids from the set/list and picking BATCH_SIZE ids before calling `selectByIds`. Now it will always fetch all documents from the original list. (doesn't really matter with this number of docs)
> 
> i.e.
> 
> ```
> def all = allWorksToResave.readLines() as Set
> def touched = touchedIdsFile.readLines() as Set
> def ids = (all - touched).take(BATCH_SIZE)
> selectByIds( ids ) { ...
> ```

Sure, that looks more elegant.

lrosenstrom:
Updated id list to contain only instances that where modified in the prod work extraction run + have an instanceOf relationship to a work.

olovy (approved):
LGTM!
:ship:

PR #1386, opened 2024-03-14
Embellish: Add support for integral inverse relations

olovy:
If for example `hasInstance` is defined as `:category` `:integral`. Then `['inverseOf': 'instanceOf']` in a lens definition should be treated as `:integral`, i.e. following the link "is free" and embellish stays on the same level.

If only `instanceOf` is `:integral` then `['inverseOf': 'instanceOf']` is not `:integral`

Use case: Filter Works on publication.year of their Instances. 
`@reverse.instanceOf.publication.year=2024`

We will need a mechanism to limit the amount of data pulled in when following the inverse of a relation that has already been "traversed in the forward direction". But this is a separate issue.
For example `Instance instanceOf Work @reverse.instanceOf Instance2`
This is already a problem with Instances showing up as search hits for a ISBN of another Instance linking to the same Work.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Embellisher.groovy
+++ b/whelk-core/src/main/groovy/whelk/Embellisher.groovy
@@ -67,8 +67,8 @@ class Embellisher {
 
         // TODO? reach doc from same doc via integral and non-integral
         // TODO reach doc vn non-integral and inverse integral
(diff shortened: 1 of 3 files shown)


lrosenstrom (approved):
LGTM!

PR #1388, opened 2024-03-20
Remove ArchivalTerms in LDR

klngwll:
Notes about obsolete, now removed `ArchivalUnit` and `marc:typeOfControl`. Also removed obsolete spec to pass integTest.

Addresses FMT-229.

* [x] Includes whelktool  script.
* [x] Tested whelktool script locally.
* [x] Name whelktool script after correct Jira issue.


niklasl (approved):
:shipit:

PR #1389, opened 2024-03-22
Feature/imagelinking

jannistsiroyannis:
Scan regularly, for *new* images (and link them up from existing instances).
Also scan for *new* instances (and link them to existing images where ISBN matches).

"Updated" images/instances are ignored.


andersju (approved):
Nice! I've only glanced at this, not tested. Approving so we can start doing some proper testing.

PR #1391, opened 2024-03-27
Feature/lws 41 xlql filtering

kwahlin:
Generate facets with selectable filters based on (requires!) https://github.com/libris/definitions/pull/472/files#diff-6696b36e4b8890f8ec4beba9b087927c296d79e1d41f894af4827fa00362f5b7 (More filters to be added, see https://jira.kb.se/browse/LWS-62)

~~Not yet tested with the new beta frontend, only the cataloguing.~~

 Some known problems/todos:
- Only works for "and filters" so far
- ~~Since we now put the whole query in the `_q` parameter, any added filter will also be visible in the input field as long as the input field is mapped to `_q` (as e.g. "textinput AND added:filter").~~
**UPDATE:** Put free text part in `_i` parameter. See https://jira.kb.se/browse/LWS-41 for more info. Update https://github.com/libris/lxlviewer/pull/984 accordingly.
- ~~Contributor filter does not work, it seems that the `#it` part of an id is interpreted as a comment, so e.g. `_q=contributor:"https://libris.kb.se/fcrtpljz1qp2bdv#it"` is received as `contributor:"https://libris.kb.se/fcrtpljz1qp2bdv` (missing `#it")` in the backend.~~ **SOLVED**
- The form for the "slices" in `PartialCollectionView.stats.sliceByDimension` should be reviewed, they are now unnecessarily verbose. (Marked as TODO in code)
- Each filter is now 1-to-1 mapped to a defined property, however some properties (e.g. `contributor`) may appear at both Work and Instance level. In these cases I've now included values from both levels in the filters. _If_ that's how we want it to work, we need to find a better solution for generating the "bucket count" because the current solution can generate inaccurate counts when merging fields from the "aggs" query result. It would probably be better if the merging was done already at runtime by Elastic when running the aggs query.
Not a big problem in practice but we should find a better a solution once we have a clearer picture of how filters "on different levels" should work.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/Crud.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/Crud.groovy
@@ -158,6 +158,8 @@ class Crud extends HttpServlet {
         } catch (InvalidQueryException e) {
             log.warn("Invalid query: ${queryParameters}")
             throw new BadRequestException("Invalid query, please check the documentation. ${e.getMessage()}")
+        } catch (RedirectException e) {
+            sendRedirect(request, response, e.getMessage())
         }
     }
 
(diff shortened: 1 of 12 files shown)


olovy (approved):
Let's try it out.

PR #1392, opened 2024-03-28
Change lddb__dependencies pk column to bigint (int no longer enough)

andersju:
We've already done this change in QA/STG/PROD. Doing it again in those environments should be almost-but-not-quite-a-noop.


olovy (approved):
:ship:

PR #1397, opened 2024-04-09
Don't lazy-load HouseKeeping servlet

andersju:
After the transition to embedded Jetty, HouseKeeping was lazy-loaded and wouldn't start until one visited `<server>:/8181/housekeeping/`. But it needs to start immediately. This seems to do the trick.

https://eclipse.dev/jetty/javadoc/jetty-12/org/eclipse/jetty/ee10/servlet/ServletHolder.html#setInitOrder(int)


olovy (approved):
:ship:

PR #1398, opened 2024-04-09
Change the way recipient lists for CXZ messages are compiled.

jannistsiroyannis:
This makes it so that:

1. Messages on instances goes to every library holding that instance

2. Messages on works goes to libraries holding non-electronic instances of the work.

3. Messages on anything else goes to libraries holding anything linking (in any number of steps) to the thing being messaged about.


housekeeping/src/main/groovy/whelk/housekeeping/InquirySender.groovy
```diff
@@ -170,6 +158,46 @@ class InquirySender extends HouseKeeper {
 
     }
 
+    private List<String> getRecipientsForLibraries(List<String> libraries, Map<String, List<Map>> heldByToUserSettings) {
+        List<String> recipients = []
```
olovy, line 162: Should be `Set`? (was in the old implementation)

olovy (approved):
LGTM!

PR #1400, opened 2024-04-10
Adapt cleanEtag to nginx

andersju:
Apache added the suffix `-gzip`, nginx (which we now use everywhere) adds the prefix `W/`. Just like before we need to clean it. Oherwise, if compression is enabled in nginx, the comparison in XL will fail and it won't be possible to, well, update records (at least not when the client has compression enabled).

So with this change we can re-enable compression for application/ld+json from libris.kb.se in nginx (for a while it's only been enabled for id.kb.se).

Tested on dev2.

https://jira.kb.se/browse/LXL-4401

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/CrudUtils.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/CrudUtils.groovy
@@ -276,7 +276,7 @@ class CrudUtils {
         }
 
         private static String cleanEtag(String str) {
-            return stripQuotes(str)?.replaceAll('-gzip', '')
+            return stripQuotes(str)?.replaceAll('W/', '')
         }
 
         private static String stripQuotes(String str) {


jannistsiroyannis (approved):
Will this break my local (still apache) setup? :) If so, thats ok I guess.

andersju:
> Will this break my local (still apache) setup? :) If so, thats ok I guess.

Probably! But you can un-break it; it should be enough to just remove `AddOutputFilterByType DEFLATE application/json application/ld+json` from your Apache config (in the VirtualHost sections, might be two of those lines)

olovy:
> > Will this break my local (still apache) setup? :) If so, thats ok I guess.
> 
> Probably! But you can un-break it; it should be enough to just remove `AddOutputFilterByType DEFLATE application/json application/ld+json` from your Apache config (in the VirtualHost sections, might be two of those lines)

PR #1403, opened 2024-04-12
Add LibrisID to the subject header on CXZ emails (on instances).

jannistsiroyannis:


andersju (approved):
LGTM!

olovy (changes requested):
It should be controlnumber :sob: 
See comment in XL-4475

jannistsiroyannis:
> It should be controlnumber, not systemId 😭 See discussion in XL-4475

But it *is* control number already :)

olovy (approved):
LGTM!

PR #1406, opened 2024-04-17
Index @reverse.itemOf as nested

olovy:
* Index `@reverse.itemOf` as nested
* Handle query parameter prefixes (`and-`, `or-`, `not-`, `exists-`) on nested fields/documents
* Make faceting work on nested documents
* Handle nested documents where `included_in_parent = false`
* Use the same type of filter inside nested docs as outside.
    * Makes e.g. prefix search on ISBN work

Not handled yet
* `and-matches`
* ranges (`min-`, `max-` etc.)

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -339,6 +339,14 @@
                 }
             },
             {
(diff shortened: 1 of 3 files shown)


andersju (approved):
Yep, nice work!

PR #1407, opened 2024-04-17
Add a short glossary to the REST API documentation.

jannistsiroyannis:


andersju (approved):
LGTM but while we're at it I think we should also explain what "lens" means (and the difference between "chip", "card" and "none").

poacherone:
> LGTM but while we're at it I think we should also explain what "lens" means (and the difference between "chip", "card" and "none" (default)).

Yes, please!

jannistsiroyannis:
> LGTM but while we're at it I think we should also explain what "lens" means (and the difference between "chip", "card" and "none" (default)).

Vill du fakta-granska åt mig? Linserna är lite diffusa även för mig ibland! :)

PR #1410, opened 2024-04-18
Support _o parameter in new search

kwahlin:
Use alternative parameter `_o` (instead of `o`) in the new search route for finding all links to the specified resource.


olovy (approved):
LGTM.

TODO: facets on relation (`p`)

PR #1414, opened 2024-04-23
Feature/query language various

kwahlin:
- Use prefixed URIs for a more compact normalized query form, e.g. "sao:Ekonomi" instead of "https://id.kb.se/term/sao/Ekonomi"
- Fix resource showing up as filter when on that resource's product page, e.g. an agent showing up in the contribution filter.
- Felt that some refactoring was necessary, mainly to distinguish between linked resources and literals already in the first disambiguation step.


jannistsiroyannis (approved):
LGTM!

PR #1415, opened 2024-04-23
Handle sorting in new style search API

olovy:
* Handle `_sort` parameter
* Refactor checking for ES field settings into `EsMappings`

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -339,7 +339,7 @@ public class SearchUtils2 {
             public Map<?, ?> toSortClause(XLQLQuery xlqlQuery) {
                 // TODO nested
                 return Map.of(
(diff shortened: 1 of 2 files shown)


rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
```diff
@@ -333,4 +328,52 @@ static class Offsets {
             }
         }
     }
+
+    static class Sort {
+        private enum Order {
+            asc,
```
kwahlin, line 334: Should be capital letters?

kwahlin (approved):
Good work! LGTM

rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
olovy: Made them lowercase to be compatible with ES DSL...
I think I will refactor and move all the ES stuff.

PR #1416, opened 2024-04-25
Feature/type facets

kwahlin:
Adds support for displaying both instance type and work type as facets. 

Since work and instance types are to be displayed as 'type' and 'carrier type' respectively it was necessary to map `rdf:type` to different aliases depending on which type of entities we're looking at in a search result. These mappings are now hardcoded in `XLQLQuery.getAliasMappings()`. Feels a bit ad hoc of course but does the trick for now. And I think at least the mechanism itself might be of use in the future since it's likely that we want to display properties differently depending on the context.

Required for this to work:
- https://github.com/libris/definitions/pull/482
- https://github.com/libris/lxlviewer/pull/1011


olovy (approved):
LGTM! :ship:

PR #1417, opened 2024-04-25
Deterministic sort order for *-search

olovy:
Sort results deterministically when doing *-search. Sort on doc id for now.


kwahlin (approved):
image

PR #1420, opened 2024-04-29
Add support for filtering with _p when using _o

olovy:
Not generating facets for `_p` yet since we need a new way of doing it. Probably by extending `reverseLinks` in the indexed doc with all different relations instead of just `totalItems`. So that we know what to generate stats for. The old API gets the facet counts from postgres which makes them wrong when filtering on any (ES) property.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -174,7 +174,8 @@ public class SearchUtils2 {
             var filters = new ArrayList<>(DEFAULT_FILTERS);
             if (object != null) {
                 if (predicates.isEmpty()) {
-                    filters.add(SimpleQueryTree.pvEqualsLink("_links", object));
+                    // TODO: this also generates for instance "filter": { "simple_query_string": { "fields": ["@reverse.instanceOf._links"
+                    filters.add(SimpleQueryTree.pvEqualsLiteral("_links", object));
                 } else {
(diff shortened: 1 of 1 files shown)


rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
```diff
@@ -154,7 +156,11 @@ public Map<String, Object> getPartialCollectionView(Map<String, Object> esRespon
         private SimpleQueryTree getFilteredTree() {
             var filters = new ArrayList<>(DEFAULT_FILTERS);
             if (object != null) {
-                filters.add(SimpleQueryTree.pvEqualsLiteral("_links", object));
+                if (predicates.isEmpty()) {
+                    filters.add(SimpleQueryTree.pvEqualsLink("_links", object));
```
kwahlin, line 177: `object` should actually be wrapped as a Literal rather than Link here since `_links` is just a list of strings (no proper `@id` links)

rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
```diff
@@ -154,7 +156,11 @@ public Map<String, Object> getPartialCollectionView(Map<String, Object> esRespon
...
+                } else {
+                    filters.addAll(predicates.stream().map(p -> SimpleQueryTree.pvEqualsLink(p, object)).toList());
```
kwahlin, line 179: Link -> Literal

kwahlin (changes requested):
See comment, otherwise LGTM

PR #1431, opened 2024-05-06
Handle nested fields in new search

kwahlin:
...to enable filtering on library (@reverse.instanceOf.@reverse.itemOf.heldBy).

I tried to simplify the code but I think it's still a bit messy. More comprehensive refactoring will be needed at some point but this should at least simplify the nested field building part.

There is still work to do for making all types of nested queries work (see TODOs in code).

We should index `contribution` and `@reverse.instanceOf.identifiedBy` as nested too for `isbn:x` and `author:x`to work properly.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
+++ b/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
@@ -39,6 +39,8 @@ import java.util.stream.Stream;
 import static whelk.component.ElasticSearch.flattenedLangMapKey;
 import static whelk.util.DocumentUtil.NOP;
 
+import static whelk.xlql.Disambiguate.RDF_TYPE;
+
 public class XLQLQuery {
     private final Whelk whelk;
     private final Disambiguate disambiguate;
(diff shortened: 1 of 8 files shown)


whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
```diff
@@ -58,12 +61,8 @@ public XLQLQuery(Whelk whelk) {
         this.esMappings = new EsMappings(whelk.elastic != null ? whelk.elastic.getMappings() : Collections.emptyMap());
     }
 
-    public QueryTree getQueryTree(SimpleQueryTree sqt) {
-        return new QueryTree(sqt, disambiguate);
-    }
-
     public QueryTree getQueryTree(SimpleQueryTree sqt, Disambiguate.OutsetType outsetType) {
-        return new QueryTree(sqt, disambiguate, outsetType);
+        return new QueryTree(sqt, disambiguate, outsetType, esMappings.nestedFields);
```
olovy, line 65: I think we should try to refactor this later and keep the ES stuff separate,.

olovy (approved):
LGTM!

PR #1432, opened 2024-05-07
Feature/xlql handle ambiguities

kwahlin:
Use designated query codes to avoid the problem with ambiguous property alias mappings such as "språk" mapping to both `language`/`associatedLanguage` and  "år" mapping to `year` when we instead want `yearPublished`.

Also prefer query code over property name to get a more concise normalized query form, e.g.:
image

See suggested codes: https://github.com/libris/definitions/pull/485

There are some issues with defaulting to the code instead of the property name in the normalized form:
- The codes are in Swedish so it will be weird when the user language is English.
  - We could create new corresponding codes in English (e.g. YEAR), however this option will require a new way of declaring them because as far as I understand language tagging an instance of a datatype such as LibrisQueryCode is not possible.
  - Another option is to not include any codes in the normalized form when the language is set to English and instead just default to the property name as before (e.g. yearPublished).
- When the language is set to Swedish on the other hand it will be weird having some codes in Swedish while those properties lacking associated code will be shown in English (the property name). We should probably instead default to the Swedish label when there is no code.

This PR also includes a small fix for the interaction between exact year filter and year range filter. Before this fix we could end up with useless queries such as "year>2000 and year<2010 and year=1990".

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -128,7 +128,7 @@ public class SearchUtils2 {
                         this.esQueryDsl = getEsQueryDsl();
                     } else {
                         qSqt.replaceTopLevelFreeText(i.get());
-                        throw new Crud.RedirectException(makeFindUrl(i.get(), qSqt.toQueryString()));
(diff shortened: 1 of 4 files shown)


kwahlin:
Merging this. We'll continue refining it later anyway.

PR #1434, opened 2024-05-10
Fix indexing of "virtual works"

olovy:
Delay flipping position of instance and work entities until after embellish and computed properties.
This fixes properties missing in the instance of the virtual work.
* reverse properties, e.g `@reverse.instanceOf.@reverse.itemOf`
* computed properties in `@reverse.instanceOf` e.g. ISBN forms


olovy:
FYI

PR #1435, opened 2024-05-10
Index alternate identifier forms for other nodes than mainEntity

olovy:
So that e.g. Works can be found by any ISBN form (10/13) of their Instances.

Doing `DocumentUtil.findKey` on the whole document (with embellishments) will have a performance impact on indexing. Maybe we should do this in a more directed way for selected property paths.


andersju (approved):
LGTM! Let's see what the performance impact is like.
(Anyway I guess we should look at overall index (and search) performance more in-depth as a separate issue)

PR #1436, opened 2024-05-14
Feature/remove wilcard from query when needless

kwahlin:
We don't want the wildcard symbol (*) to show up in the input field (`_i` should be empty) once any filter has been activated.


kwahlin:
See https://github.com/libris/librisxl/pull/1437.

PR #1437, opened 2024-05-21
Feature/lws 66 bool filters

kwahlin:
- Support aliased predefined filters that get special meaning in the query language, i.e. won't be interpreted as free text.
- Each filter is either active or inactive by default.
- The syntax for deactivating a filter in the query language is `NOT filterAlias` (e.g. `NOT excludeEplikt`) and for activating it's simply `filterAlias`.
- If a user tries to deactivate a filter that is _inactive_ by default by typing `NOT filterAlias` nothing will happen (that part will simply be excluded from the query string). Same thing if typing `alias` when the filter is _active_ by default.
- `selectHeader` describes what happens when switching from the _default mode_ (`_selected: false`) to the _non-default mode_ (`_selected: true`) (not necessarily from inactive to active).

Some "side effects":
- Selected filters are detected during disambiguation so free text node merging had to be moved to _after_ the disambiguation step.
- The redirection flow in `SearchUtils2.Query` has been simplified (see [9a83a6c](https://github.com/libris/librisxl/pull/1437/commits/9a83a6cb0b756cf63fc6009b6b848fa77c23ceef)). Before it was unnecessarily complex and based on a previous idea how the frontend would work. Now it's adjusted to how the frontend actually works. We can adjust as needed later should the frontend flow change.
- As a consequence of the above point, https://github.com/libris/librisxl/pull/1436 is no longer relevant. A simpler solution to the unwanted asterisk in the input field is included in this PR instead, see [6d6ae1e](https://github.com/libris/librisxl/pull/1437/commits/6d6ae1ed7759bdb93065c7800dc2b34348f9ac9e).

TODO:
- Decide how to represent these filters in https://github.com/libris/definitions/blob/develop/source/apps.jsonld. The filter declarations are now hardcoded and incomplete, see https://github.com/libris/librisxl/blob/53fa0aab279a77caaa6f01083e33d8cfa824c69f/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java#L664.
- Decide how selected filters should be reflected in `search.mapping`, i.e. modify this placeholder method https://github.com/libris/librisxl/blob/53fa0aab279a77caaa6f01083e33d8cfa824c69f/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java#L344
- Frontend

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
+++ b/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
@@ -516,10 +516,17 @@ public class XLQLQuery {
     }
 
     public Map<String, Object> getStats(Map<String, Object> esResponse, Map<String, Object> statsRepr, SimpleQueryTree sqt, Map<String, String> nonQueryParams, Map<String, String> aliases) throws InvalidQueryException {
+        var sliceByDimension = getSliceByDimension(esResponse, statsRepr, sqt, nonQueryParams, aliases);
+        var boolFilters = getBoolFilters(statsRepr, sqt, nonQueryParams);
+        return Map.of(JsonLd.ID_KEY, "#stats",
+                "sliceByDimension", sliceByDimension,
+                "_boolFilters", boolFilters);
+    }
+
+    private Map<String, Object> getSliceByDimension(Map<String, Object> esResponse, Map<String, Object> statsRepr, SimpleQueryTree sqt, Map<String, String> nonQueryParams, Map<String, String> aliases) {
         var buckets = collectBuckets(esResponse, statsRepr);
         var rangeProps = getRangeProperties(statsRepr);
-        var boolFilters = getDefaultBoolFilters(statsRepr);
-        return buildStats(buckets, sqt, nonQueryParams, aliases, rangeProps, boolFilters);
+        return buildSliceByDimension(buckets, sqt, nonQueryParams, aliases, rangeProps);
     }
 
     // Problem: Same value in different fields will be counted twice, e.g. contribution.agent + instanceOf.contribution.agent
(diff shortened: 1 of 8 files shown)


whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
```diff
@@ -587,11 +618,71 @@ private Map<String, Object> buildStats(
...
+    }
+
+    public Map<String, Map<String, Object>> getDefaultBoolFilters(Map<String, Object> statsRepr) throws InvalidQueryException {
+        // TODO: get from apps.jsonld
+        List<Map<String, String>> boolFilters = List.of(
+                Map.of("alias", "excludeEplikt", "filter", "NOT bibliography:\"sigel:EPLK\"", "status", "active", "selectHeader", "Inkludera e-plikt"),
+                Map.of("alias", "excludePreliminary", "filter", "NOT encodingLevel:(\"marc:PartialPreliminaryLevel\" OR \"marc:PrepublicationLevel\")", "status", "active", "selectHeader", "Inkludera kommande publiceringar"),
+//                Map.of("alias", "availableOnline", "filter", "_TODO", "status", "inactive", "selectHeader", "Endast material som finns online"),
+//                Map.of("alias", "isDigital", "filter", "_TODO", "status", "inactive", "selectHeader", "Endast digitalt material"),
+                Map.of("alias", "hasImage", "filter", "image:*", "status", "inactive", "selectHeader", "Endast resurser med bild")
```
olovy, line 669: I had imagined this as to separate levels of abstraction.

something like
"defining the symbol"
`"alias": "excludeEplikt", "filter": "NOT bibliography:\"sigel:EPLK\""`

"specifying the site default filter"
filter: "rdf:type: Work excludeEplikt excludePreliminary"

"defining the bool filter options"
`"filter": "excludeEplikt" something something bool

now we limit it so that every aliased filter has to have a bool filter.
maybe that is fine.

`selectHeader` should either be a `prefLabelByLang` or a key that can beused for translation client side.

"Endast resurser med bild" is ambigous -> "Endast resurser med omslags-/miniatyrbild"

olovy: selectHeader should either be a prefLabelByLang or a key that can beused for translation client side.

olovy: "Endast resurser med bild" is ambigous -> "Endast resurser med omslags-/miniatyrbild"

olovy (approved):
:ship:

PR #1438, opened 2024-05-21
Add facets for _p when doing _o search

olovy:
Add facets for predicates (_p) when doing object (_o) search. (RDF subject-predicate-object)
All `_p` values for `_o` are included even when a `_p` is selected. 
So that they can be displayed in a tabbed view in the interface.

For now 
* does an extra database access to lookup the type of the object
* does an extra roundtrip to elastic since the aggregates cannot be generated in the same query (could use multi search API)

TODO
* get the object type -> predicates mapping from configuration

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
+++ b/whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
@@ -32,14 +32,15 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static whelk.component.ElasticSearch.flattenedLangMapKey;
 import static whelk.util.DocumentUtil.NOP;
-
 import static whelk.xlql.Disambiguate.RDF_TYPE;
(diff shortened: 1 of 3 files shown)


whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
```diff
@@ -424,6 +427,40 @@ public static String quoteIfPhraseOrContainsSpecialSymbol(String s) {
...
+        return query;
+    }
+
+    private List<String> curatedPredicates(Entity object) {
+        // TODO don't hardcode
+        Map<String, List<String>> typeToRelations =
+                Map.of(
+                        "Agent", List.of("subject", "contributor", "publisher"),
+                        "Concept", List.of("subject", "genreForm", "hasOccupation", "fieldOfActivity"),
+                        "Work", List.of("subject", "contributor")
```
kwahlin, line 452: `[] :contributor [ a :Work ]` seems odd! Guess we'll evaluate these declarations when we move them to configurations anyway.

kwahlin (approved):
LGTM so far. Should be fine to merge! And we can sort out the configuration separately.

whelk-core/src/main/groovy/whelk/search/XLQLQuery.java
olovy: Yes, copy-paste typo. Fixed.
Yes.

PR #1439, opened 2024-05-21
ES: store number of incoming links per type, not just total

andersju:
Previously, when indexing a document into Elasticsearch, we'd only store the number of incoming links to the document (in `reverseLinks.totalItems`). This change makes it so that we also store number of links per type/relation, so that we can reduce the number of `/find` requests in frontend code. Right now a search like https://libris.kb.se/katalogisering/search/libris?q=katt&_limit=20&@type=Instance&_sort=-reverseLinks.totalItems&_offset=0 triggers 41 `/find.jsonld` requests -- the main one, plus two for each instance: one to see whether the currently-logged-in sigel has a holding for that instance, and one to get the total number of holdings for that instance. At least the latter we should be able to avoid easily.

So for example, given an instance like this:

```
whelk_dev=# select count(id) from lddb__dependencies where dependsonid = '9tm3815m5jsr1t6';
 count
-------
    96
(1 row)

whelk_dev=# select relation, count(id) from lddb__dependencies where dependsonid = '9tm3815m5jsr1t6' group by relation;
     relation     | count
------------------+-------
 itemOf           |    95
 meta.derivedFrom |     1
(2 rows)
```

 ...we'd now get this from ES:

```
~ curl -sk -u elastic:elastic "https://localhost:9200/libris_local/_doc/9tm3815m5jsr1t6"|jq '._source.reverseLinks'
{
  "@type": "PartialCollectionView",
  "totalItems": 96,
  "totalItemsByRelation": {
    "itemOf": 95,
    "meta.derivedFrom": 1
  }
}
```

Similarly, for an agent:

```
whelk_dev=# select count(id) from lddb__dependencies where dependsonid = '000n6qq312rjv2cf';
 count
-------
   312
(1 row)

whelk_dev=# select relation, count(id) from lddb__dependencies where dependsonid = '000n6qq312rjv2cf' group by relation;
           relation           | count
------------------------------+-------
 hasComponent.heldBy          |    20
 heldBy                       |    35
 meta.descriptionCreator      |    55
 meta.descriptionLastModifier |   202
 ```

=>

 ```
 ~ curl -sk -u elastic:elastic "https://localhost:9200/libris_local/_doc/000n6qq312rjv2cf"|jq '._source.reverseLinks'
{
  "@type": "PartialCollectionView",
  "totalItems": 312,
  "totalItemsByRelation": {
    "hasComponent.heldBy": 20,
    "heldBy": 35,
    "meta.descriptionCreator": 55,
[...]

PR #1441, opened 2024-05-24
In ES, store not just incoming link total but also per type (take 2)

andersju:
Same as https://github.com/libris/librisxl/pull/1439 except ES mapping has now been updated. With the former PR many things failed to index because of (for example) `object mapping for [reverseLinks.totalItemsByRelation.instanceOf] tried to parse field [instanceOf] as object, but found a concrete value`. These failures were all silent (!) because it seems like we don't actually check for ES errors when doing bulk indexing? (I'll create a separate issue for this.)

Anyway, the problem, and the solution, was exactly this: https://www.elastic.co/guide/en/elasticsearch/reference/8.13/subobjects.html (e.g. we wanted to set `reverseLinks.totalItemsByRelation.instanceOf.subject = 1` and `reverseLinks.totalItemsByRelation.instanceOf = 1` but, having already seen `instanceOf.subject`, `instanceOf` was considered to be an object).

( I copied the dynamic mapping for `reverseLinks` generated by ES and added `totalItemsByRelation` with `subobjects` set to false.)

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -205,6 +205,23 @@
                     }
                 }
             },
+            "reverseLinks": {
+              "properties": {
+                "@type": {
+                  "type": "keyword",
+                  "copy_to": [
+                    "_all"
+                  ]
+                },
+                "totalItems": {
+                  "type": "long"
+                },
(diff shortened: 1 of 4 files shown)


olovy (approved):
LGTM! :ship

PR #1442, opened 2024-05-27
Expose/log Elasticsearch bulk index errors

andersju:
Previously, when bulk indexing, we didn't check whether Elasticsearch actually succeeded in indexing the documents. So looking at the logs we couldn't tell if a reindex actually went well or not.

This simply exposes and logs indexing failures. For example, if in a batch of 300 two documents couldn't be indexed, it could now look like this:

```
[SimplePool-3] ERROR whelk.component.ElasticSearch - Failed indexing document: [_index:libris_local_29, _id:n3b4p48rq2lvnn5s, status:400, error:[type:document_parsing_exception, reason:[1:2] failed to parse: field name cannot be an empty string, caused_by:[type:illegal_argument_exception, reason:field name cannot be an empty string]]]
[SimplePool-3] ERROR whelk.component.ElasticSearch - Failed indexing document: [_index:libris_local_29, _id:nzj5p48rq2sng368, status:400, error:[type:document_parsing_exception, reason:[1:2] failed to parse: field name cannot be an empty string, caused_by:[type:illegal_argument_exception, reason:field name cannot be an empty string]]]
[SimplePool-3] WARN  whelk.component.ElasticSearch - Tried bulk indexing 300 docs: 298 succeeded, 2 failed. Took 637 ms
```

Whereas previously it'd just have said "Bulk indexed 300 docs in 637 ms".

See https://kbse.atlassian.net/browse/LXL-4511 for more details.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -238,7 +238,21 @@ class ElasticSearch {
             if (bulkString) {
                 String response = bulkClient.performRequest('POST', '/_bulk', bulkString, BULK_CONTENT_TYPE)
                 Map responseMap = mapper.readValue(response, Map)
-                log.info("Bulk indexed ${docs.count{it}} docs in ${responseMap.took} ms")
+                int numFailed = 0
+                if (responseMap.errors) {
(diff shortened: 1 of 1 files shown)


jannistsiroyannis (approved):
:ship:

PR #1443, opened 2024-05-28
Do AND-search instead of OR-search in multi-word valued fields

kwahlin:
See https://kbse.atlassian.net/browse/LWS-185. The default operator in Elasticsearch is apparently OR (thought it was AND) and since we didn't override this before a query like e.g. `titel:”harry potter och fången från azkaban”` was interpreted as `"hasTitle.mainTitle": "harry || potter || och || fången || från || azkaban"` giving lots of irrelevant hits.

Now we get instead:
`titel:”harry potter och fången från azkaban”` --> `"hasTitle.mainTitle": "harry && potter && och && fången && från && azkaban"`
`contributor:"astrid lindgren"` --> `"contribution.agent": "astrid && lindgren"`

Should do for now.


olovy (approved):
Let's add a TODO and merge this as a workaround

PR #1444, opened 2024-05-29
Default vs selected filters

kwahlin:
Should solve both https://kbse.atlassian.net/browse/LWS-188 and https://kbse.atlassian.net/browse/LWS-183


olovy (approved):
:ship:

olovy (approved):
LGTM

PR #1447, opened 2024-06-10
Add a script-running housekeeping template, and schedule running of W…

jannistsiroyannis:
…D linking script every other month.


olovy (approved):
LGTM

* `generationProcess.@id` will have a weird path + always end with `housekeeping_script_execution/script.groovy`? i.e. it is not possible to see which script was run on the document?
* I think the housekeeping whelktool scripts should log to somewhere a little more long lived.

PR #1449, opened 2024-06-12
When bulk indexing, retry if some docs failed due to ES error

andersju:
Since #1442 we log individual indexing errors when doing bulk indexing. With this PR we also retry indexing if an individual document failed due to an Elasticsearch server error.

If a document is b0rked and ES returns a 400 for that particular document, there's no point in automatically retrying, so we just log the error:

`ERROR whelk.component.ElasticSearch - Failed indexing document: [_index:libris_local_30, _id:pjxct28sr39vvz84, status:400, error:[type:document_parsing_exception, reason:[1:1090] object mapping for [reverseLinks.totalItemsByRelation.instanceOf.genreForm] tried to parse field [genreForm] as object, but found a concrete value]]`

But sometimes individual documents can't be indexed because of a problem (possibly temporary) in Elasticsearch itself, e.g.:

`ERROR whelk.component.ElasticSearch - Failed indexing document: [_index:libris_qa_11, _id:v9kz6128smvdjkvk, status:500, error:[type:mapper_exception, reason:timed out while waiting for a dynamic mapping update]]`

In this case we DO want to retry. So let's just throw an exception, which then ensures ElasticReindexer will [try that batch again](https://github.com/libris/librisxl/blob/develop/importers/src/main/groovy/whelk/reindexer/ElasticReindexer.groovy#L107-L120) (retrying only the exact documents that failed would be unnecessarily complicated; a batch is small enough).

https://kbse.atlassian.net/browse/LXL-4512

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -238,18 +238,31 @@ class ElasticSearch {
             if (bulkString) {
                 String response = bulkClient.performRequest('POST', '/_bulk', bulkString, BULK_CONTENT_TYPE)
                 Map responseMap = mapper.readValue(response, Map)
-                int numFailed = 0
+                List<String> idsFailedDueToDocError = []
+                List<String> idsFailedDueToESError = []
                 if (responseMap.errors) {
                     responseMap.items?.each { item ->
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
LGTM!

olovy (approved):
LGTM!

PR #1451, opened 2024-06-17
Fix ImageLinker state keys

andersju:
ImageLinker has `scanForNewInstances()` and `scanForNewImages()`. Both use `lddb__state` to get/set state (`linkedNewInstancesUpTo` + timestamp, `linkedNewImagesUpTo` + timestamp) but use the same key (`Image linker`), so only one can exist at a time (https://github.com/libris/librisxl/blob/develop/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy#L440).

As observed:

```
 whelk=> select * from lddb__state ;
     key      |                        value
--------------+-----------------------------------------------------
 Image linker | {"linkedNewImagesUpTo": "2024-06-16T14:40:20.811Z"}
(1 row)

...a moment later...

whelk=> select * from lddb__state ;
     key      |                         value
--------------+--------------------------------------------------------
 Image linker | {"linkedNewInstancesUpTo": "2024-06-16T13:52:54.761Z"}
(1 row)
```

Currently on dev ImageLinker keeps linking the same instances to the same images over and over. Hopefully this should fix it.

Should probably also insert these manually on DEV and QA pgsql.

### DIFF
--- a/housekeeping/src/main/groovy/whelk/housekeeping/ImageLinker.java
+++ b/housekeeping/src/main/groovy/whelk/housekeeping/ImageLinker.java
@@ -115,7 +115,7 @@ public class ImageLinker extends HouseKeeper {
         if (linkedNewInstancesUpTo.isAfter(linkNewInstancesSince.toInstant())) {
             Map<String, String> newState = new HashMap<>();
             newState.put(INSTANCES_STATE_KEY, linkedNewInstancesUpTo.atOffset(ZoneOffset.UTC).toString());
-            whelk.getStorage().putState(getName(), newState);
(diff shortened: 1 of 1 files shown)


andersju:
Should probably also make it so that [linkImage()](https://github.com/libris/librisxl/blob/d0f7c0a4b15d2f269376fa9c9072ae8e62ee796b/housekeeping/src/main/groovy/whelk/housekeeping/ImageLinker.java#L274) doesn't save the instance if the specified imageUri already exists in the instance?

olovy (approved):
LGTM

PR #1452, opened 2024-06-17
Rotate log4j logs automatically

andersju:
See https://kbse.atlassian.net/browse/LXL-4523

We could use the system's logrotate, but then we'd either have to introduce signal handling in XL, so that something like `kill -USR1 <xl service pid>` would make the XL service re-open its log file (without restarting the entire service), _or_ use logrotate's `copytruncate` option, but then "Note that there is a very small  time slice between copying the file and truncating it, so some logging data might be lost."

Log4j has a massive (and confusing) amount of options to do the same: https://logging.apache.org/log4j/2.x/manual/appenders.html#RollingFileAppender

After some trial and error this seems to work. This simple policy deletes logs after 30 days but there are many other possibilities.


olovy (approved):
:ship:

PR #1453, opened 2024-06-18
ImageLinker: check existing images before linking

andersju:
This adds a little overhead but ensures ImageLinker never saves a document if the specified image has already been linked (for example, if the state in `lddb__state` is whack or accidentally deleted or if the same image has returned with a newer `created` time).

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Document.groovy
+++ b/whelk-core/src/main/groovy/whelk/Document.groovy
@@ -161,7 +161,7 @@ class Document {
     List getImages() {
         def images = get(imagePath)
         if (images == null)
-            return null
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/Document.groovy
```diff
@@ -161,6 +160,8 @@ class Document {
 
     List getImages() {
         def images = get(imagePath)
+        if (images == null)
+            return null
```
olovy, line 164: return `Collections.emptyList()` to avoid having to deal with nulls?

olovy (approved):
LGTM!

PR #1456, opened 2024-07-22
Optionally copy document history when using WhelkCopier

andersju:
Related to https://kbse.atlassian.net/browse/LXL-4531

Currently we can't do some export tests properly because the dev environment lacks historical versions of records. This change makes it possible to optionally copy records from `lddb__versions` for each imported record (optionally restricted to type, e.g. Instance; otherwise with `import_work_example_data` we'd get a couple million historical auth records, which adds a bunch of extra time).

Also makes the "additional types" parameter more readable.

Example:

`java -Dxl.secret.properties=secret.properties -jar importers/build/libs/xlimporter.jar copywhelk temp_source_secret.properties temp_example_records.tsv --additional-types=Meeting,Person,Topic,Agent --include-items --copy-versions=Instance`


jannistsiroyannis (approved):
LGTM!

PR #1459, opened 2024-08-01
find API: add experimental spell checking

andersju:
A first stab at adding spell checking, to be used by Libris sök. Uses Elasticsearch's [phrase suggester](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/search-suggesters.html#phrase-suggester) with two generators as described in the ES docs.

Uses `_sortKeyByLang.sv` as the field to get suggestions from, as it seems like a reasonable choice; with e.g.  `_all` we'd get lots of "bad" suggestions since it contains lots of oft-repeated vocab stuff. (sv vs en shouldn't matter for these purposes.)

Quick testing on the dev data is promising. We might want to tinker with both the suggester configuration/query and the field(s) to target. This is just somerthing to start with.

With `/find?q=foobar&_spell=true` the spell checking is done in _addition_ to the regular query and returned along with the usual results.

With `/find?q=foobar&_spell=only` _only_ the spell checking query is performed.

(Possibly we want to keep it separate from /find though? :thinking: ...or have it both in /find, for internal use, and a bibspell-like thing (returning "bibspell-compatible" results would of course be trivial))

```
~ curl -s "http://localhost:8180/find?q=dynimical%20systems&_spell=true"|jq '._spell'
[
  {
    "text": "dynamical systems",
    "highlighted": "<em>dynamical</em> systems",
    "score": 0.005399387
  }
]
~ curl -s "http://localhost:8180/find?q=dynimical%20systemms&_spell=true"|jq '._spell'
[
  {
    "text": "dynamical systems",
    "highlighted": "<em>dynamical systems</em>",
    "score": 0.005399387
  }
]
```

curl directly against ES, local dev:
```
curl --request POST \
  --url https://localhost:9200/libris_local/_search \
  -k \
  -u elastic:elastic \
  --header 'Content-Type: application/json' \
  --data '{
  "suggest": {
    "text" : "dynimical system",
    "simple_phrase" : {
      "phrase" : {
        "field" : "_sortKeyByLang.sv.trigram",
        "size" : 1,
	"max_errors": 2,
        "direct_generator" : [
		{
			"field" : "_sortKeyByLang.sv.trigram",
			"suggest_mode" : "always"
        	},
		{
			"field" : "_sortKeyByLang.sv.reverse",
			"suggest_mode" : "always",
			"pre_filter" : "reverse",
			"post_filter" : "reverse"
      		}
	],
	"highlight": {
          "pre_tag": "<em>",
          "post_tag": "</em>"
        }
      }
    }
  }
}'
```

https://kbse.atlassian.net/browse/LWS-87

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
@@ -64,6 +64,7 @@ class SearchUtils {
         String lens = getReservedQueryParameter('_lens', queryParameters)
         String addStats = getReservedQueryParameter('_stats', queryParameters)
         String suggest = getReservedQueryParameter('_suggest', queryParameters)
+        String spell = getReservedQueryParameter('_spell', queryParameters)
 
         if (queryParameters['p'] && !object) {
             throw new InvalidQueryException("Parameter 'p' can only be used together with 'o'")
@@ -86,6 +87,7 @@ class SearchUtils {
                           '_lens' : lens,
                           '_stats' : addStats,
                           '_suggest' : suggest,
+                          '_spell': spell,
         ]
 
         Map results = queryElasticSearch(
@@ -121,6 +123,7 @@ class SearchUtils {
         List<String> predicates = pageParams['p']
         String addStats = pageParams['_stats']
         String suggest = pageParams['_suggest']
+        String spell = pageParams['_spell']
         lens = lens ?: 'cards'
 
         log.debug("Querying ElasticSearch")
@@ -132,7 +135,7 @@ class SearchUtils {
         // TODO Only manipulate `_limit` in one place
         queryParameters['_limit'] = [limit.toString()]
 
-        Map esResult = esQuery.doQuery(queryParameters, suggest)
+        Map esResult = esQuery.doQuery(queryParameters, suggest, spell)
         Lookup lookup = new Lookup()
         
         List<Map> mappings = []
@@ -182,7 +185,7 @@ class SearchUtils {
         } 
 
         Map stats = null
-        if (addStats == null || (addStats == 'true' || addStats == 'on')) {
+        if ((addStats == null || (addStats == 'true' || addStats == 'on')) && spell != "only") {
             stats = buildStats(lookup, aggregations,
                                makeFindUrl(SearchType.ELASTIC, stripNonStatsParams(pageParams)),
                                (total > 0 && !predicates) ? reverseObject : null,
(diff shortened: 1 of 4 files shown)


olovy (changes requested):
Very nice!

* Agree that `_sortKeyByLang.sv` is a reasonable starting point
* Perhaps `_spell=true` should be the default. :thinking: Has performance implications though.
* `_spell=only` might be a good option for clients to do spell checking "on the side" with a separate request without slowing down the main search query.
* We probably want to promote `_spell` (or whatever we decide on) to an official vocab (non-underscore) term . Let's do this later.
* TODO: Add support in "new style" search API after https://github.com/libris/librisxl/pull/1455 is merged

Changes requested:
* I think we should provide a link to the corrected query in the alongside the suggestion. So that the client doesn't have to do any URL manipulation.
* I think we should remap the terms in the result to keep the API free from elasticsearch details. 

What about:
```
curl -s "http://localhost:8180/find?q=flyttning%20och%20peldning&%40type=Instance&_spell=true" | jq '._spell'
[
  {
    "label": "flyttning och <em>pendling</em>",
    "view": { "@id": "/find?q=flyttning%20och%20pendling&%40type=Instance&_spell=true" }
  }
]
```
?

(I don't think the score is useful?)

andersju:
@olovy Sounds good! See latest commit.

(Probably we wouldn't want to actually show suggestions unless there are no search hits (or perhaps very few hits?), otherwise there might often be annoying suggestions even if you spell correctly. Though this is something for the client to decide.)

rest/src/main/groovy/whelk/rest/api/SearchUtils.groovy
```diff
@@ -221,11 +224,27 @@ class SearchUtils {
...
             stats[JsonLd.ID_KEY] = '#stats'
             result['stats'] = stats
         }
 
+        if (spell) {
+            result['_spell'] = []
+            esResult["spell"].each { Map suggestion ->
+                result['_spell'] << [
+                        'label': suggestion['text'],
+                        'labelWithMarkup': suggestion['highlighted'],
```
olovy, line 237:
```suggestion
                        'labelHTML': suggestion['highlighted'],
```

Found this usage of type coercion for description with type `rdf:HTML`.
We can stick to the same naming convention.

https://github.com/libris/definitions/blob/a320ff90c87a96e143826812a50afb0635c8cccd/source/app/help.jsonld#L12

olovy (approved):
LGTM!

See suggestion about using the same naming convention as `descriptionHTML`

PR #1461, opened 2024-08-02
Detect if index supports spell checking

olovy:
Otherwise we get a 400 bad query from elastic when requesting spelling suggestions.


andersju (approved):
:ship:

PR #1463, opened 2024-08-07
batchimport: log when skipping holding records due to library config…

olovy:
…uration

### DIFF
--- a/batchimport/src/main/java/whelk/importer/XL.java
+++ b/batchimport/src/main/java/whelk/importer/XL.java
@@ -240,11 +240,15 @@ class XL
             String librarySystemId = m_whelk.getStorage().getSystemIdByIri(libraryUri);
             if (librarySystemId == null)
             {
+                logger.warn(String.format("Could not find %s. Not importing holding record for %s",
+                        libraryUri, relatedWithBibResourceId));
                 return null;
             }
             Document libraryDoc = m_whelk.getDocument(librarySystemId);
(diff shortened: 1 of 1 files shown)


jannistsiroyannis:
Just consider that logger.warn(..) will now result in an email to all us! Is that really what we want for the 2nd of these? Maybe .info() instead ?

olovy:
> Just consider that logger.warn(..) will now result in an email to all us! Is that really what we want for the 2nd of these? Maybe .info() instead ?

Ok, changing to info. :+1: 
I don't think errwatch is set up on import.libris.kb.se yet though.

batchimport/src/main/java/whelk/importer/XL.java
```diff
@@ -240,11 +240,15 @@ private String importNewRecord(MarcRecord marcRecord, String collection, String
...
             if (librarySystemId == null)
             {
+                logger.warn(String.format("Could not find %s. Not importing holding record for %s",
+                        libraryUri, relatedWithBibResourceId));
                 return null;
             }
             Document libraryDoc = m_whelk.getDocument(librarySystemId);
             if (!libraryDoc.libraryIsRegistrant())
             {
+                logger.warn(String.format("%s does not have category Registrant. Not importing holding record for %s",
```
olovy, line 250:
```suggestion
                logger.info(String.format("%s does not have category Registrant. Not importing holding record for %s",
```

PR #1467, opened 2024-08-20
Add support for Item subclasses.

olovy:
Make it possible to have linked `SingleItem` (sv: "Exemplar") inside `Item` `hasComponent`.

* `SingleItem` is in collection `none` instead of `hold` so that we don't get any of the special handling for holdings
    * => it is possible to have multiple `SingleItem` records per library
    * => it is not involved in MARC export except as an embellished `hasComponent` of a regular holding record
* Access control for holdings is extend to all subclasses of `Item` instead of just collection `hold`
    * i.e. only owning library + global registrant can write `SingleItem`

See also
https://github.com/libris/definitions/pull/494

Later on `Item` might be replaced by `ItemHolding` for the holding, with `SingleItem` and `SomeItem` for 1 vs many items.
See https://github.com/libris/definitions/pull/483

### DIFF
--- a/rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
+++ b/rest/src/test/groovy/whelk/rest/api/CrudSpec.groovy
@@ -1190,7 +1190,7 @@ class CrudSpec extends Specification {
                                     "@type": "Item",
                                     "contains": "some new data",
                                     "heldBy":
-                                            ["code": "Ting"]]]]
+                                            ["@id": "https://libris.kb.se/library/Ting"]]]]
         request.getInputStream() >> {
             new ServletInputStreamMock(mapper.writeValueAsBytes(postData))
         }
@@ -1201,7 +1201,7 @@ class CrudSpec extends Specification {
             "POST"
         }
         LegacyIntegrationTools.determineLegacyCollection(_, _) >> {
-            return "bib"
+            return "hold"
         }
         request.getContentType() >> {
             "application/ld+json"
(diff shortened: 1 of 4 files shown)


kwahlin (approved):
LGTM

olovy:
Make it possible to have linked SingleItem (sv: "Exemplar") inside Item hasComponent.

    SingleItem is in collection none instead of hold so that we don't get any of the special handling for holdings
        => it is possible to have multiple SingleItem records per library
        => it is not involved in MARC export except as an embellished hasComponent of a regular holding record
    Access control for holdings is extend to all subclasses of Item instead of just collection hold
        i.e. only owning library + global registrant can write SingleItem

See also
https://github.com/libris/definitions/pull/494

Later on Item might be replaced by ItemHolding for the holding, with SingleItem and SomeItem for 1 vs many items.
See https://github.com/libris/definitions/pull/483

olovy:
Make it possible to have linked SingleItem (sv: "Exemplar") inside Item hasComponent.

    SingleItem is in collection none instead of hold so that we don't get any of the special handling for holdings
        => it is possible to have multiple SingleItem records per library
        => it is not involved in MARC export except as an embellished hasComponent of a regular holding record
    Access control for holdings is extend to all subclasses of Item instead of just collection hold
        i.e. only owning library + global registrant can write SingleItem

See also
https://github.com/libris/definitions/pull/494

Later on Item might be replaced by ItemHolding for the holding, with SingleItem and SomeItem for 1 vs many items.
See https://github.com/libris/definitions/pull/483

PR #1468, opened 2024-08-26
Add support for spell check in new search

kwahlin:
Continuation of https://github.com/libris/librisxl/pull/1459 which added spell check support to the cataloguing version of the search API. Now adding the equivalent to the new search API version.

Spell check will only be performed on the free text part of a query, so for a query like e.g. "ryda rummer contributor:agaust" we get "röda rummet" as suggestion.


olovy (approved):
LGTM

PR #1470, opened 2024-09-11
The 'agent' of a script-save should always be SEK (not whoever happen…

jannistsiroyannis:
…d to be the latest manual editor).


kwahlin (approved):
LGTM

PR #1471, opened 2024-09-12
Add scaffolding for running bulk changes

olovy:
Depends on https://github.com/libris/definitions/pull/496

All names are preliminary... Suggestions appreciated.

* Refactor WhelkTool
    * Add support for in-memory scripts
    * Don't do system.exit inside execution. So that WhelkTool can be embedded.
    * Move WhelkTool class so that directory structure matches package name
* Add BulkChange to whelktool
    * Executes a bulk change specified in a BulkChange record
* Add BulkChangeRunner to housekeeping
    * Scans for BulkChange records in status ready and starts them
* Add BulkChangePreviewApi to rest
    * Move some shared stuff to HttpTools


kwahlin (approved):
:ship:

PR #1473, opened 2024-09-13
Add script for re-saving records with broken generationProcess

andersju:
Some records got a broken generationProcess id -- `https://libris.kb.se/sys/globalchanges/me/devops/lddb-cleanups-PROD/librisxl/whelktool/./revert-13-uris.groovy` -- due to a whelktool script (not checked in here) being invoked as `./revert-13-uris.groovy` rather than `revert-13-uris.groovy`. Jena's IRI validator doesn't allow `/./` in an IRI (see [RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.3)), so saving these records through (at least) the REST API then didn't work.

(The fact that we enforce such validation when saving with through the REST API but not when saving with whelktool is itself problematic, but a separate issue)


kwahlin (approved):
:+1:

PR #1475, opened 2024-09-13
Simplify modification logic + more

kwahlin:
See commit. The whole `ModifiedThing.Modification` class was unnecessarily verbose. Fixed a couple of other minor things while at it. The next step is to add comprehensive test cases for the modification step to identify which cases it fails to handle.

### DIFF
--- a/whelktool/src/test/resources/whelk/datatool/form/modify-specs.json
+++ b/whelktool/src/test/resources/whelk/datatool/form/modify-specs.json
@@ -4,23 +4,161 @@
       "name": "Remove string",
       "matchForm": {
         "_id": "1",
-        "x1": "y1",
-        "x2": "y2"
+        "p1": "v1",
+        "p2": "v2"
       },
       "targetForm": {
         "_id": "1",
-        "x1": "y1"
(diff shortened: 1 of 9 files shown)


whelktool/src/test/resources/whelk/datatool/form/modify-specs.json
```diff
@@ -0,0 +1,164 @@
...
+        "p1": "v1",
+        "p2": {
+          "p1": "v2",
+          "p3": {
+            "@id": "https://id.kb.se/x"
+          }
+        }
+      },
+      "shouldFailWithException": true
+    },
```
olovy, line 125: If I actually wanted to replace p3, I should have done like this?
```
    "matchForm": {
        "_id": "1",
        "p1": "v1"
        "p2": {
          "p3": {
            "_id": "2"
          }
        }
      },
      "targetForm": {
        "_id": "1",
        "p1": "v1",
        "p2": {
          "p3": {
            "_id": "2"
            "@id": "https://id.kb.se/y"
          }
        }
      },
```

PR #1477, opened 2024-09-16
whelktool: validate JSON-LD by default

andersju:
Enforces JSON-LD validation by default (like in [Crud](https://github.com/libris/librisxl/blob/develop/rest/src/main/groovy/whelk/rest/api/Crud.groovy)) when updating/creating records. Can be skipped with `--no-validation`.


kwahlin (approved):
Nice!

olovy (approved):
LGTM!

PR #1478, opened 2024-09-16
Make it impossible to save holdings with bad heldBy-URIs.

jannistsiroyannis:


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -1002,6 +1013,18 @@ class PostgreSQLComponent {
...
+                String heldBy = doc.getHeldBy()
+                if (heldBy == null) {
+                    log.warn("Was asked to save a holding record lacking a correct library/heldBy (so, did nothing).")
+                    return null
+                }
+                String libraryRecordId = getRecordId(heldBy, connection)
+                if (libraryRecordId == null) {
+                    log.warn("Was asked to save a holding record for a library that could not be located: " + heldBy + " (so, did nothing).")
+                    return null
+                }
```
olovy, line 1026: Should it be possible to delete documents with bad heldBy links?
Now it is not possible without first changing heldBy to a valid value.

olovy (approved):
See comment.

Otherwise, LGTM

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
jannistsiroyannis: You are right! I'll fix it!

PR #1479, opened 2024-09-18
Interpret empty valued properties in matchForm as wildcards

kwahlin:
See added test cases. Up for discussion how remove/add to/replace **any** should work. If we think this logic could work I suggest we use an explicit wildcard marker rather than interpreting empty value as wildcard.


kwahlin:
See WIP branch https://github.com/libris/librisxl/tree/feature/bulk-change-optional-exact-match instead

PR #1481, opened 2024-09-20
Set maxItems on each Slice in sliceByDimension

kwahlin:
To facilitate https://github.com/libris/lxlviewer/pull/1106

The definition for https://id.kb.se/vocab/maxItems needs to be updated too.


olovy (approved):
LGTM!

I think we should refactor this at a later time. It is confusing that itemLimit is called size here.

PR #1483, opened 2024-09-24
Bulk change: Optional exact match and comprehensive unit tests for modification logic

kwahlin:
Mark any blank node to be exactly matched like this:
```
{
    "_match": "Exact",
    "x": "y"
}
```
It should also be possible to mark sets (arrays) to be exactly matched, for example in this way:
```
{
    "x": {
        "matchExactSet": [ {...}, {...} ]
    }
}
```
Support for exact set matching however is not yet implemented. Let's focus on more urgent/important stuff first and keep that as a todo for now.


olovy (approved):
LGTM!

PR #1485, opened 2024-09-30
fix(whelktool): Actually catch internal exceptions in processBatch

olovy:
Thread.setDefaultUncaughtExceptionHandler had no effect here. Any exceptions thrown by the runnable will get swallowed by the executorservice. I think the intent was make it easier to find any future unintended exceptions in processBatch. (Exceptions thrown by the script are already handled inside processBatch.)

We also don't want to set the global UncaughtExceptionHandler when running Whelktool embedded.

Instead wrap the Runnable and catch any exceptions.


andersju (approved):
LGTM!

PR #1487, opened 2024-10-01
Map all changes to respective node before mutating any data

kwahlin:
_Not_ doing so could cause unwanted changes, see for example the preview in https://libris-qa.kb.se/katalogisering/directory-care/bulkchanges/3m0zp6p91j2wv685 (role being added to the wrong Contribution).

See added test case too (this would fail before).

Also took opportunity to refactor and clean up a little bit.


olovy:
We need to add back tests for the ChangeSet addedPaths/removedPaths

PR #1490, opened 2024-10-03
Fix a race condition in the marc export code which resulted in double…

jannistsiroyannis:
… exports of the same record.

### DIFF
--- a/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
+++ b/marc_export/src/main/java/whelk/export/marc/ProfileExport.java
@@ -357,9 +357,11 @@ public class ProfileExport
 
         String profileName = profile.getProperty("name", "unknown");
         String systemId = document.getShortId();
-        if (exportedIDs.contains(systemId))
-            return;
-        exportedIDs.add(systemId);
+        synchronized (exportedIDs) {
+            if (exportedIDs.contains(systemId))
+                return;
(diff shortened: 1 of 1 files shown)


kaipoykio (approved):
Nice!

andersju (approved):
LGTM!

marc_export/src/main/java/whelk/export/marc/ProfileExport.java
```diff
@@ -357,9 +357,11 @@ private void exportDocument(Document document, ExportProfile profile, MarcRecord
...
         String profileName = profile.getProperty("name", "unknown");
         String systemId = document.getShortId();
-        if (exportedIDs.contains(systemId))
-            return;
-        exportedIDs.add(systemId);
+        synchronized (exportedIDs) {
+            if (exportedIDs.contains(systemId))
+                return;
+            exportedIDs.add(systemId);
+        }
```
olovy, line 364: I suggest keeping the concurrent set and:
```suggestion
        if (!exportedIDs.add(systemId)) {
            return;
        }
```

jannistsiroyannis: Hm. While also correct (ofc), I like the original one better. I think it's easier to read and understand. Yes, its subjective, I guess.

PR #1493, opened 2024-10-08
Bugfix/bulkchange form changesets

kwahlin:
Test case "Remove link: repeatable property, some values" would fail before due to values at `"matchForm"."r1".1` and `"targetForm"."r1".0` being treated as different entities.


olovy (approved):
LGTM

PR #1494, opened 2024-10-08
Feature/bulk change soft type matching

kwahlin:
I think this is all that's needed from the backend side regarding soft type matching?


olovy (approved):
:ship:

PR #1496, opened 2024-10-10
Schedule recurring scripts.

jannistsiroyannis:


kwahlin (approved):
LGTM! We need to revisit this when the type normalization is in place.

PR #1498, opened 2024-10-17
Feature/bulk change id list

kwahlin:
Handle id lists specified like this (with either of value/valueFrom):

```
{
    "@type": "FormSpecification"
    "matchForm": {
      "_id": "#1",
      "_idList": [
        {
          "@type": "AnyOf"
          "value": [ "asdcknjk3232", "lksdjhfglk343", ... ]
        },
        {
          "@type": "AnyOf"
          "valuesFrom": { "@id": "https://git.kb.se/..." }
        }
      ]
    }
}
```

The given ids apply to the node associated with `_idList`. In the above example it means that node `#1` should be either of the given ids.

The ids can be full URIs, fnurgl-ids or even old Voyager ids, however the latter only works if marc collection can be determined from the entity type.

Different id lists can be associated with different nodes in the graph and the selection of records is limited to only those with matching ids for each of these nodes. This is achieved by putting the ids for each node inside a `VALUES` clause in the Sparql query. (We have to keep an eye on the performance when using `VALUES`!)

We can remove/add to/replace a node associated with `_idList` just like with any other node.

---

This PR also includes support for unspecified type (`Any`).


whelktool/src/main/groovy/whelk/datatool/WhelkTool.groovy
```diff
@@ -150,24 +151,38 @@ class WhelkTool {
...
-        def idItems = shortIds.collect { "'$it'" }.join(',\n')
+        def idItems = idLoader.collectXlShortIds(ids).collect { "'$it'" }.join(',\n')
         if (idItems.isEmpty()) {
             return
         }
         doSelectBySqlWhere("id IN ($idItems) AND deleted = false", process,
                 batchSize)
     }
 
+    void selectByIdsAndCollection(Collection<String> ids, String collection, Closure process,
```
olovy, line 163: Do we really need this method? What is the use case?

whelktool/src/main/groovy/whelk/datatool/form/Transform.groovy
```diff
@@ -182,6 +302,133 @@ class Transform {
...
+
+                    def nodeType = node[TYPE_KEY]
+                    def marcCollection = nodeType ? getMarcCollectionInHierarchy((String) nodeType, whelk.jsonld) : null
+                    def xlShortIds = idLoader.collectXlShortIds(shortIds as List<String>, marcCollection)
+                    def parentProp = dropIndexes(path).reverse()[1]
+                    def isInRange = { type -> whelk.jsonld.getInRange(type).contains(parentProp) }
+                    // TODO: Fix hardcoding
+                    def isRecord = whelk.jsonld.isInstanceOf((Map) node, "AdminMetadata")
+                            || isInRange(RECORD_TYPE)
+                            || isInRange("AdminMetadata")
```
olovy, line 412: AdminMetadata is a platform term so I think it is fine.
But it could be defined in `JsonLd`

olovy (approved):
Very nice!
We should guard against SQL injection in the id list then I think it is good to go.
At least when we fetch them from a file. But I think also when using selectByIds in a script since that list is usually from some source..

whelktool/src/main/groovy/whelk/datatool/WhelkTool.groovy
olovy: Ok, so collection is needed if we want do do selection by voyager id since they are by collection

klngwll: Voyager-ids are only unique to their collection.

PR #1501, opened 2024-10-23
Feature/bulk change match base type

kwahlin:
Use  `_match: BaseType` to express that all subclasses of a type should be matched, e.g.:
```
"matchForm": {
        "_id": "#1",
        "@type": "Instance",
        "_match": ["BaseType"]
      }
```

I've reworked the code quite a bit once again, mainly to make it more efficient now that we end up with even more variants of the form to match (different ids, different types).

### DIFF
--- a/whelktool/src/main/groovy/whelk/datatool/form/Transform.groovy
+++ b/whelktool/src/main/groovy/whelk/datatool/form/Transform.groovy
@@ -30,41 +30,25 @@ class Transform {
     private static final String VALUE_FROM = 'valueFrom'
     private static final String ANY_TYPE = "Any"
     private static final String BASE_TYPE = "BaseType"
-
     private static final String BASE_TYPE_TMP_PROP = '_baseTypeTmp'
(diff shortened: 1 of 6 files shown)


whelktool/src/main/groovy/whelk/datatool/form/Transform.groovy
```diff
@@ -420,6 +298,30 @@ class Transform {
...
+                Set<String> subTypes = getSubtypes(baseType, jsonLd) as Set
+                mappings[baseType] = subTypes
+                return new DocumentUtil.Nop()
+            }
+        }
+
+        return mappings
+    }
+
+    @Memoized
```
olovy, line 320: This will give us trouble if we ever want to be able to refresh vocab without restarting whelk.
I think it would be better to reuse the Transform object within one bulk change, i.e. pass it to modify() inside the whelktool script.
But we can leave it as is for now.

olovy (approved):
:ship:

PR #1502, opened 2024-10-24
Feature/bulk delete

kwahlin:
Need to sort out the terminology.


kwahlin:
See https://github.com/libris/librisxl/pull/1506

PR #1504, opened 2024-10-29
Feature/bulk 1 to 1 changes

kwahlin:


olovy (approved):
LGTM!

Add test for changing some and deleting some duplicate nodes at the same time?

kwahlin:
> LGTM!
> 
> Add test for changing some and deleting some duplicate nodes at the same time?

Good suggestion! This actually revealed a couple of bugs (now fixed).

PR #1506, opened 2024-10-30
Implement bulk delete

kwahlin:
We probably want to refactor more but let's start with this and see how it works.


olovy (approved):
:ship:

PR #1507, opened 2024-10-30
Feature/bulk create

kwahlin:
I think this is all that's needed (in backend) to create records from a target form. The target form itself should suffice as preview for now.

The script expects exactly one id list attached in the target form and it must not be attached to the thing or record. This could maybe be restricted already in the interface. One record per unique and valid id in the list will be created.

For example
```
{
    "bulk:formBlankNodeId": "#1"
    "@type": "Item",
    "heldBy": {"@id": "https://libris.kb.se/library/SEK"},
    "hasComponent": [
        {
            "bulk:formBlankNodeId": "#2"
            "@type": "Item",
            "heldBy": {"@id": "https://libris.kb.se/library/SEK"}
        }
    ],
    "itemOf": { 
        "bulk:formBlankNodeId": "#3",
        "bulk:hasId": [
            {
                "bulk:formBlankNodeId": "#4"
                "@type": "bulk:AnyOf",
                "value": [
                    "gzrmdkks2zw7zs6",
                    "j1thr9kv2jbjk34",
                    "r93gg3c32rgcg2f"
                ]
            }
        ]
    }
}
```
would create three holding records with different links in itemOf.

### DIFF
--- a/whelktool/src/main/groovy/whelk/datatool/WhelkTool.groovy
+++ b/whelktool/src/main/groovy/whelk/datatool/WhelkTool.groovy
@@ -147,7 +147,6 @@ class WhelkTool {
         if (!silent) {
             log "Select by ${ids.size()} IDs"
         }
-
         def idItems = idLoader.collectXlShortIds(ids)
         if (idItems.isEmpty()) {
             return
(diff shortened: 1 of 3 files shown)


lrosenstrom (approved):
LGTM!

PR #1509, opened 2024-11-04
Feature/bulk merge

kwahlin:
I think we can merge this now. Should be fairly easy to just replace the part where we emulate the changes (https://github.com/libris/librisxl/blob/72e410ff654b2e79083e616f4b3b698d6a5fa98c/rest/src/main/groovy/whelk/rest/api/BulkChangePreviewAPI.java#L62) with the results of a dry-run once we have that in place.

TODO:
- Address this: https://github.com/libris/librisxl/blob/72e410ff654b2e79083e616f4b3b698d6a5fa98c/whelktool/src/main/resources/bulk-change-scripts/merge.groovy#L16 
I don't think we should replace links to record uris with links to thing uri so the question is what to do with them (if they ever appear)?
- Would be nice to make the frontend preview aware of and able to display different types of changes depending on the change sets content.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/BulkChangePreviewAPI.java
+++ b/rest/src/main/groovy/whelk/rest/api/BulkChangePreviewAPI.java
@@ -58,12 +58,7 @@ public class BulkChangePreviewAPI extends HttpServlet {
             var changeDoc = load(systemId);
 
             // TODO? let Specifications create their own previews? create preview in Whelktool instead?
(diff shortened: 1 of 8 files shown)


olovy (approved):
I think that there is now a bit of unclear division of responsibility between the different Specification classes and the corresponding whelktool scripts. But that will hopefully be fixed when we use recorded dry-runs for previews.

PR #1513, opened 2024-11-11
Add bulk script for removing topic subdivisions

olovy:
* Add a new generic bulk spec type `Other` that includes the name of the script to be run.
* Add a script for removing topic subdivisions


kwahlin (approved):
LGTM!

PR #1514, opened 2024-11-11
Bugfix/bulk deprecated

kwahlin:
Keep URIs of deleted resources in `bulk:changeSpec.bulk:deprecated`.

Had to hardcode the path in `LinkFinder.replaceSameAsLinksWithPrimaries` to avoid passing a parameter through long chains of methods. We really should refactor all the normalizer stuff sometime for better flexibility.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
+++ b/whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
@@ -115,8 +115,11 @@ class LinkFinder {
     }
 
     private void replaceSameAsLinksWithPrimaries(Map data, List path = []) {
+        def exceptedProperties = ['bulk:deprecate'] as Set
+        if (path && exceptedProperties.contains(path.last())) {
+            return
+        }
         // If this is a link (an object containing _only_ an id)
-
         String id = data.get("@id")
         if (id != null && data.keySet().size() == 1) {
             // Path to same form as in lddb__dependencies.relation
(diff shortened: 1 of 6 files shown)


olovy:
I'm not sure this is worth it. Is there any point of having a link to the deleted entity?
We could just define deprecated (deprecatedUri) as a string instead?

whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
```diff
@@ -115,8 +115,9 @@ class LinkFinder {
     }
 
     private void replaceSameAsLinksWithPrimaries(Map data, List path = []) {
-        // If this is a link (an object containing _only_ an id)
+        def exceptedPaths = ['bulk:changeSpec.bulk:deprecate'] as Set
```
olovy, line 118: I think this should be a constant with a more descriptive name and a small comment explaining what it does.

whelktool/src/main/resources/bulk-change-scripts/removeTopicSubdivision.groovy
```diff
@@ -14,41 +14,65 @@ import static whelk.JsonLd.ID_KEY
...
+                    var parent = DocumentUtil.getAtPath(thing, parentPath)
+                    if (parent instanceof List) {
+                        modifiedListPaths.add(parentPath)
                         // TODO? add way to do this with an op? SplitReplace? [Replace, Insert]?
-                        if (keepLink && path.size() > 1) {
-                            var parent = DocumentUtil.getAtPath(thing, path.dropRight(1))
-                            if (parent instanceof List && !parent.contains(keepLink)) {
-                                parent.add(keepLink)
-                            }
+                        if (addLink && addLink[ID_KEY]) {
```
olovy, line 40: Do we need to restrict it to links?
It could be any object?

PR #1515, opened 2024-11-12
Feature/bulk merge uris as strings

kwahlin:
See also https://github.com/libris/lxlviewer/commit/d63983188207e055c632d36c74e969185f5ad031 and https://github.com/libris/definitions/commit/da11d036b3ade71c51a33855983f85aed72489d4

Let's go with this for now? Instead of https://github.com/libris/librisxl/pull/1514.


olovy (approved):
:ship:

PR #1517, opened 2024-11-13
Serve bulk reports

olovy:
* Add `bulk:execution` with document counts and links to reports
* Serve bulk whelktool reports from housekeeping

Depends on https://github.com/libris/definitions/pull/504

### DIFF
--- a/whelktool/src/main/java/whelk/datatool/bulkchange/BulkJob.java
+++ b/whelktool/src/main/java/whelk/datatool/bulkchange/BulkJob.java
@@ -32,11 +32,13 @@ public class BulkJob implements Runnable {
     protected final String id;
     protected final String systemId;
     protected final Whelk whelk;
+    protected final String executionId;
 
     public BulkJob(Whelk whelk, String id) {
         this.whelk = whelk;
         this.id = id;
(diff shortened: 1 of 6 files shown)


whelktool/src/main/java/whelk/datatool/bulkchange/BulkJob.java
```diff
@@ -59,14 +80,65 @@ public void run() {
             printScriptLogHeader(tool, jobDoc);
             logger.info(String.format("Running %s: %s", JOB_TYPE, id));
 
+            storeUpdate(doc -> doc.setStatus(Completed));
```
kwahlin, line 83: Is this intentional?

kwahlin (approved):
Great job! :star: Looks good. Just a small thing that I suspect is an error, see comment.

whelktool/src/main/java/whelk/datatool/bulkchange/BulkJob.java
olovy: No!!

olovy, line 83:
```suggestion
```

PR #1520, opened 2024-11-18
More flexible subdivision job

kwahlin:
See especially 0a2990090896c701c7a477f47040843820a11ad9.

Depends on https://github.com/libris/lxlviewer/tree/feature/bulk-subdivision-template and https://github.com/libris/definitions/tree/feature/bulk-remove-subdivision.

Example 1:
image
image

Example 2:
image
image


olovy (approved):
Nice improvements to the subdivision job!

We can merge this as is. But I suggest the following changes. As discussed:
* No logic in `Specification`. Don't call into it from inside the scripts. They should be regular whelktool scripts that either implement their own logic or use (new) whelktool facilities.
* Keep selectByForm but make `MatchForm` a standalone class that can be used on its own (and within `Transform`). Let it have the responsibility of generating the SPARQL pattern.
* Remove the `skipPrelude` parameter to `toTurtle`. Use a separate method if needed. "prelude" or "prefixes"?

PR #1522, opened 2024-11-19
Feature/adjust subvision script

kwahlin:
Just need to test that all of this works before we can merge.


whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+        this.nodeIdToMatchFormPath = collectNodeIdToPath(form)
+        this.nodeIdMappings = collectNodeIdMappings(whelk)
+        this.baseTypeMappings = collectBaseTypeMappings(whelk?.jsonld)
+    }
+
+    MatchForm(Map matchForm) {
+        this(matchForm, null)
+    }
+
+    MatchForm() {}
```
olovy, line 55: It is not clear when MatchForm can be constructed without whelk.
It looks like it is only for testing?
I suggest making these two constructors `@PackageScope` with a comment "For testing only"

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+
+    Map<String, Set<String>> collectNodeIdMappings(Whelk whelk) {
+        return collectNodeIdMappings(form, whelk)
+    }
+
+    static Map<String, Set<String>> collectNodeIdMappings(Map form, Whelk whelk) {
```
olovy, line 61: I think a short comment would be helpful here.
Since the method name is not self explanatory.
What is the result? `Map<String, Set<String>>`

olovy: Or rather: descibe the member variables
```
    Map<String, List> nodeIdToMatchFormPath
    Map<String, Set<String>> nodeIdMappings
    Map<String, Set<String>> baseTypeMappings
```

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+    private static final String VALUE = 'value'
+    private static final String VALUE_FROM = 'bulk:valueFrom'
+    private static final String ANY_OF = 'bulk:AnyOf'
+    private static final String HAS_BASE_TYPE_TMP = '_hasBaseTypeTmp'
+
+    Map form
+
+    Map<String, List> nodeIdToMatchFormPath
+    Map<String, Set<String>> nodeIdMappings
+    Map<String, Set<String>> baseTypeMappings
```
olovy, line 42: I think all these need a short description.
What do they contain?

(In the long run we should try to move away from using strings everywhere. With proper types these could maybe be self explanatory)

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
olovy, line 61: Should be private?

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+    static Map<String, List> collectNodeIdToPath(Map form) {
+        Map<String, List> nodeIdToPath = [:]
+        DocumentUtil.findKey(form, BNODE_ID) { nodeId, path ->
+            nodeIdToPath[(String) nodeId] = path.dropRight(1)
+            return new DocumentUtil.Nop()
+        }
+        return nodeIdToPath
+    }
+
+    Map<String, Set<String>> collectBaseTypeMappings(JsonLd jsonLd) {
```
olovy, line 119: Should be private?

olovy, line 119:
```suggestion
    private Map<String, Set<String>> collectBaseTypeMappings(JsonLd jsonLd) {
```

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+        return valuesClauses + sparqlPattern
+    }
+
+    String getVar(String bNodeId) {
+        return bNodeId == getRecordTmpId()
+                ? "?$GRAPH_VAR"
+                : "?${bNodeId.replace('#', '')}"
+    }
+
+    @Memoized
```
olovy, line 223: We have to watch out here if we ever want to implement "hot reloading" of vocab.
Better to put `@Memoized` on `jsonLd.getSubClasses(type)`?

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
```diff
@@ -0,0 +1,307 @@
...
+
+    boolean matches(Object node) {
+        return matches(form, node)
+    }
+
+    boolean matches(Object matchForm, Object node) {
+        return comparator.isSubset(["x": matchForm], ["x": node], this::mapMatches)
+    }
+
+    boolean mapMatches(Map matchForm, Map bNode) {
```
olovy, line 244:
```suggestion
    private boolean mapMatches(Map matchForm, Map bNode) {
```

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
olovy, line 1: It think the methods should be reordered to make the class more readable.
```
collectNodeIdToPath(form)
collectNodeIdMappings(whelk)
collectBaseTypeMappings(whelk?.jsonld)
```
are only used for setup. Their implementation is not that interesting -> Move to the end of the file.

```
boolean matches(Object node) {
boolean matches(Object matchForm, Object node)
```
is the main purpose and public -> move to the top

followed by `getSparqlPattern`

whelk-core/src/main/groovy/whelk/converter/JsonLDTurtleConverter.groovy
```diff
@@ -18,18 +18,17 @@ class JsonLDTurtleConverter implements FormatConverter {
     }
 
     Map convert(Map source, String id) {
-        return [(JsonLd.NON_JSON_CONTENT_KEY) : _toTurtle(source, null, base, false)]
+        return [(JsonLd.NON_JSON_CONTENT_KEY) : toTurtle(source, null, base)]
     }
 
-    static String toTurtle(Map source, Map context, boolean skipPrelude) {
-        return _toTurtle(source, context, null, skipPrelude)
+    static String toTurtleData(source, Map context) {
```
olovy, line 24: Perhaps?
```suggestion
    static String toTurtleNoPrelude(source, Map context) {
```

olovy (approved):
Nice!
I think the structure is much improved. 

See comments for suggested improvements.

whelktool/src/main/groovy/whelk/datatool/form/MatchForm.groovy
kwahlin: What's the benefit? Wouldn't that mean we risk getting outdated vocab data in even more places?

olovy: I don't think so.
a JsonLd instance contains one "version" of vocab+context+display. 
For now it is never/updated replaced in one instance of whelk.
Keeping all the code that caches calculations made on vocab in one place reduces the risk of missing something if we ever change that.

kwahlin: Good point. I'll move `@Memoized` to the "source".

olovy: Ok!
Let's do it on develop though and not on the release branch.
Such a change in the core should have a bit more time for testing.

kwahlin: Was thinking the same :+1: 

9be79a03301f1974d125a784205f3272abadd9ae

PR #1527, opened 2024-11-21
Refactor MatchForm (reduce complexity)

kwahlin:
Couldn't resist cleaning up some more. Not necessary to include in the release.


olovy (approved):
:ship: 

Would be nice to refactor the tests so that MatchForm can use JsonLd::subTypes directly

PR #1528, opened 2024-11-22
Support replacing Subdivision with GenreForm

kwahlin:
One last addition! :crossed_fingers: Makes sense to be able to add GenreForm term when removing a GenreSubdivision. See example :broom: : 
image
image

Depends on: https://github.com/libris/definitions/pull/507
Will update the Subdivison template too.


olovy (approved):
:ship:

PR #1531, opened 2024-12-02
Feature/lws 88 boosting

kwahlin:
Boost same fields in Libris Search as in the cataloging.

- Code adapted from mainly [ESQueryLensBoost.groovy](https://github.com/libris/librisxl/blob/fef5958a9eb6798500d2f59f49748ab997323807/whelk-core/src/main/groovy/whelk/search/ESQueryLensBoost.groovy) and some parts from [ESQuery.groovy](https://github.com/libris/librisxl/blob/fef5958a9eb6798500d2f59f49748ab997323807/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy). Should produce exactly the same boosted fields.
- Set `_debug=esScore` to include scores + explanation in the API response. Works with both the "old" query style (using  `q` param) and the "new"  (`_q`). Thought it might facilitate working with improved relevance.
- Use `_boost` for custom boosting, e.g. `_boost=hasTitle.mainTitle^200,prefLabel^100...`.
- A bug was discovered in the existing code: When computing boosting from cards a field would not be boosted if another field containing the same nested subfield was already boosted. E.g. boosting `publication.year` would block `production.year` from being boosted. See fix: bdb3d0847be69881a58361dad22f19a801ee63dd.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -67,8 +67,15 @@ public class SearchUtils2 {
 
         Map<String, Object> partialCollectionView = getPartialCollectionView(queryRes, qTree, queryParams, appParams);
 
+        Map<String, Object> debugView = new HashMap<>();
         if (queryParams.debug.contains(QueryParams.Debug.ES_QUERY)) {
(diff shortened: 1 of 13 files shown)


jannistsiroyannis (approved):
I wont claim to understand much of it, but I saw nothing obviously problematic!

PR #1532, opened 2024-12-02
fix(rest): sameAs redirect when exists deleted doc with same id

olovy:
sameAs should redirect to the main id for the entity. 

This fixes the issue that a deleted (merged) doc with the same id would give 410 DELETED instead


andersju (approved):
LGTM!

PR #1533, opened 2024-12-02
Don't do nested selectBy in merge script

kwahlin:
Likely the reason why some updated records have not been saved properly in e.g. https://libris.kb.se/katalogisering/directory-care/bulkchanges/g1lb7dp4dfnj9s0k.


olovy (approved):
LGTM

PR #1535, opened 2024-12-10
Fix docs falsely showing as modified

kwahlin:
...due to the after version being modified before being compared with before version.


olovy (approved):
:ship:

PR #1537, opened 2024-12-10
fix(rest): sameAs redirect for merged docs

olovy:
sameAs should redirect to the main id for the entity.
This fixes the issue that a deleted (merged) doc with the same id would give 410 DELETED instead

TODO this could/should be implemented as a replacedBy inside the tombstone?


kwahlin (approved):
LGTM

PR #1538, opened 2024-12-11
Bugfix/delete referenced by bulk job

kwahlin:
https://libris.kb.se/katalogisering/8tgnxg2q6d26k657 failed due to the term to deprecate (https://id.kb.se/term/barn/%C3%85lderdomshem) being referenced in another `bulk:Job` which prevented the term from being deleted. We should not exclude all links in `bulk:Job` from relinking but rather exclude links in `bulk:depŕecate`specifically.

Also fixed the error message which I think was very confusing, see https://libris.kb.se/_bulk-change/reports/8tgnxg2q6d26k657-2024-12-11T133742/ERRORS.txt. Only `k4rz49t1h5nfxls5 at bulk:changeSpec.bulk:matchForm.instanceOf.subject` is of interest, the rest are just records linking to k4rz49t1h5nfxls5 (directly or indirectly). This information is already available via "Används i" in the cat client anyway.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -2773,8 +2773,7 @@ class PostgreSQLComponent {
                 def allow = JsonLd.ALLOW_LINK_TO_DELETED + (jsonld?.cascadingDeleteRelations() ?: Collections.EMPTY_SET)
                 def referencedBy = followDependers(identifier, allow)
                 if (!referencedBy.isEmpty()) {
-                    def referencedByStr = referencedBy.collect { shortId, path -> "$shortId at $path" }.join(', ')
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2773,8 +2773,7 @@ class PostgreSQLComponent {
                 def allow = JsonLd.ALLOW_LINK_TO_DELETED + (jsonld?.cascadingDeleteRelations() ?: Collections.EMPTY_SET)
                 def referencedBy = followDependers(identifier, allow)
```
olovy, line 2774: Note to selves: the actual problem is that we are using followDependers though we only care about docs linking directly to the doc being deleted? But there are only links when we block the delete so probably fine.

Changed in https://github.com/libris/librisxl/commit/230fe6f4f515cb11b1b2619977eeee5a84329b97

olovy (approved):

PR #1540, opened 2024-12-17
Include score by field for each item in result

kwahlin:
Cherry-pick the most interesting stuff (field + score) from Elastic's explain API response and include in each result item when using `_debug=esScore`. Example:
image


olovy (approved):
Nice! :ship: 

Maybe it is more convenient to also have `_explain` inside `items._debug`?

kwahlin:
> Nice! 🚢
> 
> Maybe it is more convenient to also have `_explain` inside `items._debug`?

Indeed! [ac32795](https://github.com/libris/librisxl/pull/1540/commits/ac3279553e68e757a124780b2e4f114f7d4c9da3)

PR #1541, opened 2024-12-18
Don't throw exception when encountering invalid query terms

kwahlin:
Proceed with the query in the usual way (we won't get any hits though), however mark the invalid parts in `search.mapping` like this:
```
{
    "@type": "_Invalid",
    "label": "theInvalidInputString"
}
```
This can apply to both keys and values. So if we type for example `type:Kalle x:y subject:horses` we get
```
"search": {
    "mapping": [
      {
        "and": [
          {
            "property": {
              "@id": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
              "@type": "ObjectProperty",
              "labelByLang": {
                "en": "type",
                "sv": "typ"
              }
            },
            "equals": {
              "@type": "_Invalid",
              "label": "Kalle"
            },
            "up": {
              "@id": "/find?_i=&_q=x:y+subject:horses&_limit=200"
            }
          },
          {
            property": {
              "@type": "_Invalid",
              "label": "x"
            },
            "equals": "y",
            "up": {
              "@id": "/find?_i=&_q=%22rdf:type%22:Kalle+subject:horses&_limit=200"
            }
          },
          {
            "property": {
              "@id": "https://id.kb.se/vocab/subject",
              "@type": "ObjectProperty",
               ...
            },
            "equals": "horses",
            "up": {
              "@id": "/find?_i=&_q=%22rdf:type%22:Kalle+x:y&_limit=200"
            }
          }
        ],
        "up": {
          "@id": "/find?_i=&_q=*&_limit=200"
        }
      }
    ]
  }
```
since `x` is an invalid key and `Kalle` is and invalid value for the property `rdf:type`.

Eventually we may need more details about why a certain string is invalid and, consequently, also a more expressive model. Let's see. Stick with the temporary `_Invalid` type for now.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/querytree/InvalidKey.java
+++ b/whelk-core/src/main/groovy/whelk/search2/querytree/InvalidKey.java
@@ -1,5 +1,6 @@
 package whelk.search2.querytree;
 
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import static whelk.JsonLd.TYPE_KEY;
@@ -11,6 +12,9 @@ public sealed interface InvalidKey {
     String key();
 
     default Map<String, Object> getDefinition() {
-        return Map.of(TYPE_KEY, "_Invalid", "label", key());
+        var m = new LinkedHashMap<String, Object>();
+        m.put(TYPE_KEY, "_Invalid");
+        m.put("label", key());
+        return m;
     }
 }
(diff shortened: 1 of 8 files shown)


olovy (approved):
LGTM!

Perhaps InvalidKey could be aligned with Property so that the don't have to be treated separately in this part of the code?
```
interface Foo { ... }

public class Property implements Foo { ... }

public sealed interface InvalidKey extends Foo { ... }
```

olovy:
> LGTM!
> 
> Perhaps InvalidKey could be aligned with Property so that the don't have to be treated separately in this part of the code?
> 
> i.e.
> 
> ```
> interface Foo { 
>     String name();
>     Map<String, Object> definition();
> }
> 
> public class Property implements Foo { ... }
> 
> public sealed interface InvalidKey extends Foo { ... }
> ```

for example https://github.com/libris/librisxl/pull/1543

PR #1542, opened 2024-12-18
Add functionality for sorting facets/observations into a tree

lrosenstrom:
Maybe we can try this out on dev2 to see how it works in practice.

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -27,10 +27,12 @@ import static whelk.util.Jackson.mapper;
 public class SearchUtils2 {
     private final QueryUtil queryUtil;
     private final Disambiguate disambiguate;
+    private final JsonLd jsonLd;
 
     SearchUtils2(Whelk whelk) {
         this.queryUtil = new QueryUtil(whelk);
         this.disambiguate = new Disambiguate(whelk);
+        this.jsonLd = whelk.getJsonld();
     }
 
     Map<String, Object> doSearch(Map<String, String[]> queryParameters) throws InvalidQueryException, IOException {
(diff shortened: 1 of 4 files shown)


lrosenstrom:
This might be a too naive approach to how this should work? Hopefully it can be a start...

rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
```diff
@@ -27,10 +27,12 @@
 public class SearchUtils2 {
     private final QueryUtil queryUtil;
     private final Disambiguate disambiguate;
+    private final JsonLd jsonLd;
 
     SearchUtils2(Whelk whelk) {
         this.queryUtil = new QueryUtil(whelk);
         this.disambiguate = new Disambiguate(whelk);
+        this.jsonLd = whelk.getJsonld();
```
kwahlin, line 35: No need to pass on the JsonLd object, use `Disambiguate.isSubclassOf()` instead.

whelk-core/src/main/groovy/whelk/search2/Stats.java
```diff
@@ -123,6 +127,9 @@ private Map<String, Object> buildSliceByDimension(Map<Property, Map<PropertyValu
             var sliceNode = new LinkedHashMap<>();
             var isRange = rangeProps.contains(property);
             var observations = getObservations(buckets, isRange ? queryTree.removeTopLevelPropValueWithRangeIfPropEquals(property) : queryTree, nonQueryParams);
+            if (property.name().equals(Disambiguate.RDF_TYPE)) {
```
kwahlin, line 130: Use `property.isType()` to catch both work and instance types (`hasInstanceType`/`instanceOfType` in addition to `rdf:type`).

kwahlin (changes requested):
See comments, otherwise LGTM! Let's try it out on e.g. dev.

kwahlin (approved):

whelk-core/src/main/groovy/whelk/search2/FacetTree.java
```diff
@@ -0,0 +1,66 @@
...
+                queue.add(observation);
+            }
+        });
+
+        while (!queue.isEmpty()) {
+            var observation = queue.remove();
+            var children = findChildren(observation, observations);
+            if (!children.isEmpty()) {
+                queue.addAll(children);
+                observation.put("children", children);
```
kwahlin, line 38:
```suggestion
                observation.put("_children", children);
```
Since `children` is not a defined property

PR #1544, opened 2024-12-19
Feature/vocab guided query expansion

kwahlin:
Depends on https://github.com/libris/definitions/pull/508 and https://github.com/libris/lxlviewer/pull/1194.

The main objective with this is to get rid of obviously bad hardcodings and make the whole query expansion machinery more consistent. Apart from some general cleanup/refactoring the main improvements are:

- Finding the correct full path(s) for a property does no longer depend on [OutsetType](https://github.com/libris/librisxl/blob/bb09334681e8304f5733f3b25339d893f177db2b/whelk-core/src/main/groovy/whelk/search2/OutsetType.java) or [DomainCategory](https://github.com/libris/librisxl/blob/bb09334681e8304f5733f3b25339d893f177db2b/whelk-core/src/main/groovy/whelk/search2/querytree/Property.java#L16). Instead we check for properties in the category `:integral` that are applicable to the queried type(s). 
Example query: `type:Work yearPublished:x`
There are three `:integral` properties (`:translationOf`, `:exactMatch`, `:hasInstance`) applicable to `:Work` which, if we didn't know the domain of `:yearPublished`, would give us these alternative paths for `:yearPublished`:
     -  `translationOf.yearPublished`
     -  `exactMatch.yearPublished`
     -  `hasInstance.yearPublished`
We can however ignore both `:translationOf` and `:exactMatch` since we know that `:yearPublished :domain :Instance` and `:Instance` is not in range of `:translationOf` and `:exactMatch`, i.e. we won't find `:yearPublished` on any resource that these properties point to. So the only path that will be checked is `hasInstance.yearPublished` (`@reverse.instanceOf.publication.year` when fully expanded).
- Both property expansion and field boosting depend on which type of resource is queried, however only the types within the same group are relevant. Before we just collected all types that appeared _anywhere_ in the query tree. This change enables grouped queries with different types while still getting the property expansion and free text boosting right. E.g.
`(type:Work x) OR (type:Instance p1:v1) OR (type:Agent p2:v2)`
where field boosting for `x` is based on the `:Work` type and expansion of `p1` and `p2` is based on `:Instance` and `:Agent` respectively.
- Redundant types within the same group are removed: 
https://github.com/libris/librisxl/blob/8d3fa034902e69b0f2213df01cc8f3d3e3d3259d/whelk-core/src/main/groovy/whelk/search2/querytree/Group.java#L110 
If for example the query is `type:(Electronic Instance)` then `Instance` is removed. Just like in the old search:
https://github.com/libris/librisxl/blob/bb09334681e8304f5733f3b25339d893f177db2b/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy#L440
- Rely on marker `:category :shorthand` in definition for knowing which properties should be expanded via `owl:propertyChainAxiom`. a16fcb3450fa36b18d0011f6d3075f550bdce11b
- Don't inherit domain/range from propertyChainAxiom. Add these explicitly to short forms where needed instead, see https://github.com/libris/definitions/pull/508.
- Search all subclasses by default for all types: https://github.com/libris/librisxl/blob/8d3fa034902e69b0f2213df01cc8f3d3e3d3259d/whelk-core/src/main/groovy/whelk/search2/querytree/PropertyValue.java#L76
We only did this for Work and Instance before (hardcoded).
Eventually we'll need to make this optional.
- Using alias for instance/work type was such a bad and confusing solution that I decided to just bin it (d69a3b2d1cfa187a0d127eda70dc6d624b7823eb). Mapping `hasInstanceType` to `Format` (https://github.com/libris/lxlviewer/pull/1194) is enough to get the filter headers right at least when searching for works which is most important at this point. We can figure something out for instance search later.
- The whole `Disambiguate` class was also a mess so I decided to give it a proper makeover. At least is less messy now.

Can't guarantee that this won't break anything. Planned to add more tests but ran out of time. _Seems_ to work when I've tested myself.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
+++ b/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
@@ -338,11 +338,6 @@ public class Disambiguate {
         getAsListOfMaps(propertyDefinition, "equivalentProperty")
                 .forEach(ep -> getDefinition(ep, whelk).ifPresent(inheritable::add));
 
-        getAsOptionalListOfMaps(propertyDefinition, "propertyChainAxiom")
-                .map(List::getFirst)
-                .flatMap(firstInChain -> getDefinition(firstInChain, whelk))
-                .ifPresent(inheritable::add);
-
         getAsListOfMaps(propertyDefinition, Rdfs.SUBPROPERTY_OF)
                 .forEach(superProp -> getDefinition(superProp, whelk).ifPresent(inheritable::add));
 
--- a/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
+++ b/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
@@ -12,6 +12,7 @@ import whelk.search2.querytree.Property;
 import whelk.search2.querytree.VocabTerm;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
@@ -71,8 +72,8 @@ public class Disambiguate {
         this.jsonLd = whelk.getJsonld();
         this.vocab = jsonLd.vocabIndex;
         this.integralRelations = jsonLd.getCategoryMembers("integral");
-        this.domainByProperty = new HashMap<>();
-        this.rangeByProperty = new HashMap<>();
+        this.domainByProperty = new ConcurrentHashMap<>();
+        this.rangeByProperty = new ConcurrentHashMap<>();
         setAliasMappings();
         // FIXME: This should probably not be a static variable...
         if (freeTextDefinition.isEmpty()) {
--- a/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
+++ b/whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
@@ -18,6 +18,7 @@ import java.util.stream.Stream;
 
 import static whelk.JsonLd.ID_KEY;
 import static whelk.JsonLd.TYPE_KEY;
+import static whelk.JsonLd.VOCAB_KEY;
 import static whelk.JsonLd.asList;
 import static whelk.search2.QueryUtil.loadThing;
 import static whelk.util.DocumentUtil.getAtPath;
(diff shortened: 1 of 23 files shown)


whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
```diff
@@ -33,42 +39,49 @@ public class Disambiguate {
...
     public Disambiguate(Whelk whelk) {
         this.whelk = whelk;
         this.jsonLd = whelk.getJsonld();
         this.vocab = jsonLd.vocabIndex;
-        this.domainByProperty = loadDomainByProperty(whelk);
-        setAliasMappings(whelk);
-        setTypeSets(jsonLd);
+        this.integralRelations = jsonLd.getCategoryMembers("integral");
+        this.domainByProperty = new HashMap<>();
+        this.rangeByProperty = new HashMap<>();
```
olovy, line 74: These can't be regular HashMaps since multiple requests/threads add to them.

whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
```diff
@@ -95,379 +108,91 @@ public Set<String> getAmbiguousEnumMapping(String alias) {
...
+        var fromVocab = get(vocab, jsonLd.toTermKey(iri), Map.of());
+        return fromVocab.isEmpty() ? loadThing(iri, whelk) : Optional.of(fromVocab);
     }
 
     private static boolean isKbvTerm(Map<?, ?> termDefinition) {
-        return getAsOptionalMap(termDefinition, "isDefinedBy")
-                .flatMap(Disambiguate::getLinkIri)
-                .filter("https://id.kb.se/vocab/"::equals)
-                .isPresent();
+        return "https://id.kb.se/vocab/".equals(get(termDefinition, List.of(Rdfs.IS_DEFINED_BY, ID_KEY), ""));
```
olovy, line 181: Get from `@vocab` in system context instead?
("isSystemVocabTerm")

olovy (approved):
Great work!

See comment about thread safety. Otherwise this is good to go.

Let's profile this and see where efficiency can be improved (perhaps in loading definitions etc).

whelk-core/src/main/groovy/whelk/search2/Disambiguate.java
kwahlin: Good idea

PR #1548, opened 2025-01-20
fix(bulk):Identify hold records via old style instance ids

kwahlin:
See https://github.com/libris/librisxl/pull/1547


olovy (approved):
LGTM

PR #1549, opened 2025-01-20
Generate _str recursively (frame before)

kwahlin:
If a **token** lens is defined for the entity type, get properties to include in `_str` from this, otherwise get them from the **chip** lens.

Some adjustments in the tokens definitions had to be done in order to produce appropriate strings for `Contribution` and `ComplexSubject`. See https://github.com/libris/definitions/pull/510.


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -501,7 +468,56 @@ class ElasticSearch {
         })
     }
 
-    private static void setComputedProperties(Document doc, Set<String> links, Whelk whelk) {
+    private static void setComputedProperties(Document doc, Map framed, Set<String> links, Whelk whelk) {
```
olovy, line 471: Now that this method takes both framed and doc (copy) it is not on different abstraction level than the caller. I think it should be inlined in getShapeForIndex

olovy (approved):

PR #1550, opened 2025-01-23
Add more batch import logging.

jannistsiroyannis:


olovy (approved):
LGTM

Can't tell if this is in a a format that makes it easy filter out records that need manual attention. But it follows the existing pattern.

PR #1552, opened 2025-02-10
Feature/lxl 4610 allow dead links in bulk change spec

kwahlin:
To be able to delete records referenced in a bulk change specification, allow any property inside a `bulk:ChangeSpec` (i.e. the object of `bulk:changeSpec` to hold links to deleted resources. We already have a similar mechanism for exact paths, see https://github.com/libris/librisxl/blob/a56f059e2780e3863dfd6e35139fc2cd371d11e8/whelk-core/src/main/groovy/whelk/JsonLd.groovy#L59).
This should not however be hardcoded but rather defined in the vocab (TODO).

See https://kbse.atlassian.net/browse/LXL-4610 for more info.

I sneaked in another commit too (a56f059) that is unrelated to the above. When the data in a record is not compatible with the operations specified in a `bulk:ChangeSpec` of a `bulk:Update`, the record is just ignored and won't be affected by the job. This can happen for example when trying to add another value to a property that is not repeatable.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -27,6 +27,7 @@ import java.time.Instant
 import java.time.ZoneId
 
 import static whelk.FeatureFlags.Flag.INDEX_BLANK_WORKS
+import static whelk.exception.LinkValidationException.IncomingLinksException
 
 /**
  * The Whelk is the root component of the XL system.
@@ -556,7 +557,10 @@ class Whelk {
             log.warn "Could not remove object from whelk. No entry with id $id found"
         }
         if (doc) {
-            storage.remove(id, changedIn, changedBy, force)
+            if (!force) {
+                assertNoDependers(doc)
+            }
+            storage.remove(id, changedIn, changedBy)
             indexAsyncOrSync {
                 elastic.remove(id)
                 if (features.isEnabled(INDEX_BLANK_WORKS)) {
@@ -566,6 +570,14 @@ class Whelk {
                     reindexAffected(doc, doc.getExternalRefs(), Collections.emptySet())
                 }
             }
+        }
+    }
+
+    private void assertNoDependers(Document doc) {
+        boolean isDependedUpon = storage.getIncomingLinkCountByIdAndRelation(doc.getShortId())
+                .any { relation, _ -> !JsonLd.isWeak(relation) }
+        if (isDependedUpon) {
+            throw new IncomingLinksException("Record is referenced by other records")
         }
     }
 
--- a/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
@@ -18,7 +18,6 @@ import whelk.IdType
 import whelk.JsonLd
 import whelk.Link
 import whelk.exception.CancelUpdateException
-import whelk.exception.LinkValidationException
 import whelk.exception.MissingMainIriException
 import whelk.exception.StaleUpdateException
 import whelk.exception.StorageCreateFailedException
@@ -47,7 +46,9 @@ import java.util.regex.Pattern
 
 import static groovy.transform.TypeCheckingMode.SKIP
 import static java.sql.Types.OTHER
+import static whelk.JsonLd.Owl.SAME_AS
 import static whelk.util.Jackson.mapper
+import static whelk.exception.LinkValidationException.OutgoingLinksException
 
 /**
  *  It is important to not grab more than one connection per request/thread to avoid connection related deadlocks.
@@ -457,7 +458,7 @@ class PostgreSQLComponent {
     private static final String DELETE_USER_DATA =
             "DELETE FROM lddb__user_data WHERE id = ?"
 
-    private static final String GET_IRI_IS_LINKABLE = """
+    private static final String GET_IRI_IS_DELETED = """
             SELECT lddb.deleted
             FROM lddb__identifiers
             JOIN lddb ON lddb__identifiers.id = lddb.id WHERE lddb__identifiers.iri = ?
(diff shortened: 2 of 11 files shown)


jannistsiroyannis (changes requested):
Please don't (even further) weaken these basic data integrity invariants. Things that are linked to should not be deleted. No exceptions. Full stop.

Links from jobs to affected records can be unqualified instead. If needed for the frontend to draw it correctly, please make the unqualified links into temporary qualified ones in memory *there*, not here, where it compromises these checks.

whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -547,6 +551,14 @@ class Whelk {
         }
     }
 
+    private void assertNoDependers(Document doc) {
+        boolean isDependedUpon = storage.getIncomingLinkCountByIdAndRelation(doc.getShortId())
+                .any { relation, _ -> !JsonLd.isWeak(relation) }
+        if (isDependedUpon) {
+            throw new IncomingLinksException("Document is referenced by other records")
```
olovy, line 558:
```suggestion
            throw new IncomingLinksException("Record is referenced by other records")
```

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -775,8 +776,7 @@ class PostgreSQLComponent {
                         throw new ConflictingHoldException("Already exists a holding record for ${heldBy} and bib: $holdingFor")
                 }
 
-                if (linkFinder != null)
-                    linkFinder.normalizeIdentifiers(doc)
```
olovy, line 779: Why has this call been removed?

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -1010,7 +1010,11 @@ class PostgreSQLComponent {
             }
             
             boolean deleted = doc.getDeleted()
-            
+
+            if (!deleted) {
+                assertNoLinksToDeleted(doc.getExternalRefs() - preUpdateDoc.getExternalRefs())
```
olovy, line 1015:
```suggestion
                var addedLinks = doc.getExternalRefs() - preUpdateDoc.getExternalRefs()
                assertNoLinksToDeleted(addedLinks)
```

whelktool/src/main/groovy/whelk/datatool/form/ModifiedThing.groovy
olovy, line 1: This stuff isn't really related to the rest of the PR?

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
kwahlin: This call is redundant, see https://github.com/libris/librisxl/blob/23315155947f69969f6f352f678071421b0a461b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy#L741 further up.

olovy: Ok, see https://github.com/libris/librisxl/pull/1552/commits/cfa88469d6615a9a9666fbef9deabcfdf0d0603c

`normalizeIdentifiers` is already called `normalizeDocumentForStorage`
https://github.com/libris/librisxl/blob/23315155947f69969f6f352f678071421b0a461b/whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy#L741

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2094,19 +2096,18 @@ class PostgreSQLComponent {
...
-    boolean iriIsLinkable(String iri, String path) {
-        if (path in JsonLd.ALLOW_LINK_TO_DELETED) {
-            return true
-        }
+    boolean isDeleted(String iri) {
         withDbConnection {
             PreparedStatement preparedStatement = null
             ResultSet rs = null
             try {
                 preparedStatement = getMyConnection().prepareStatement(GET_IRI_IS_LINKABLE)
```
olovy, line 2104: `GET_IRI_IS_LINKABLE` should be renamed

whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2976,7 +2969,16 @@ class PostgreSQLComponent {
             }
         }
     }
-    
+
+    private static void assertNoLinksToDeleted(Set<Link> links) {
+        links.each {link ->
+            // Except sameAs since it's sometimes necessary to add sameAs pointing at already deleted records
```
olovy, line 2975:
```suggestion
            // sameAs is allowed because when merging two entities, the id of the deleted entity is added to sameAs of the remaining entity
```

whelk-core/src/main/groovy/whelk/filter/LinkFinder.groovy
```diff
@@ -112,72 +109,51 @@ class LinkFinder {
...
+    private void replaceSameAsLinksWithPrimaries(data, List<String> propertyPath) {
+        if (data instanceof List) {
+            for (def element : data) {
+                replaceSameAsLinksWithPrimaries(data, propertyPath)
+            }
         }
-    }
+        if (data instanceof Map) {
+            // Keep original ID for "weak relations"
+            if (isLink(data) && !JsonLd.isWeak(propertyPath.join("."))) {
```
olovy, line 137: Will not return here if it is weak link. 
i.e. the loop below will be run for weak links

olovy (changes requested):
See comments about HTTP status code and `replaceSameAsLinksWithPrimaries`

Other than that, LGTM!

olovy (approved):
LGTM!

PR #1553, opened 2025-02-11
Include input values in search mapping + refactor search2

kwahlin:
The actual feature can be found here: https://github.com/libris/librisxl/blob/15a2585024fb0ad25f5540b3b26e8acd1ce79d99/whelk-core/src/main/groovy/whelk/search2/querytree/PathValue.java#L162

The rest is just refactoring, see commit message: 15a2585


whelk-core/src/main/groovy/whelk/JsonLd.groovy
```diff
@@ -642,6 +663,12 @@ class JsonLd {
 
     //==== Class-hierarchies ====
 
+    Set<String> getSuperClasses(String type) {
```
olovy, line 666: In general more useful to have this as an ordered List?
Convert to Set at use instead?

olovy (approved):
LGTM!

Agree the refactoring improves the structure.

PR #1554, opened 2025-02-11
fix(bulk): Skip Sparql when an id list was given for the top level node

kwahlin:
This means no limit on number of records per job when using id list. We could set a number but I think the id list itself should be enough limitation. 

When uploading id list for a specific property, e.g. itemOf, we'll still use Sparql so those lists should be limited to a few thousand ids. We probably want to discard too long lists and inform the user automatically but I guess at this point it's enough that they are aware of limitation.

https://kbse.atlassian.net/browse/LXL-4620


poacherone:
Having the limitation in properties (specifically itemOf) will continue to be a problem, see comment on the issue in Jira.

PR #1555, opened 2025-02-13
Add basic support for Fresnel Format vocabulary

olovy:
Add basic support for the [format part of the Fresnel vocabulary](https://www.w3.org/2005/04/fresnel-info/manual/#stylecore). 

The immediate goal is "arbitrary data" -> "plain chip string with punctuation/separators". 
```
FresnelUtil f = ...
Decorated d = f.format(f.applyLens(data, lens), locale)
String s = d.asString()
```
`Decorated` is the same data structure that is used in lxl-web for, aka "decorated data".
It can also be used as a base for rendering richer formats, e.g. HTML.
(There are some differences to how transliterated strings are represented. This should be aligned when we are happy with it.)

This can be used for formatting data in e.g.
- `computedLabel` in API responses
- Record summaries in CXZ emails
- `_str` and `sortKeyByLang` in elastic index

FresnelUtil is basically as transcription of [xl.ts in lxl-web](https://github.com/libris/lxlviewer/blob/develop/lxl-web/src/lib/utils/xl.ts) but became more object oriented.

The intermediate structure after applying lenses "`Lensed`" should probably be refactored. 
I will probably split FrenelUtil over multiple files.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/JsonLd.groovy
+++ b/whelk-core/src/main/groovy/whelk/JsonLd.groovy
@@ -30,6 +30,7 @@ class JsonLd {
     public static final String REVERSE_KEY = "@reverse"
     // JSON-LD 1.1
     public static final String PREFIX_KEY = "@prefix"
+    public static final String NONE_KEY = "@none"
 
     public static final String DISPLAY_KEY = "dataDisplay"
     public static final String THING_KEY = "mainEntity"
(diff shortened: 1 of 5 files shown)


kwahlin (approved):
Looks like a good start! I haven't dug into the details but seems fairly easy to do so when needed thanks to informative classes, comments and unit tests.

Splitting up `FresnelUtil` sounds like a good idea. Some methods can surely be generalized and reused in more places too.

PR #1556, opened 2025-02-18
Add the --ignoreHistory flag to the batch import.

jannistsiroyannis:
This is not to beused in any automatic flows, but for special operations it is sometimes usefull to be able to override/ignore the existing record history, to merge/replace some incoming data that you know is good.


olovy (approved):
LGTM

PR #1557, opened 2025-02-18
Always load selection directly from Postgres when given ID list in bulk change specification

kwahlin:
Follow-up on https://github.com/libris/librisxl/pull/1554 as per discussion in https://kbse.atlassian.net/browse/LXL-4620. Works with multiple ID lists at different paths, see for example:
bild


olovy (approved):
LGTM!

PR #1560, opened 2025-02-21
Make it possible to sort works on number of holdings

olovy:
Make it possible to sort works on number of holdings.
For now, including this count is hardcoded.

Also fix a bug where adding an item (holding) to an instance wouldn't reindex the instance's linked work.
Even though items are now included with the work in the index. (So that works can be faceted on holdings)


olovy:
Maybe should count unique `heldBy`

kwahlin (approved):
Obviously needs some refinement later on, but looks to do the job for now!

PR #1561, opened 2025-02-21
Feature/lxl 4627

jannistsiroyannis:
This adds a way to import data (not history) on some subset (types) from one XL to another, using EMM.


olovy (approved):
LGTM!

PR #1564, opened 2025-02-27
Bugfix/alternate props boost

kwahlin:
For queries like `_q=abc type:T` or `q=abc&@type=T` "language container" properties appearing in `showProperties.alternateProperties` of `T`'s lens definition weren't boosted. The problem was that when looping through `showProperties.alternateProperties` property `p` has already been expanded to an array: `[ "p", "pByLang" ]` and this array was ignored. So for e.g. `Language`, having the following chips definition
```
"Language": {
          "@id": "Language-chips",
          "@type": "fresnel:Lens",
          "classLensDomain": "Language",
          "showProperties": [
            {"alternateProperties": ["prefLabel", "label", "code"]}
          ]
        }
``` 
only `code` was boosted since `alternateProperties` in this case is expanded to
 ```
{
    [ "prefLabel", "prefLabelByLang" ],
    [ "label", "labelByLang" ],
    "code"
}
```

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search/ESQueryLensBoost.groovy
+++ b/whelk-core/src/main/groovy/whelk/search/ESQueryLensBoost.groovy
@@ -2,6 +2,8 @@ package whelk.search
 
 import whelk.JsonLd
 
+import static whelk.JsonLd.asList
+
 /**
  * A utility to compute a list of boosted fields for ElasticSearch.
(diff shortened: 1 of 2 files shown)


olovy (approved):
This works.

I would prefer doing an explicit check for prop instanceof List and spell out that it is handling expanded language aliases

PR #1569, opened 2025-03-05
wip/experimental(elasticsearch): Use secondary index for Item and ImageObject

andersju:
So as to not burden the main ES index with a massive amount of Item and ImageObject records (that we usually don't care much about when searching) we've talked about using separate indices. Here's a very simple way of doing that. Not saying it's necessarily a _good_ way, but just to open up the discussion and get more ideas. It Seems To Work (tm).

Essentially this just splits ES into a "main" index and a "secondary" index, for lack of a better name, with the latter having only Item and ImageObject records. Both are used by default when searching, so no change from the outside. `_searchMainOnly=true` can be added to _not_ search the secondary index. Index to use when adding/updating records is determined based on type.

To try it locally see https://github.com/libris/devops/tree/feature/use-secondary-es-index for some quick and dirty devops support.


kwahlin (approved):
Nice! If it seems to work let's try it out in a test environment.

PR #1572, opened 2025-03-13
feat: Remove gui-whelktool

olovy:
Replaced by BulkJob (MÄV)


andersju (approved):
:broom: :partying_face:

PR #1573, opened 2025-03-17
Feature/fmt 313 change textinstance to print

branchedelac:
Hej!

Den här branchen uppdaterar marcframe.json så att MARC `007[0] = t` ger instans med `"@type": "Print"` istället för det tidigare `"TextInstance"`.


whelk-core/src/main/resources/ext/marcframe.json
```diff
@@ -2399,6 +2399,7 @@
           "addLink": "genreForm",
           "uriTemplate": "https://id.kb.se/marc/BooksContentsType-{_}",
           "matchUriToken": "^[abcdefgijklmnopqrstuvwyz256]$",
+          "NOTE:lcMarc2bfCodeMap": "marcgt",
```
niklasl, line 2402: Should these go into the typenormalization branch (are they notes on suggested mapping changes)?

branchedelac: Good catch! I will remove them.

niklasl (approved):
Change looks good! We need to ensure that nothing has type `TextInstance` when this is released.

PR #1574, opened 2025-03-17
fix(bulk): Load IDs from list only when necessary

kwahlin:
When generating `changeSets` we are only interested in the diff between `matchForm` and `targetForm` and thus we don't need to bother with loading ID lists possibly contained in `matchForm`. Loading ID lists over and over for _every_ call to the bulk preview API is likely the reason for the heap space errors occurring lately.

The key fix is to not provide a `Whelk` instance here https://github.com/libris/librisxl/blob/e91e02ac236ae0b176262fc180139a22f4af3439/housekeeping/src/main/groovy/whelk/housekeeping/BulkChangePreviewAPI.java#L97 while the rest is to clarify a bit what's going on when loading IDs.


jannistsiroyannis (approved):
LGTM!

PR #1576, opened 2025-03-25
Support custom field_value_factor function scoring via API

kwahlin:
Example usage:
`_fnScore=reverseLinks.totalItemsByRelation.instanceOf;1000;log1p,reverseLinks.totalItemsByRelation.itemOf.instanceOf;1000;log1p`

bild


olovy (approved):
Nice! :ship:

PR #1577, opened 2025-03-27
experimental(search2): enable boosting on field value

kwahlin:
Example usage:
`_fnBoost=mfv;language.@id;https://id.kb.se/language/swe;10` (Add 10 to the total score if language is Swedish).

There might be more effective ways to achieve the same thing (e.g. https://www.elastic.co/guide/en/app-search/current/boosts.html#boosts-value-boosts) should we decide that boosting on specific values is something we want. 

Note that function type now needs to be specified in `_fnBoost`. Use `mfv` for boosting on specific value (see above example) and `fvf` for field value factor, e.g. `fvf;reverseLinks.totalItems;10;ln1p;0;10`.


olovy (approved):
Parameter values should be float, not int?

Other than that, LGTM!

PR #1580, opened 2025-04-03
Normalize score for _topChipStr

kwahlin:
For free text search, create two different queries and put them in a `should` clause. The reason that we need two variants is that, when adding a score function to a query, that function will apply to _all_ fields in the query and we want it to apply only to the `_topChipStr` field. However it seems that all terms must be matched in _the same_ query for the score from that query to be counted. So if we have the following query
```
"should": [
                        {
                          "simple_query_string": {
                            "default_operator": "AND",
                            "query": "pippi långstrump",
                            "analyze_wildcard": true,
                            "fields": [
                              "_cardStr^50.0",
                            ]
                          }
                        },
                        {
                          "script_score": {
                            "script": {
                              "source": "_score * 1 / doc['_topChipStr.length'].value"
                            },
                            "query": {
                              "simple_query_string": {
                                "default_operator": "AND",
                                "query": "pippi långstrump",
                                "analyze_wildcard": true,
                                "fields": [
                                  "_topChipStr^400.0",
                                ]
                              }
                            }
                          }
                        }
                      ]
```
and `_topChipStr="pippi"` and `_cardStr="pippi långstrump"`, only the `_cardStr` score will count.

The solution I found was to include all fields that we want to search in both queries, although making sure that a field won't be counted twice by setting its boost factor to 0 in either query:

```
"should": [
                        {
                          "simple_query_string": {
                            "default_operator": "AND",
                            "query": "pippi långstrump",
                            "analyze_wildcard": true,
                            "fields": [
                              "_searchCardStr^1.0",
                              "_cardStr^50.0",
                              "_topChipStr^0.0",
                              "_topChipStr.exact^0.0",
                              "_cardStr.exact^50.0",
                              "_searchCardStr.exact^1.0",
                              "_chipStr^200.0",
                              "_chipStr.exact^200.0"
                            ]
                          }
                        },
                        {
                          "script_score": {
                            "script": {
                              "source": "_score * 1 / doc['_topChipStr.length'].value"
                            },
                            "query": {
                              "simple_query_string": {
                                "default_operator": "AND",
                                "query": "pippi långstrump",
                                "analyze_wildcard": true,
                                "fields": [
                                  "_searchCardStr^0.0",
                                  "_cardStr^0.0",
                                  "_topChipStr^400.0",
                                  "_topChipStr.exact^400.0",
                                  "_cardStr.exact^0.0",
                                  "_searchCardStr.exact^0.0",
                                  "_chipStr^0.0",
                                  "_chipStr.exact^0.0"
                                ]
                              }
                            }
                          }
                        }
                      ]
```

There might be less "hacky" ways to do this but it seems to work as intended!


olovy (approved):
Let's try it out! :ship:

PR #1581, opened 2025-04-07
Add filter alias for My libraries

lrosenstrom:
~* Add `myLibraries` filter alias. Works like other aliases such as `includeEplikt`, but the definition of the alias is parsed from the request query parameters.~

~* Include more data in stats so that frontend can exclude this filter from the "Other" filters below the facets (among other things).~

~Depends on https://github.com/libris/definitions/tree/feature/filter-definitions.~

* Support generic query defined aliased filters. 
* Expects a `alias-`-prefixed part of `q` (e.g.` alias-myLibraries`) and a corresponding query parameter passed from the frontend (`_alias-myLibraries` in this case)
* This is independent of definitions, hence `prefLabelByLang` must by supplied in `i18n` in the frontend.
* `search.mapping` contains a parsed version of the filter:

Screenshot from 2025-04-15 15-00-03

### DIFF
--- a/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
+++ b/rest/src/main/groovy/whelk/rest/api/SearchUtils2.java
@@ -34,7 +34,7 @@ public class SearchUtils2 {
         }
 
         QueryParams queryParams = new QueryParams(queryParameters);
-        AppParams appParams = new AppParams(getAppConfig(queryParameters));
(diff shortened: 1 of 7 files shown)


lrosenstrom:
Added preparations for distinguishing query defined filters from regular aliased filters in [23b71e7](https://github.com/libris/librisxl/pull/1581/commits/23b71e7828e2eb816010513a3ac42e240a23ea9f) . These can be made accessible as a field when needed.

PR #1582, opened 2025-04-07
Feature/search2 miscellaneous

kwahlin:
I think this is good to go now.

Depends on https://github.com/libris/definitions/pull/520.

Main changes:
- Enable applying different filters depending on search context. When to a apply a certain filter is declared in apps.jsonld (https://github.com/libris/definitions/blob/44775cd5527be2e85f726da9f570dc83caf22fd1/source/apps.jsonld#L107).  Solves https://kbse.atlassian.net/browse/LWS-189.
- Filtering did not work properly in the "1-to-many" view. For example the "has image" filter did not work due to the image property appearing at instance level and the search path would not expand to `hasInstance.image`. This is now fixed and the predicate-object search in general has been refactored and should now work as intended (up for discussion though how it really should work!).
- Use Elasticsearch multi query API for predicate-object instead of doing two round-trips to Elastic.
- A new class, `VocabMappings`, has been factored out from the `Disambiguate` class. This class is instantiated only once while we create a new Disambiguate instance for every query, as we now also load it with term mappings coming from application configuration or query parameters (i.e. filters). This also facilitates unit testing as it's easier now to provide "real data" and parse query strings. More unit tests have been added and the old ones have been rewritten and hopefully make more sense now.
- The `_i` parameter has been binned as it's no longer used anywhere. The part of the code that handled `_i` was a bit constrained so it's nice to get rid of it.
- Load all links appearing in facets, search mappings etc. at once instead of running a separate database query for each link. See https://github.com/libris/librisxl/blob/a0fed4db6095779bb11e7bf2f5b6ba19ce41523d/whelk-core/src/main/groovy/whelk/search2/Query.java#L268.
- General refactoring where the main change is that we now wrap the entire query in one object (`Query`).

I've probably missed mentioning something. May add more to the list later.


whelk-core/src/main/groovy/whelk/search2/Query.java
```diff
@@ -0,0 +1,476 @@
...
+            return DocumentUtil.NOP;
+        });
+        return framedThing;
+    }
+
+    private class LinkLoader {
+        private final Map<String, Collection<Link>> links = new HashMap<>();
+
+        private void loadChips() {
+            whelk.bulkLoad(links.keySet()).forEach((id, doc) -> {
```
olovy, line 272: should be getCards as in the old implementation?

whelk-core/src/main/groovy/whelk/search2/querytree/Link.java
```diff
@@ -1,17 +1,48 @@
...
+
+    public void loadChip(Map<String, Object> chip) {
+        this.chip.putAll(chip);
     }
 
-    public Link(String string) {
-        this(string, Collections.emptyMap());
+    public void loadThing(Map<String, Object> thing) {
+        this.thing.putAll(thing);
+    }
```
olovy, line 38: Found this naming a bit confusing since "LinkLoader" also does "load".
`setChip`/`setThing`?

olovy (approved):

PR #1583, opened 2025-04-10
Feature/lxl 4134

jannistsiroyannis:
This changes Elastic indexing to occur strictly in order with postgres writes.

Every change is logged (in a table) with a sequence number (assigned within the postgres transaction), and indexing occurs by reading that log in the same order. There is still a tolerance for bad data, in the sense that any 400-response from Elastic will be error-logged but ignored. Other types of errors however (be they timeouts, broken connections, or whatever else) are NOT TOLERATED, and will be retried, in order (forever).

A big change for developers in this, is that the housekeeping module must now be running for any indexing to occur at all. It is however perfectly fine to start it up late (it will catch up on its own).

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -395,7 +395,6 @@ class Whelk {
             return false
         }
 
-        //reindexUpdated(updated, preUpdateDoc)
         sparqlUpdater?.pollNow()
 
         return true
@@ -410,7 +409,6 @@ class Whelk {
             return
         }
 
-        //reindexUpdated(updated, preUpdateDoc)
         sparqlUpdater?.pollNow()
     }
 
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -302,117 +302,6 @@ class Whelk {
                 .collectEntries { id, doc -> [(idMap.getOrDefault(id, id)): doc] }
     }
 
-    private void reindexUpdated(Document updated, Document preUpdateDoc) {
-        indexAsyncOrSync {
-            elastic.index(updated, this)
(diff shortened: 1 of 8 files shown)


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -1624,24 +1658,21 @@ class PostgreSQLComponent {
     boolean saveVersion(Document doc, Connection connection, Date createdTime,
```
olovy, line 1658: Could be changed to `void` ?

whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -505,7 +385,7 @@ class Whelk {
      */
     boolean storeAtomicUpdate(String id, boolean minorUpdate, boolean writeIdenticalVersions, String changedIn, String changedBy, UpdateAgent updateAgent) {
         Document preUpdateDoc = null
```
olovy, line 387: I believe this variable can be removed?

whelk-core/src/main/groovy/whelk/Whelk.groovy
```diff
@@ -524,13 +403,12 @@ class Whelk {
     void storeAtomicUpdate(Document doc, boolean minorUpdate, boolean writeIdenticalVersions, String changedIn, String changedBy, String oldChecksum) {
         normalize(doc)
         Document preUpdateDoc = storage.load(doc.shortId)
```
olovy, line 405: This variable is unused now and can be removed

olovy (approved):
Nice! LGTM!

Remaining possibility for inconsistencies:
If an elastic request fails after increment/decerement reverse links they will be bumbed again when the whole document is retried. I think this is fine for now. Lots of added complexity to fix this?

I would like to have a gauge for "seconds behind" in /metrics so that we can track it. 
I can add that.

Interesting to see the transcribed (groovy -> java) indexing code. A bit more clunky but not that bad.

PR #1584, opened 2025-04-15
Feature/property search scoring

kwahlin:
Count score from free text search within specific fields, e.g.  `medverkande:strindberg`. 

The boost factor of 400 is set arbitrarily now as a starting point.

Properties that may appear on both instance and work will be "favored" with this solution as each individual field (e.g. `hasTitle.mainTitle`/`@reverse.instanceOf.hasTitle.mainTitle`) will contribute to the score. This is of course problematic, but maybe let that be until we've settled the whole instance vs work issue. (We should soon!)

We should also sort out the use of keyword fields as discussed.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/querytree/QueryTree.java
+++ b/whelk-core/src/main/groovy/whelk/search2/querytree/QueryTree.java
@@ -4,6 +4,7 @@ import whelk.JsonLd;
 import whelk.exception.InvalidQueryException;
 import whelk.search2.AppParams;
 import whelk.search2.Disambiguate;
+import whelk.search2.EsMappings;
 import whelk.search2.Filter;
 import whelk.search2.EsBoost;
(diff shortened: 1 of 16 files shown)


whelk-core/src/main/groovy/whelk/search2/querytree/QueryTree.java
```diff
@@ -310,6 +306,10 @@ private static Node _removeTopLevelNodesByCondition(Node tree, Predicate<Node> p
         };
     }
 
+    private static boolean isWild(Node tree) {
+        return tree instanceof FreeText ft && ft.isWild();
```
olovy, line 310:
```suggestion
    private static boolean isWild(Node node) {
        return node instanceof FreeText ft && ft.isWild();
```

olovy (approved):
LGTM!

PR #1587, opened 2025-04-29
Feature/lws 102

jannistsiroyannis:
Add a (partial, because no full CQL-compliance) implementation of the SRU protocol, which should cover the needs of our current SRU/z39 user base.

Test locally for example like so:
`curl "http://localhost:8187/?operation=searchRetrieve&query=isbn=9789130008650"`

### DIFF
--- a/sru/src/main/java/whelk/sru/servlet/SruServlet.java
+++ b/sru/src/main/java/whelk/sru/servlet/SruServlet.java
@@ -116,7 +116,6 @@ public class SruServlet extends HttpServlet {
                 writer.writeEndElement(); // recordPacking
 
                 writer.writeStartElement("recordSchema");
-                // This is in imitation of the old implementation. Perhaps it should really be http://www.loc.gov/MARC21/slim ? *shrug*
                 writer.writeCharacters("info:srw/schema/1/marcxml-v1.1");
                 writer.writeEndElement(); // recordSchema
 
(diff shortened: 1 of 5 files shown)


sru/src/main/java/whelk/sru/servlet/SruServlet.java
```diff
@@ -0,0 +1,140 @@
...
+
+        String queryString = parameters.get("query")[0];
+        String instanceOnlyQueryString = "(" + queryString + ") AND type=Instance";
+
+        Map<String, Object> results;
+        try {
+            // This part is a little weird
+            HashMap<String, String[]> paramsAsIfSearch = new HashMap<>();
+            String[] q = new String[]{instanceOnlyQueryString};
+            paramsAsIfSearch.put("_q", q);
```
olovy, line 67:
```suggestion
            paramsAsIfSearch.put("_q", q);
            paramsAsIfSearch.put("_stats", new String[] { "false" }); // don't need facets
```

can save some work in elastic here

olovy (approved):
Awesome

sru/src/main/java/whelk/sru/servlet/SruServlet.java
kwahlin: Could maybe wrap all of this in a static method of `Query`? Something like `Query.simpleQuery(String q, ...)` (no filters, facets etc.)

PR #1588, opened 2025-04-29
Feature/es score quoted freetext

kwahlin:
Should improve the relevancy for queries like
https://beta.libris-qa.kb.se/find?_q=Physiology+of+sport+and+exercise&_limit=20&_offset=0&_sort=&_spell=true
https://beta.libris-qa.kb.se/find?_q=jakt+och+fiske&_limit=20&_offset=0&_sort=&_spell=true
https://beta.libris-qa.kb.se/find?_q=h%C3%A4star+h%C3%A4star+h%C3%A4star&_limit=20&_offset=0&_sort=&_spell=true

A small bugfix (f868209613b58a84152f937ef275c361d129c087) was required for the last example to work. Before this fix, the query `hästar hästar hästar` would be transformed into just `hästar` which didn't really matter before however now that we also want to match the entire phrase it does matter.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/QueryUtil.java
+++ b/whelk-core/src/main/groovy/whelk/search2/QueryUtil.java
@@ -25,10 +25,14 @@ public class QueryUtil {
 
     public static String quoteIfPhraseOrContainsSpecialSymbol(String s) {
         // TODO: Don't hardcode. Keep quotes in parsing instead of requoting?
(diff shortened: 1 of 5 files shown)


olovy (approved):
Let's try.

TODO: evaluate if the quoted version of the query should have the same boost weights.

e.g.
query: _T1_ _T2_

doc 1: _T1_ T3 _T2_
vs
doc 2: T4 T5 _T1_ _T2_ T6 T7

PR #1589, opened 2025-05-06
Link isPartOf

andersju:
See https://kbse.atlassian.net/browse/LXL-4645

### DIFF
--- a/whelktool/scripts/cleanups/2025/05/lxl-4645-link-ispartof-redux.groovy
+++ b/whelktool/scripts/cleanups/2025/05/lxl-4645-link-ispartof-redux.groovy
@@ -1,0 +1,184 @@
+/**
+ * Replace local isPartOf with link
+ * Some code borrowed from ../../2020/11/lxl-3379-link-isPartOf.groovy
+ * See https://kbse.atlassian.net/browse/LXL-4645
+ */
+
+String where = """
+    collection = 'bib' and deleted = false and data#>>'{@graph,1,isPartOf}' LIKE '%"controlNumber":%'
+"""
+
+skipped = getReportWriter("skipped")
+
+selectBySqlWhere(where) { doc ->
+    def source_thing = doc.graph[1]
+    def _logSkip = { msg -> skipped.println("${doc.doc.getURI()}: ${msg}") }
+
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2025/05/lxl-4645-link-ispartof-redux.groovy
```diff
@@ -0,0 +1,251 @@
...
+        _logSkip("more than one describedBy: ${isPartOf.describedBy}")
+        return
+    }
+
+    def describedBy = isPartOf["describedBy"][0]
+    if (!(describedBy instanceof Map && describedBy.keySet().equals(["@type", "controlNumber"].toSet()))) {
+        _logSkip("describedBy contains something other than [@type, controlNumber]: ${describedBy.keySet()}")
+        return
+    }
+    if (!(describedBy["controlNumber"] instanceof String)) {
```
kwahlin, line 52: Could also be a single-element array (containing a string)? This goes for all places where a string value is expected.

kwahlin (approved):
Thorough work! LGTM!

whelktool/scripts/cleanups/2025/05/lxl-4645-link-ispartof-redux.groovy
```diff
@@ -0,0 +1,251 @@
...
+    }
+    return null
+}
+
+static String sanitize(String value) {
+    return value.replaceAll(/\9/, '')
+}
+
+static Set extractWords(String title) {
+    return title.replaceAll(/[^a-zA-Z0-9 ]/, "").toLowerCase().split('\\s+')
```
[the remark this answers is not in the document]
andersju: Good point! There are indeed some cases like that. Changed it so that diacritics are removed first.

whelktool/scripts/cleanups/2025/05/lxl-4645-link-ispartof-redux.groovy
andersju: Looking at the logs there were 164 records (with local isPartOf) where controlNumber was a list, and in each case it had multiple elements, often with weird-looking stuff -- we can deal with stragglers like that later.
(And only one instance where hasTitle.mainTitle was not a string)

PR #1590, opened 2025-05-07
Initialize whelk in a safe way

andersju:
A problem common to most or all (?) of the XL services is that Whelk is initialized in a way that if it fails -- usually because the database is unavailable, due to scheduled maintenance -- the servlet simply crashes and requests will 404 until the systemd service is restarted. This has caused multiple partial outages.

So let's make it more resilient. This should fix the problem for the OAI-PMH service. If it looks reasonable I can do the same for the others.


olovy:
> > I tried to sketch out a way to do this in a reusable manner here: #1591 Overengineered? It won't work for all the static stuff in OaiPmh though. So I think OaiPmh can be handled separately.
> 
> Nice! I'm in favor of your reusable approach. I think it's lagom engineered. So maybe merge this one as-is and (possibly) refactor the whole thing later, and do all the rest in #1591?

Great, you can continue in #1591

olovy (approved):

PR #1593, opened 2025-05-15
Feature/multi select filters

kwahlin:
Add support for multi-select filters.

It pretty much works the same way as before (i.e. as in the old `search` module), however some flexibility has been added for better consistency with the new query language.

The new class `SelectedFilters` extracts all query tree nodes that may map to a filter and can then be used for looking up whether a filter option has been selected or not. Special filters (e.g. "my libraries") and default filters (e.g. type:Work) are now also handled by this class. 

The default appearance of a filter group is still configured in https://github.com/libris/definitions/blob/develop/source/apps.jsonld but can easily be overridden, should one prefer "OR-filter" over "AND-filter" or vice versa. There isn't (yet) a way toggle between the two modes in the search interface but for now the mode automatically switches depending on the query's composition. See examples below (AND vs. OR).
bild

bild

For more complex expressions that can't be mirrored in the filter panel, the concerned filter group will be excluded. See example:

bild


olovy (approved):
Nice! LGTM.

Did a little bit of testing locally and seems to work fine.

Let's discuss how to deal with facets that should be togglable between AND and OR.

PR #1596, opened 2025-05-23
fix(bulk): Memory leak in whelktool/bulk job

olovy:
Fix memory leak when WhelkTool is used inside long running process e.g. bulk jobs in housekeeping. Caused by Runtime.addShutDownHook for each WhelkTool instance.


jannistsiroyannis (commented):
It seems to me that main2() still adds a shutdown hook on every invocation, and that main2() will still be called repeatedly before JVM shutdown, from housekeeping. At least from the ScriptRunner class. Will this not still leak then?

olovy:
> It seems to me that main2() still adds a shutdown hook on every invocation, and that main2() will still be called repeatedly before JVM shutdown, from housekeeping. At least from the ScriptRunner class. Will this not still leak then?

Good catch. Only focused on BulkJob

PR #1598, opened 2025-05-26
feat(search2): Allow escaping of one-char wildcard (?)

olovy:
When ? only appears as the last character of a word (or words) we don't
treat it as a wildcard. It is unlikely that the user intent is to use a
wildcard. This makes pasted titles work as queries e.g. `who is who?`.

If ? is used in any other position, wildcards are enabled. Including trailing.
examples: `blom??uist`, `wom??`, `?ome?`.

Add the possibility to force a single trailing ? to be interpreted as a
wildcard by escaping it: `who\?`. (or maybe rather "unescaping" it)

Also fix  StringIndexOutOfBoundsException in Lex when last
character in string is escaped https://github.com/libris/librisxl/pull/1598/commits/1c65dcbdcf2e1df5767dea18752be13374938dd5

TODO?
We might want to handle this in a different way in the lexing step.
i.e. by parsing ? as a symbol. Worth it?

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/QueryUtil.java
+++ b/whelk-core/src/main/groovy/whelk/search2/QueryUtil.java
@@ -134,13 +134,36 @@ public class QueryUtil {
                 .orElse(Collections.emptyMap());
     }
 
-    static Pattern NON_SIMPLE_QUERY = Pattern.compile("([*?])\\S+");
+    // leading wildcards e.g. "*foo" are removed by simple_query_string
+    static Pattern NON_SIMPLE_QUERY = Pattern.compile("\\\\[?]|([*?])\\S+");
+
     /**
      * Can this query string be handled by ES simple_query_string?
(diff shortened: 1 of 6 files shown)


whelk-core/src/main/groovy/whelk/search2/QueryUtil.java
```diff
@@ -143,4 +143,27 @@ public static boolean isSimple(String queryString) {
         // leading wildcards e.g. "*foo" are removed by simple_query_string
         return !NON_SIMPLE_QUERY.matcher(queryString).find();
     }
+
+    public static String escapeNonSimpleQueryString(String queryString) {
```
olovy, line 147: Adapted from https://github.com/libris/librisxl/blob/fd6d9638dba79759aea2e1aa8b1c55ddf8c0440e/whelk-core/src/main/groovy/whelk/search/ESQuery.groovy#L863

PR #1601, opened 2025-06-13
fix(bulk): Skip and log conflict when moving holdings

olovy:
- Skip and log ConflictingHoldException in bulk update (same behavior as bulk create)
- Log to standard FAILED.txt and ERRORS.txt so that failed docs are included in count.
- Make the above logs available to scripts through reportFailed() on DocumentItem


kwahlin (approved):
LGTM

PR #1603, opened 2025-06-24
Bugfix/empty whelktool reports

kwahlin:
Should help to avoid unnecessary log warnings of type

`2025-06-24T11:24:17,150 [Thread-37218] WARN  whelk.datatool.bulkchange.BulkJob - Could not get line count`


andersju (approved):
LGTM!

PR #1604, opened 2025-07-01
Feature/simpler view links

kwahlin:
Enables https://github.com/libris/lxlviewer/pull/1338 (see Jesper's TODO comment). The `_limit` parameter won't be included in find links for facets, pills, suggestions etc. This isn't problem since we have a hard coded default limit (20) as fallback when `_limit` is missing. We probably do want to include `_sort` however, since e.g. removing/adding a filter shouldn't change the sort order.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/QueryParams.java
+++ b/whelk-core/src/main/groovy/whelk/search2/QueryParams.java
@@ -6,7 +6,6 @@ import whelk.exception.InvalidQueryException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


whelk-core/src/main/groovy/whelk/search2/QueryParams.java
```diff
@@ -83,57 +82,79 @@ public QueryParams(Map<String, String[]> apiParameters) throws InvalidQueryExcep
...
+                        params.put(ApiParams.SORT, sort);
+                    }
+                }
+                case ApiParams.OFFSET -> {
+                    if (offset > 0) {
+                        params.put(ApiParams.OFFSET, "" + offset);
+                    }
+                }
+                case ApiParams.LIMIT -> params.put(ApiParams.LIMIT, "" + limit);
+                case ApiParams.LENS -> {} // TODO
```
kwahlin, line 124: I may have discovered a bug related to the lens parameter. Will investigate it further. No harm in omitting it for now.

PR #1609, opened 2025-07-30
fix(emm): Use xattrs for dump creation time checking

andersju:
We've used `BasicFileAttributes.creationTime()` to check the creation time of EMM dumps. This is supported by ext4. However, we use XFS on our servers, which _doesn't_ support creation time, so: "If the file system implementation does not support a time stamp to indicate the time when the file was created then this method returns an implementation specific default value, typically the [last-modified-time](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/nio/file/attribute/BasicFileAttributes.html#lastModifiedTime()) or a FileTime representing the epoch (1970-01-01T00:00:00Z)."

So let's use something else. [Extended file attributes](https://wiki.archlinux.org/title/Extended_attributes) aren't _super_ portable but are supported by all the major Linux file systems and I've verified that it works on our servers too. (There are various other ways to solve this but this seemed like the cleanest option.)
```
/t/dumps getfattr -d itemAndInstance:SRo.dump
# file: itemAndInstance:SRo.dump
user.creationTime="2025-07-30T14:13:44.446938635Z"
```

https://kbse.atlassian.net/browse/LXL-4664

### DIFF
--- a/emm/src/main/java/whelk/Dump.java
+++ b/emm/src/main/java/whelk/Dump.java
@@ -21,11 +21,11 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.attribute.BasicFileAttributes;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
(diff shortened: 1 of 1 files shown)


olovy (approved):
:ship:

PR #1613, opened 2025-08-01
feat(rest): Migrate from Groovy to Java, part I

andersju:
Experimenting with letting AI overlords do some of the boring work. Let's start slowly migrating stuff to Java where it makes sense. I've reviewed the code (and made some changes). All tests work. If there's some subtle change that matters, let's fix that and also add a test for it.


olovy (approved):
LGTM!

PR #1616, opened 2025-08-14
 Move Elasticsearch boost settings to dedicated configuration file

kwahlin:
- Created a configuration file to store all Elasticsearch boost parameters.
- Removed hard-coded boost settings from the code.
- Updated relevant components to read boost values from the configuration file.

I'm not sure where to place the configuration, probably not where it is right now (whelk-core) since the configuration is app specific. However, I don't think it makes sense to put it in the definitions repo either (apps.jsonld).
I'm also unsure how we want this to be documented, but a README in Swedish has been included to describe the parameters for now. Feedback on both file placement and documentation approach is welcome.


kwahlin:
I think you're right, it's probably the most fitting place in the current structure. Thanks!

olovy (approved):
LGTM!

Agree with placing the boost settings next to the other ES config.
Ideally both would be split now that e.g. _topChipStr, _chipStr are XL "platform" fields while e.g. reverseLinks.totalItemsByRelation.instanceOf.contribution.agent is a Libris field.

Don't we want to still be able to experiment with tweaking boost settings via query paramenters?

andersju (approved):
LGTM!

PR #1617, opened 2025-08-18
feat(search): Handle ISBNs with hyphens

olovy:
Remove hyphens from strings that look like ISBNs in queries. e.g. 978-917-8034-239
For now only works if the whole query is an ISBN.
Should we update it to search for both the original string and the one with hyphens removed?


kwahlin (approved):
LGTM! Let's go with this for now.

PR #1620, opened 2025-08-21
feat(api): Add Libris legacy xsearch API

olovy:
Add partial support for xsearch on top of XL.

https://libris.kb.se/help/xsearch_swe.jsp?open=tech
https://libris.kb.se/xsearch?query=qwerty&format=json&start=1

Tries to match the behaviour of the existing API. 
e.g. return error as XML with code 200 even when json requested.

I haven't bothered looking at the old implementation yet.
This is just reverse engineered from the docs + API responses.

Implemented "json" from scratch before I added the old transform for MODS.
"json" could be transformed in the same way.


jannistsiroyannis (approved):
LGTM!

PR #1622, opened 2025-08-28
perf: Speed up DocumentUtil.traverse

olovy:
It is used in many places, e.g. when shaping docs for ES indexing.

Just porting it to Java without any other changes cut indexing time with xx% on my computer.
It can probably be improved.

Before
Skärmbild från 2025-08-28 13-42-44

After
Skärmbild från 2025-08-28 13-46-09

Before
Skärmbild från 2025-08-28 13-19-46

After
Skärmbild från 2025-08-28 13-43-18

Seems like it was partly because of lock contention in the groovy interpreter. on org.codehaus.groovy.vmplugin.v8.CacheableCallsite at DFS.node/descend?
Skärmbild från 2025-08-28 10-01-13

### DIFF
--- a/whelk-core/src/main/groovy/whelk/util/DFS.java
+++ b/whelk-core/src/main/groovy/whelk/util/DFS.java
@@ -1,0 +1,59 @@
+package whelk.util;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+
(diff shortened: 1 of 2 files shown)


olovy:
> Want it even faster? Write it with loops and a little local stack for backtracking, instead of recursion. My guess is you get another ~3x improvement :)

Yes that was the plan :)

lrosenstrom (approved):
Sounds great!

PR #1623, opened 2025-09-01
Index Libris record ids into a separate field

kwahlin:


whelk-core/src/main/groovy/whelk/util/FresnelUtil.java
```diff
@@ -195,6 +195,10 @@ public void insertComputedLabels(Object data, LangCode locale) {
         });
     }
 
+    public List<?> fslSelect(Map<?, ?> thing, String fslSelector) {
```
olovy, line 198: :heart:

olovy (approved):
LGTM!

PR #1624, opened 2025-09-02
Feature/improve length normalizer

kwahlin:
Will improve relevancy for e.g. https://beta.libris-qa.kb.se/find?_q=mio+min+mio&_limit=20&_offset=0&_sort=&_spell=true

Apparently it's good to use parameters for scripts: https://www.elastic.co/docs/explore-analyze/scripting/modules-scripting-using#prefer-params


olovy (approved):
:ship:

PR #1631, opened 2025-09-09
feat(whelktool): add script for more linking of isPartOf

andersju:
Essentially the same as #1589 and #1592 but with more action in `isSeeminglySameTitle()` (should possibly just have modified that one, but :shrug: )

https://kbse.atlassian.net/browse/LXL-4676

### DIFF
--- a/whelktool/scripts/cleanups/2025/09/lxl-4676-link-ispartof-yet-again.groovy
+++ b/whelktool/scripts/cleanups/2025/09/lxl-4676-link-ispartof-yet-again.groovy
@@ -1,7 +1,8 @@
 /**
  * Replace local isPartOf with link
+*  Based on ../../2025/05/lxl-4645-link-ispartof-redux.groovy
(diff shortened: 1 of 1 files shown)


whelktool/scripts/cleanups/2025/09/lxl-4676-link-ispartof-yet-again.groovy
```diff
@@ -0,0 +1,466 @@
...
+    String info
+
+    if (!isMatch) {
+        Set allTargetWords = new HashSet()
+        extractValues(targetThing).each { allTargetWords.addAll(extractWords(it)) }
+        if (allTargetWords.containsAll(sourceTitleWords)) {
+            isMatch = true
+        } else {
+            def wordDiff = sourceTitleWords - allTargetWords
+            if (wordDiff.size() < 3) {
```
olovy, line 387: Source titles with only one or two words will always match or am I missing something here?

olovy (approved):
LGTM!

PR #1636, opened 2025-09-16
add ÖM-removal script.

jannistsiroyannis:


olovy (approved):
LGTM

There is a difference in how single and multiple labels are handled.
Single labels are not removed if the result after removing "ÖM" is an empty string?

jannistsiroyannis:
> LGTM
> 
> There is a difference in how single and multiple labels are handled. Single labels are not removed if the result after removing "ÖM" is an empty string?

Good catch, i will fix!

olovy (approved):

PR #1638, opened 2025-09-24
build: bump various deps

andersju:
Shouldn't be anything breaking, just lots of fixes (sometimes a decade's worth!), including security fixes.

Edit: also updated the Gradle shadow plugin (for fat jars). Lots of [breaking changes](https://github.com/GradleUp/shadow/blob/main/docs/changes/README.md) but none that affect our (very simple!) usage, as far as I can tell.


olovy (approved):
LGTM

PR #1641, opened 2025-10-01
Feature/sort out year fields

kwahlin:
The main objective of this change is to fix bugs in year queries, particularly:
- Truncated/masked queries (e.g. `publication.year=19*`)
- Queries with malformed years (e.g. `publication.year=nnnn`)

Both previously failed due to the `numeric_extractor` analyzer, which only matched clean 4-digit strings.

Year fields (year, startYear, endYear) are used for various operations:  
- **Querying**: term queries, range queries, and masked/truncated queries.  
- **Aggregating**: grouping by publication year.  
- **Sorting**: ordering search results by publication year.  

Previously, year values were stored only as analyzed text and plain keywords. This PR changes the text analyzer from `numeric_extractor` to `standard` in order to support masking/truncation and matching malformed years. Valid years (yyyy) are also stored in two new sibling fields, `*_4_digits_int` and `*_4_digits_keyword` which are used for efficient queries, aggregations and sorting.

#### Example
Previous mapping:
- `publication.year` → `text` (analyzer: `numeric_extractor`)  
- `publication.year.keyword` → `keyword`  

New mapping:
- `publication.year` → `text` (analyzer: `standard`)  
- `publication.year.keyword` → `keyword` (unchanged, preserves raw value)
- `publication.year_4_digits_keyword` → `keyword` (new, normalized 4-digit values for aggregations and exact term queries)  
- `publication.year_4_digits_int` → `integer` (new, normalized 4-digit values for range queries and sorting)  

See [documentation](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/number) for motivation (section "Elasticsearch mapping numeric identifiers").

Queries:
- `yearPublished:1990` 
  - Before: `publication.year:"1990"` → matched analyzed text field (inefficient).  
  - Now: `publication.year_4_digits_keyword:"1990"` → keyword lookup.  

- `yearPublished>=1990`   
  - Before: `publication.year>="1990"` → range query on analyzed text field (inefficient).  
  - Now: `publication.year_4_digits_int>=1990` → proper numeric range query.  

- `yearPublished:nnnn`
  - Before: `publication.year:"nnnn"` → matched text field analyzed with `numeric_extractor` (failed).  
  - Now: `publication.year:"nnnn"`  → matches same field but now works with the `standard` analyzer.

- `yearPublished:19*`
  - Before: `publication.year:"19*"` → matched text field analyzed with `numeric_extractor` (failed).  
  - Now: `publication.year:"19*"`  → matches same field but now works with the `standard` analyzer.

Aggregation:
- Libris Search: `publication.year_4_digits_keyword` (clean 4-digit years only).
- Cataloging: `publication.year.keyword`: (“as-is” values, including malformed years).

Sorting:
- Sorting is performed on `publication.year_4_digits_int`.  
- Valid years are sorted numerically.  
- Malformed years are sorted last.
- Question: Do we also want to filter out known non-years like "9999" or "0000" from the 4-digit fields? 

---

#### Ingest Pipeline
- Added a script-based ingest pipeline to dynamically extract valid 4-digit years and populate `*_4_digits_keyword` and `*_4_digits_int` fields.  
- Pipeline is referenced in config and can be deployed via Fab task (see https://github.com/libris/devops/pull/282).  
- The pipeline script can easily be modified should we want to manipulate or normalize other fields dynamically in the future. 
- Indexing time increased by ~0.035 ms/document (~1h for full reindex). I think this is acceptable given the improved query-time efficiency.

### DIFF
--- a/librisxl-tools/elasticsearch/extract_4_digits.painless
+++ b/librisxl-tools/elasticsearch/extract_4_digits.painless
@@ -16,8 +16,11 @@ while (!stack.isEmpty()) {
           def matcher = /\d{4}/.matcher(value);
           if (matcher.find()) {
               def fourDigitStr = matcher.group(0);
-              updates.put(key + "_4_digits_keyword", fourDigitStr);
-              updates.put(key + "_4_digits_short", Short.parseShort(fourDigitStr));
+              // Exclude known non-year placeholders (0000, 9999) from 4-digit fields.
+              if (fourDigitStr != "0000" && fourDigitStr != "9999") {
+                  updates.put(key + "_4_digits_keyword", fourDigitStr);
+                  updates.put(key + "_4_digits_short", Short.parseShort(fourDigitStr));
+              }
           }
       } else if (value instanceof Map) {
           stack.push(value);
--- a/librisxl-tools/elasticsearch/extract_4_digits.painless
+++ b/librisxl-tools/elasticsearch/extract_4_digits.painless
@@ -1,0 +1,36 @@
+def fourDigitFields = ["year", "startYear", "endYear"];
+
+def stack = new ArrayDeque();
+stack.push(ctx);
+
+while (!stack.isEmpty()) {
+  def current = stack.pop();
+
+  def updates = new HashMap();
+
+  for (entry in current.entrySet()) {
+      def key = entry.getKey();
+      def value = entry.getValue();
+
+      if (value instanceof String && fourDigitFields.contains(key)) {
+          def matcher = /\d{4}/.matcher(value);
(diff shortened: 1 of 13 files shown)


olovy (changes requested):
Interesting!

ES docs: " you should pick the smallest type which is enough for your use-case." -> Use short instead of int for 4_digits

Future TODO: we need a way to specify that we want “as-is” values in aggregations when we switch cataloging to search2 / when we consolidate search+search2?

olovy (approved):
LGTM!

PR #1711, opened 2026-03-11
Fix missing 082 resourceType

klngwll:
No more Marc21 export mixup between hold fields 082 and 084 because of bib inherits. Also Reinstate broken spec on Hold 084.

Fixes LXL-3287


niklasl (approved):
LGTM (ping @branchedelac for visibility)

PR #1713, opened 2026-03-12
Feature/cql to xl query2

jannistsiroyannis:
Add translation of CQL to XL query language.
All valid CQL queries should be translated, but there *are* things that are lost in translation. In particular, relation modifiers, prefix assignments and sorting etc.


olovy (approved):
Nice! ⚙️ 

LGTM

PR #1715, opened 2026-03-13
Add method to determine if a field can be used in aggregations

kwahlin:
Prevents ES error caused by trying to aggregate on a text field.

Bonus effect: Fields that aren't indexed at all won't be included in the agg query either. So for example  `@reverse.instanceOf.language.@id` will now be dropped.

image


olovy (commented):
LGTM!

olovy (approved):

PR #1724, opened 2026-03-20
Restore dropped _categoryByCollection in embedded integral things

kwahlin:
Preserves `_categoryByCollection` for
- Linked works embedded into instances
- Instances embedded into linked works


olovy (approved):
🚢

PR #1727, opened 2026-03-20
feat(search2): Initial support for like operator ~

olovy:
Add support for `selector ~ <id>`

In addition to direct match on id this matches blank nodes whose `_str` contains all `toString(<id>, <needle lens>)`.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/querytree/Condition.java
+++ b/whelk-core/src/main/groovy/whelk/search2/querytree/Condition.java
@@ -8,6 +8,7 @@ import whelk.search2.QueryUtil;
 import whelk.util.Restrictions;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -16,6 +17,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static java.util.Objects.hash;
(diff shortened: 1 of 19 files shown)


whelk-core/src/main/groovy/whelk/search2/querytree/Condition.java
```diff
@@ -309,6 +315,27 @@ private Map<String, Object> esResourceFilter(String f, Resource r) {
         return esTermQueryFilter(f, r.jsonForm());
     }
 
+    private Map<String, Object> esLikeResourceFilter(String f, Link l, ESSettings esSettings) {
```
kwahlin, line 328:
```suggestion
    private Map<String, Object> esLikeResourceFilter(String f, Link l) {
```

whelk-core/src/main/groovy/whelk/search2/querytree/Condition.java
```diff
@@ -309,6 +315,27 @@ private Map<String, Object> esResourceFilter(String f, Resource r) {
...
+        if ("".equals(l.getNeedle())) {
+            return esResourceFilter(f, l);
+        }
+
+        var freeTextFilter = Map.of("simple_query_string", Map.of(
+                        "query", l.getNeedle(),
+                        "fields", List.of(selector.esField() + "." + SEARCH_KEY),
+                        "default_operator", "AND"));
+        var notLinked = mustNotWrap(existsFilter(f));
```
kwahlin:
```suggestion
        var notLinked = mustNotWrap(esResourceFilter(f, l));
```

Given the current use case `existsFilter` works here thanks to contribution being a nested field but the condition could fail for non-nested fields with multiple links? Better to filter on the specific link?

kwahlin (approved):
See comment, otherwise LGTM.

whelk-core/src/main/groovy/whelk/search2/querytree/Condition.java
olovy: You are right. We have to check if we are inside a nested field. In that case we do want filter out all other linked values. They are with certainty not the one we are looking for,

PR #1731, opened 2026-03-25
feat(search2): Add _stats=falseThisRequest

olovy:
So that stats can be disabled without this being reflected in links such as first, next etc.


whelk-core/src/main/groovy/whelk/search2/QueryParams.java
```diff
@@ -58,7 +58,27 @@ public static class Debug {
     public final String q;
     public final String r;
 
-    public final boolean skipStats;
+    public enum Stats {
+        FALSE(false),
+        FALSE_THIS_REQUEST(false),
+        TRUE(false);
```
kwahlin:
```suggestion
        TRUE(true);
```

kwahlin (approved):
LGTM! (almost)

PR #1736, opened 2026-03-31
Feature/instance vs work search

kwahlin:
Use new definitions in https://github.com/libris/definitions/pull/576 to avoid hardcoding and unnecessary complexity in query expansion


olovy (approved):
LGTM!

PR #1737, opened 2026-03-31
Feature/add complementary gfs as local entities

branchedelac:
https://kbse.atlassian.net/browse/FMT-514


niklasl (approved):
LGTM (Noting that the instance is saved even if work is linked; but if this is the most common case (and there are few instances in total), it's OK here.)

PR #1738, opened 2026-04-01
perf(search): don't do toCard on result items

olovy:
They are already (search) cards. toCard is an expensive operation

Profiling locally with [stress_search2_api_simple.py](https://github.com/libris/xl_stress/blob/master/locustfiles/stress_search2_api_simple.py)

All states (running, blocked, io, ...)

before:
elastic request ~78 %
collectItems...toCard ~7%

after:
elastic request ~88%

Response size does not grow.


andersju (approved):
Nice! LGTM!

PR #1739, opened 2026-04-08
perf(elastic): Replace JsonOutput with Jackson mapper

olovy:
Replace use of groovy std lib JsonOutput with Jackson mapper
Consistent with rest of code base.
Perhaps faster. Seems to remove a small allocation hot spot.

Skärmbild från 2026-04-08 15-29-22


andersju (approved):
LGTM!

PR #1740, opened 2026-04-08
perf(search2): cache chips for search mappings

olovy:
defaultSiteFilters are the same for all queries (for one site). They are reflected in search mappings. Meaning their chips are needed in all requests. So cache them. 10 minutes for now. Serving stale mappings should be harmless. We could expand this with an invalidation mechanism like in dependencyCache.

profiling with simple queries - stress_search2_api_simple.py

All states
LinkLoader.loadChips 9.7% -> 3.3%

More than 91% of time is now spent on the elastic call

Runnable
LinkLoader.loadChips 20.7% -> 12.4%

cache stats - over 90% hit rate
```
$ curl -s "http://localhost:8180/metrics" | grep chipCache
guava_cache_hit_total{cache="chipCache",} 90584.0
guava_cache_miss_total{cache="chipCache",} 8038.0
...
```

Before
Skärmbild från 2026-04-08 15-58-21

After
Skärmbild från 2026-04-08 17-20-03

More than 91% of time is now spent in the elastic call
Skärmbild från 2026-04-08 17-23-51

### DIFF
--- a/whelk-core/src/main/groovy/whelk/ChipCache.java
+++ b/whelk-core/src/main/groovy/whelk/ChipCache.java
@@ -1,0 +1,77 @@
+package whelk;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import org.jspecify.annotations.NullMarked;
+import whelk.util.Metrics;
+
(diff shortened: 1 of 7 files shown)


lrosenstrom (approved):
Nice! LGTM!

PR #1741, opened 2026-04-09
Various changes to improve indexing speed

andersju:
I did a whole lot of profiling (locally) while reindexing and before/after each of the following, plus full reindexes.

### Don't re-split path on every FslPath.select() call; pre-compile ArcStep regex

Straight-forward.

### Remove unnecessary/intermediate stuff in DFS

* Stack in synchronized, but we don't need that overhead here. So let's use a plain old ArrayList instead.
* Getting rid of `Collections.unmodifiableList(path)`: this was originally (when DFS was elsewhere) `path.collect()`, which was later changed to `Collections.unmodifiableList()`, which is cheaper. But this still allocates a wrapper on every node visit, and I'm not sure the defensiveness is justified.
* We don't need intermediate Node objects and thus not `descend()` either. The list/map isn't mutated during traversal so this should also be fine.

13-15% reduction in index time with this change (and the ArcStep regex).

### `@CompileStatic` all the things!
Well, at least some more. And the changes to make it work (e.g. explicit casts), including some restructuring of `_get`.

### doc and IRI->ID cache

I also added a doc and IRI->ID cache to CachingPostgreSQLComponent (which also necessitated some access modifierchanges). This reduced indexing time by about 15%. This also affects the card cache but I _think_ it's fine because the card cache is only used when reindexing (right?).

Stats after a full QA reindex:

INFO  whelk.component.CachingPostgreSQLComponent - Doc cache: CacheStats{hitCount=642912128, missCount=112343106, loadSuccessCount=60485781, loadExceptionCount=0, totalLoadTime=309437839975245, evictionCount=112088908}
INFO  whelk.component.CachingPostgreSQLComponent - IRI->ID cache: CacheStats{hitCount=1019271475, missCount=2146846, loadSuccessCount=1775640, loadExceptionCount=0, totalLoadTime=12590020579073, evictionCount=1646821}

99.8% hit rate for IRI-ID, 85% for the doc cache (lots of evictions, should try increasing DOC_CACHE_MAX_SIZE) for the latter.

### Jackson in Document.deepCopy

Should also be fine (see comment). Haven't done a full reindex with this but saw a 10-15% reduction in indexing time when testing locally.

### MAX_CONNECTIONS_PER_HOST + CONNECTION_POOL_SIZE

Was previously too low, at least when reindexing with lots of CPUs to the cluster. Needs more testing to figure out what's best. Maybe set it dynamically or make it configurable.

With all these changes, the ES cluster is still never the bottleneck. In the future we should do more Groovy-to-Java porting and maybe some larger restructuring/rethinking of the whole process.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Document.groovy
+++ b/whelk-core/src/main/groovy/whelk/Document.groovy
@@ -618,6 +618,7 @@ class Document {
         return _preparePath(path, data)
     }
 
+    @CompileStatic
     static boolean _preparePath(List path, Object root) {
         // Start at root data node
         Object node = root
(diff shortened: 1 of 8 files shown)


kaipoykio (commented):
Very nice improvements!

olovy (approved):
Very nice! LGTM!

> This also affects the card cache but I think it's fine because the card cache is only used when reindexing (right?).

The in-memory card cache in CachingPostgresqlComponent is only used when reindexing.
The card cache table is also used by embellish

https://github.com/libris/librisxl/blob/df622a09ec05fe58dd165dfdce528ac709d92672/whelk-core/src/main/groovy/whelk/Whelk.groovy#L616

PR #1742, opened 2026-04-09
perf(search): reuse app desc

olovy:
AppConfig is static. Actually use the cached version...
Saves a couple of percent CPU.
[perf(search): reuse app desc](https://github.com/libris/librisxl/commit/5b63408e652b3cecae1d84c64d8bc4665673583d)
TODO: don't pass the serialized version through queryParameters.

naming stuff:
[chore(search2): naming in AppParams](https://github.com/libris/librisxl/commit/00efafdc651454c945fd828ecf1f3161d7bfbadc)


andersju (approved):
LGTM!

PR #1743, opened 2026-04-09
feat(api): format parameter in /compilemarc. Set file extension.

olovy:
to be used in link from lxl-web
- Add format parameter that overrides profile if present
- Change downloaded file name to <fnurgel>.marc / <fnurgel>.marc.xml
    - could break some script somewhere out there?


andersju (approved):
LGTM

PR #1744, opened 2026-04-13
feat: make it possible to override secret.properties options on the command line

andersju:
We already have service-and-environment-specific Java memory options in our devops stuff. It's important that we can also handle secret.properties options per-service-and-environment in certain cases. So we can do (for example) `java -Dxl.secret.properties=/some/where/secret.properties -Dxl.sqlMaxPoolSize=32 -jar importers/build/libs/xlimporter.jar`.


olovy (approved):
LGTM!

kwahlin (approved):

PR #1748, opened 2026-04-16
fix(marcframe): ContributionByRoleStep missing required resources

olovy:
ContributionByRoleStep missing required resources
https://github.com/libris/librisxl/blob/540e92146f1f450d8c8e941aa6161849137e15b5/whelk-core/src/main/groovy/whelk/converter/marc/ContributionByRoleStep.groovy#L42

relatorResources was not initialized because elasticFind wasn't yet initialized. Because elastic initialization was moved in cf91b2066

Guarded because when running only core whelk elastic+elasticFind are not available. 
https://github.com/libris/librisxl/blob/540e92146f1f450d8c8e941aa6161849137e15b5/whelk-core/src/main/groovy/whelk/ResourceCache.groovy#L34

### DIFF
--- a/whelk-core/src/main/groovy/whelk/Whelk.groovy
+++ b/whelk-core/src/main/groovy/whelk/Whelk.groovy
@@ -163,6 +163,8 @@ class Whelk {
             elasticFind = new ElasticFind(new ESQuery(this))
         }
 
+        initDocumentNormalizers()
+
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
LGTM

niklasl (approved):
🚢

PR #1749, opened 2026-04-17
fix(search): Boost for lang container props

olovy:
fooByLang is no longer indexed.
Boost on flattened __foo instead

broken by cc86a031c4517d89953aa765eac0f99cd5400795


olovy:
FYI

PR #1756, opened 2026-04-28
Feature/o search  predicates

kwahlin:
Provide observations in `stats._predicates` in a more complete and intuitive format:

```
{
  "predicate": "<PREDICATE_DEFINITION>",
  "object": "<OBJECT_DESCRIPTION>",
  "view": { "@id": "/find?_q=p:o" },
  "totalItems": 100
}
```

Before, the `object` key confusingly held the predicate definition. This change puts the predicate definition under `predicate` and the actual object description under `object`. It also includes a ready-to-use `p:o` query in the `view` URL.

By moving query construction (`_q=p:o`) to the backend we prevent inconsistent query formatting which caused a recent bug, see https://github.com/libris/lxlviewer/pull/1576.


olovy (approved):
LGTM

PR #1765, opened 2026-05-11
fix: reindex affected virtual records

olovy:
- When reindexing "affected "records, virtual records were not included.
- isCardChangedOrNonexistent is no longer sufficient to know if other records are affected. Add workaround for Item


kwahlin (approved):
LGTM

PR #1766, opened 2026-05-11
Bugfix/es nested grouping

kwahlin:
Fix bugs and inconsistencies in how composite property search filters (for example `bestånd:`) are parsed and translated into ES queries.

Problem 1:
Repeated use of composite filters such as `bestånd:x bestånd:y` was broken due to bad ad hoc logic for handling repeated use of `contributor` in queries. On review, the logic no longer made sense and has now been removed. See 442f22e and fa22299.

Problem 2:
Targeted free-text queries such as `contributor:(x y)` were not translated into nested queries, despite the underlying field being indexed as nested. This didn't affect single-token queries, but rather multi-token queries where all terms are expected to match the same nested object. See fix: 140c39a

Problem 3:
Free-text searches against composite properties could fail when the query terms were split across different sub-properties of the composite property. For example `bestånd:(kalle-samlingen kalwah)` could fail to match if `kalle-samlingen` was indexed in `librissearch:shelf` and `kalwah` was indexed in `librissearch:internalItemNote`. Previously, the generated query searched each sub-property independently:
```
{
  "should": [
    {
      "simple_query_string": {
        "default_operator": "AND",
        "query": "kalle-samlingen kalwah",
        "fields": ["librissearch:internalItemNote"]
      }
    },
    {
      "simple_query_string": {
        "default_operator": "AND",
        "query": "kalle-samlingen kalwah",
        "fields": ["librissearch:shelf"]
      }
    }
  ]
}
```
This has been changed to instead search all sub-properties together in the same query:
```
{
  "simple_query_string": {
    "default_operator": "AND",
    "query": "kalle-samlingen kalwah",
    "fields": [
      "librissearch:internalItemNote",
      "librissearch:shelf"
    ]
  }
}
```
See fix: 948362d

Problem 4:
Parsing values as `Numeric` no longer serves a useful purpose and can instead lead to incorrect query processing. For example `bestånd:(berghman 2019)` was previously parsed as two separate `Condition`s with values `FreeText(berghman)` and `Numeric(2019)` respectively instead of a single condition `FreeText(berghman 2019)`. This distinction matters because the query is processed differently depending on the value type. See fix: ece260f

See also:
https://github.com/libris/definitions/pull/590
The effect of that change is that repeated use of `bestånd:` filter will be interpreted as referring the same item (unless separated by a non-repeatable property).


olovy (approved):
LGTM!

PR #1769, opened 2026-05-15
perf: Some more @CompileStatic

olovy:
`@CompileStatic` Document, Link, Relations, DependencyCache

seems to lower groovy internal allocations (invokedynamic etc) with ~13% for GET /tfld57b2r0bnggfn?framed=true

Still lots to improve.


jannistsiroyannis (approved):

olovy:
TODO: fix CrudSpec

this mock doesn't work anymore because groovy runtime is bypassed completely now that both arguments to determineLegacyCollection are CompileStatic (which is good)
```
LegacyIntegrationTools.determineLegacyCollection(_, _) >> {
  return "hold"
}
```

PR #1771, opened 2026-05-21
Bugfix/instance modified created

kwahlin:
This complements https://github.com/libris/definitions/pull/596 to fix instance created/modified queries and improves consistency in handling of record properties in queries.

- d9a63be: Fixes repeated `meta.meta` segment in expanded query paths introduced by the definitions change. However, I'm leaning towards removing the special handling of record properties entirely, see comment in https://github.com/libris/definitions/pull/596.
- 04930f7: After the definition change, e.g. `instanceRecordCreated` now correctly expands to `@reverse.instanceOf.meta.created` when querying works. However, in virtual work records the original record data was embedded directly on the virtual work entity (`meta`) rather than on the embedded instance (`@reverse.instanceOf.meta`). Thus only linked works were matched by these queries. This commit aligns the shape of virtual work records with linked work records, with the only difference that `meta` for virtual works will always have this minimal shape:
```
{
    "@id" : "http://libris.kb.se/xxxxxxxxxxxxxxx#work-record",
    "@type" : "VirtualRecord",
    "mainEntity" : {
      "@id" : "http://libris.kb.se/xxxxxxxxxxxxxxx#work"
    }
}
```
Although this solution should work well, I also experimented with an alternative shape that's even more like the linked work graphs: https://github.com/libris/librisxl/commit/d382f3b1ef0da0f9f44066909bd17e60cf39a6e3
Not sure which option is better.
- 1f211cb: The previous commit caused some links to be restored in the wrong place after shrinking to search card. The simple solution to this was do the work/instance flip earlier in `getShapeForIndex()`, which also made it possible to remove some awkward ad hoc logic: 1d9af2c
- c2d44c4: Ensures `@reverse.instanceOf.encodingLevel` is indexed as keyword in ES. This issue was discovered through a failing test: https://github.com/libris/lxl_api_tests/blob/1f20a14dd2d780ea19c1c7e19d64e1388ec03d6a/tests/test_restapi_2.py#L56. All tests now pass locally.

See new related API tests: https://github.com/libris/lxl_api_tests/pull/49

### DIFF
--- a/librisxl-tools/elasticsearch/libris_config.json
+++ b/librisxl-tools/elasticsearch/libris_config.json
@@ -185,10 +185,6 @@
             },
             "meta": {
                 "properties": {
-                    "encodingLevel": {
-                      "type": "keyword",
-                      "copy_to": "_all"
-                    },
                     "recordStatus": {
                       "type": "keyword",
                       "copy_to": "_all"
@@ -350,6 +346,15 @@
             {
                 "issuanceType_template": {
                     "match": ["issuanceType"],
+                    "mapping": {
+                        "type": "keyword",
+                        "copy_to": "_all"
+                    }
+                }
+            },
+            {
+                "encodingLevel_template": {
+                    "match": ["encodingLevel"],
                     "mapping": {
                         "type": "keyword",
                         "copy_to": "_all"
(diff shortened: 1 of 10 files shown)


olovy:
> with the only difference that meta for virtual works will always have this minimal shape

doesn't this break sorting on meta.created and meta.modified?

olovy (commented):
See comment about sorting on meta.created / meta.modified

Other than that, LGTM

olovy (approved):

PR #1772, opened 2026-05-21
Bugfix/linked shelfmark 2

kwahlin:
Still a temporary work-around, however a more harmless one.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
+++ b/whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
@@ -839,7 +839,7 @@ class ElasticSearch {
         return JsonLd.getExternalReferences([(GRAPH_KEY): graph])
                 .findAll {
                     // FIXME
-                    if (jsonLd.isIntegral(it.property())) {
+                    if (jsonLd.isIntegral(it.relation)) {
                         return true
                     }
                     def path = it.propertyPath()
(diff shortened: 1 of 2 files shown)


whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -583,9 +584,29 @@ class ElasticSearch {
 
         var integralIds = collectIntegralIds(shapedMainGraph, whelk.jsonld)
 
+        // FIXME
+        def embedFakeIntegralThings = { graph ->
+            def thing = DocumentUtil.getAtPath(graph, Document.thingPath, [:])
+            if (thing[TYPE_KEY] == 'Item') {
+                DocumentUtil.findKey(thing, Embellisher.FAKE_INTEGRAL_RELATIONS) { value, path ->
```
olovy, line 591: findKey won't find `hasComponent.shelfMark`, `hasComponent.availability` ?

kwahlin: It will find all paths ending with `shelfMark` or `availability` when traversing the item. Maybe clearer with an explicit `DocumentUtil.findKey(thing, ['shelfMark', 'availability'])` but I wanted to keep the connection to embellish until when we fix this properly.

whelk-core/src/main/groovy/whelk/component/ElasticSearch.groovy
```diff
@@ -583,9 +584,29 @@ class ElasticSearch {
...
+                    asList(value).each { v ->
+                        if (v instanceof Map && JsonLd.isLink(v)) {
+                            embellishedGraph.find { DocumentUtil.getAtPath(it, Document.thingIdPath2) == v[ID_KEY] }
+                                    .with {
```
olovy:
```suggestion
                                    ?.with {
```

Or will get NPE if not found in graph

olovy (approved):
LGTM.

See suggestion.

PR #1774, opened 2026-05-26
feat(marc-export): auth export on chip changed, not card

olovy:
Determined to be sufficient condition for concepts and agents by KBV team and RDA editorial team.

Works use bib rules so are not affected.


kwahlin (approved):
LGTM

PR #1777, opened 2026-05-26
Limit the number of MARC instance exports a single edit in libris can…

jannistsiroyannis:
… cause (to 400 for now).


kaipoykio (approved):
*thumbs up*

PR #1779, opened 2026-05-29
Feature/fmt 548 replace other physical format with reproduction of

branchedelac:
Jupyter notebook and Groovy script to replace otherPhysicalFormat with reproductionOf, in cases where a digital and a physical instance mutually point to each other via otherPhysicalFormat > describedBy > controlNumber.

The notebook also identifies more difficult cases, for example where the otherPhysicalFormat relationship is one-sided. These need further analyses will be dealt with at another point, and are out of scope for this PR.


ebengtsson (approved):
LGTM

PR #1780, opened 2026-05-29
Feature/search2 various

kwahlin:
Complements https://github.com/libris/definitions/pull/600

- Remove the special handling of  `meta` in query expansion since it mostly just caused bugs. With shorthand properties now defined for the common use cases there is no need for it. When there is no shorthand available, queries can simply use the full qualified form, e.g. `meta.x:y`. 
- Support for identifiable restricted sub-properties, either as ID'd blank properties or named properties, e.g.:
```
# named property :x
:x :rdfs:subPropertyOf :y ;
    rdfs::range [ a owl:restriction ; owl:onProperty :p ; owl:hasValue :a ] . 
```
```
# blank property _:x
_:x :rdfs:subPropertyOf :y ;
    rdfs::range [ a owl:restriction ; owl:onProperty :p ; owl:hasValue :a ] . 
```
- Minor fixes/cleanups for things discovered when testing
- [02df0d9](https://github.com/libris/librisxl/pull/1780/commits/02df0d93ec6da210f4a0f537062d0fbd9920ca25) limits the set of fields searched when using e.g. the `title:` filter. While it removes a few superfluous fields from the expanded query, the handling of integral properties is still a bit too eager, causing `title:` to expand to more fields than necessary. This is not a big issue (better searching too many fields than too few) but we should fix it — either in the implementation or by tweaking the definitions.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/search2/querytree/Property.java
+++ b/whelk-core/src/main/groovy/whelk/search2/querytree/Property.java
@@ -115,6 +115,9 @@ public non-sealed class Property extends PathElement {
         if (isCoercing(definition)) {
             return new CoercingSubProperty(definition, jsonLd, propertyKey, queryKey);
         }
+        if (jsonLd.isIntegral(propertyKey)) {
+            return new IntegralProperty(definition, jsonLd, propertyKey, queryKey);
+        }
         if (RDF_TYPE.equals(propertyKey)) {
             return new RdfType(jsonLd, queryKey);
         }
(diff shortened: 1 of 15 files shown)


whelk-core/src/main/groovy/whelk/search2/querytree/Property.java
```diff
@@ -58,63 +54,102 @@ public non-sealed class Property extends PathElement {
...
         this.indexKey = (String) definition.get("ls:indexKey"); // FIXME: This shouldn't have a different prefix (ls: vs librissearch:)
-        this.objectRestrictions = getObjectRestrictions(jsonLd);
-    }
-
-    private Property(String name, JsonLd jsonLd, Key.RecognizedKey queryKey) {
-        this(name, jsonLd);
         this.queryKey = queryKey;
+        if (name != null) {
+            this.name = name;
+            this.langAlias =(String) jsonLd.langContainerAlias.get(name);
```
olovy:
```suggestion
            this.langAlias = (String) jsonLd.langContainerAlias.get(name);
```

olovy (approved):
LGTM!

whelk-core/src/main/groovy/whelk/search2/querytree/Property.java
kwahlin: :monocle_face:

PR #1784, opened 2026-06-01
feat(xsearch): support q= as well as query=; JSONP callback support (ugh)

andersju:
`q` is used instead of `query` by some.

`callback=` still has one prominent user. Let's support it until they change things.


olovy (approved):
LGTM!

PR #1786, opened 2026-06-02
fix(xsearch): handle the case where agent is a list

andersju:
It can happen that contribution.agent is a list. https://libris-qa.kb.se/api/xsearch?query=onr:6nwssf4h4spbr3lw&format=json


kwahlin (approved):
LGTM

PR #1788, opened 2026-06-05
Sru with transformers

kaipoykio:
The sru transformer is a bit clumsy and requires maybe a rework.


jannistsiroyannis (commented):
The indentation got a little weird, is it a tabs/spaces thing?
There's a few commented out code blocks that could be removed, and the run-local script should perhaps not be checked in, it assumes a certain directory which wont be the case for everyone.
Otherwise, LGTM (after handling the merge conflicts)!

kaipoykio:
Thx, I'll fix ;)

sru/src/main/java/whelk/sru/servlet/SruServlet.java
```diff
@@ -36,105 +35,77 @@
...
-            maximumRecords = Integer.parseInt(getParameter(parameters, "maximumRecords"));
-        }
-
         Map<String, Object> results;
+        String format;  
+
         try {
-            String CqlQueryString = getParameter(parameters, "query");
+            String CqlQueryString = parameters.get("query")[0];
+            format = parameters.get("recordSchema")[0];
```
andersju: `recordSchema` is afaik not a required parameter so this would NPE

andersju (commented):
See comment about recordSchema. There have been a lot of changes to sru since this was branched off so it needs some careful rebasing to fix conflicts/not break some newer stuff: https://github.com/libris/librisxl/commits/develop/sru/src/main/java/whelk/sru/servlet

Otherwise LGTM!

sru/src/main/java/whelk/sru/servlet/SruServlet.java
kaipoykio: Its gotten from a map, returns null if not present.

PR #1789, opened 2026-06-05
Add simple SparqlReindex script in Java

niklasl:
We _could_ integrate this properly; but is may be a "one-off" (see LXL-4798).

Run like:
```bash
java -cp whelktool/build/libs/whelktool.jar -Dxl.secret.properties=SOME-secret.properties librisxl-tools/scripts/SparqlReindex.java FILE_1_WITH_IDS ... FILE_N_WITH_IDS
```


olovy (approved):
LGTM!

Would like a comment or `--help` documenting the format of the file

PR #1791, opened 2026-06-09
Define freeOnline filter alias for XSearch/SRU

kwahlin:
We previously added a hard-coded mapping `MAT:free` -> `freeOnline` for backward compatibility, however it turned out that the `freeOnline` filter wasn't configured in XSearch/SRU. ~This is just a quick-fix; the filter definition should not be hard-coded. Can we put this configuration in [apps.jsonld](https://github.com/libris/definitions/blob/develop/source/apps.jsonld)?~
See configuration in https://github.com/libris/definitions/pull/604

### DIFF
--- a/sru/src/main/java/whelk/sru/cql/Translation.java
+++ b/sru/src/main/java/whelk/sru/cql/Translation.java
@@ -45,12 +45,6 @@ public class Translation
         Object phase1Ast = Phase1.reduce(cst);
         String xlql = Phase2.flatten(phase1Ast);
 
-        // CQL searches only for instances
-        if (xlql.startsWith("(") && xlql.endsWith(")"))
-            xlql = xlql + " AND type=Instance";
-        else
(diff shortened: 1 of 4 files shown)


olovy (commented):
Why not just fix the initialization of AppParams?

kwahlin:
> Why not just fix the initialization of AppParams?

Initialize with what configuration? Same as for regular search (with e-plikt excluded etc.)?

olovy:
> > Why not just fix the initialization of AppParams?
> 
> Initialize with what configuration? Same as for regular search (with e-plikt excluded etc.)?

define https://libris.kb.se/sru/libris in apps.jsonld?

olovy (approved):
Nice! LGTM

See comment about type:instance https://github.com/libris/definitions/pull/604
In that case -> corresponding changes here.

PR #1792, opened 2026-06-10
All CQL 'strings' must be quoted in the resulting XLQL query.

jannistsiroyannis:
This is to avoid collisions with XLQL special characters. This replaces a prior temporary solution for ':' alone.

### DIFF
--- a/sru/src/main/java/whelk/sru/cql/Phase2.java
+++ b/sru/src/main/java/whelk/sru/cql/Phase2.java
@@ -14,21 +14,21 @@ public class Phase2 {
                 return flatten(bg);
             }
             case String s -> {
-
-                String str = s;
-                if (str.startsWith("\"") && str.endsWith("\"")) // Quoted strings come out here quoted, which causes issues.
-                    str = str.substring(1, str.length()-1);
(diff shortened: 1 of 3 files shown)


kwahlin:
This means no soft matching and that the word order matters, is that OK for SRU?

olovy:
Would it be enough to just strip all XLQL special characters from the strings?

If we go down the quoting route: We interpret all queries as CQL "all". Meaning that term order and adjacency should not matter. So each term/word should be quoted separately?

Another problem with quoting the strings is that diacritics have to match exactly (no "ascii folding").
No stemming is probably ok. But results will differ more from other API:s

jannistsiroyannis:
Your right, we should strip them instead. Ill change it!

jannistsiroyannis:
I've changed it to scrub away the offending characters. But I also changed it in another way, so that strings are passed through inside parentheses.

Because consider the following query:
z3950.1003 = "Lidbeck, Lasse"
If it gets translated (without quotes) it semantically becomes
z3950.1003 = Lidbeck, AND Lasse
which is incorrect.

Now we instead
translate it to
z3950.1003 = (Lidbeck, Lasse)
which is correct and also allows for words-out-of-order etc.

olovy (approved):
LGTM! 🔡

PR #1795, opened 2026-06-16
Feature/composite property fields

kwahlin:
Currently, the `title:` search filter expands to 20 different fields, many of them irrelevant. Additionally only one of the fields (`hasTitle`) is indexed as a nested field.
bild

This PR narrows the search scope to the 9 fields that are actually relevant to title queries and ensures that all of them are indexed as nested fields. (The screenshot below still shows one non-nested query. This is only because the field is missing entirely from the local test dataset.)
bild

The same principle of narrowing the search scope to relevant fields also applies to other filters such as `contributor:`.

See also: https://github.com/libris/lxl_api_tests/pull/53


kwahlin:
Included in #1830 instead

PR #1797, opened 2026-06-17
Handle revert on non-string values

niklasl:
Error cropped up on revert of e.g. https://libris.kb.se/rd4gslvdp3283glj which for some reason contains:
```json5
    {
     // ...
      "publication": [
        {
          // ...
          "year": 1961
        }
      ]
    }
```


olovy (approved):
LGTM

Can use String.valueOf() instead if nulls are possible

niklasl:
Good call; I added a not null condition instead (consider null value as failing to match any pattern).

PR #1801, opened 2026-06-22
feat(oaipmh): add support for resumptionToken

andersju:
Most OAI-PMH server implementations limit the number of records per HTTP response to something like 100 or 1000 and use [resumptionToken](https://www.openarchives.org/OAI/openarchivesprotocol.html#FlowControl) for pagination. We haven't had support for this, which has caused problems for some users doing heavy requests that never finish properly. So let's add it. This implementation is stateless and encodes all the necessary stuff in the resumptionToken sent to the client.

I've done some local testing but will also add some tests to https://github.com/libris/lxl_api_tests.

Not sure if `x-withSilentUpdates` is still necessary (it's not publicly documented) but I kept it around for now. `x-withSilentUpdates` is probably a little slower now, but we can add an index if it's an issue.

https://kbse.atlassian.net/browse/LXL-4808

### DIFF
--- a/oaipmh/src/main/java/whelk/export/servlet/GetRecord.java
+++ b/oaipmh/src/main/java/whelk/export/servlet/GetRecord.java
@@ -78,7 +78,7 @@ public class GetRecord
         try (Connection dbconn = OaiPmh.s_whelk.getStorage().getOuterConnection())
         {
             dbconn.setAutoCommit(false);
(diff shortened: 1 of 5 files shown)


jannistsiroyannis (approved):

jannistsiroyannis:
I'm a little nervous about "order by id". Probably fine.

olovy (approved):
LGTM!

PR #1802, opened 2026-06-22
fix(find.atom): updated timstamp

olovy:
updated timestamp is mandatory on feed.

Set updated to
- last modified item in result
- now() if no results


andersju (approved):
LGTM!

PR #1807, opened 2026-06-30
Handle cyclic categories in BibTypeNormalizationStep

niklasl:
This should work, and avoids introducing a bug that would fail to get all candidates and therefore not honor `prioritizedWorkLegacyTypes`.

Also includes a basic unit test.


andersju (approved):
Did a little testing; LGTM!

PR #1808, opened 2026-07-03
Static-compile all major runtime MarcFrameConverter logic

niklasl:
This removes most `CompileStatic`-skips, and adds `CompileStatic` to several/most pre-/post-processing steps; intending to make the code faster, stricter, and reasonably more robust.

Some very simple testing (using the (fixed) `perf` feature of `MarcFrameCli`) indicates that this speeds up revert-to-MARC by a factor of 2x to 3x (200k repeated revert-runs that previously was 11 ms on average is now down to 3.7 ms).

All marcframe integration tests pass, so behaviour _should_ be identical. However, there may be some intricate bug(s) left, since Groovy calls overloaded methods using the _actual_ type of the objects (dynamic dispatch), whereas Java selects method by (compile-time) variable type (static dispatch).

The tests smoked out a bunch of those, _possibly_ all, but we need to either analyze all call sites with intent, or run some rather large export tests to be sure enough.

### DIFF
--- a/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
+++ b/whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
@@ -470,8 +470,10 @@ class MarcConversion {
             if (tag == "000") {
                 marc.leader = value
             } else {
-                if (value == null)
+                if (value == null) {
                     return
+                }
+
                 if (value instanceof List) {
                     value.each {
                         if (it) {
@@ -1195,7 +1197,7 @@ class ConversionPart {
 
         // The about map is the whole embellished record N times, framed around the "pending keys", typically:
         // ?record, ?thing, ?work, _:provision
-        return new Tuple2<Boolean, Map<String, List>>(requiredOk, aboutMap)
+        return new Tuple2(requiredOk, aboutMap)
     }
 
     static String findTokenFromId(Object node, String uriTemplate,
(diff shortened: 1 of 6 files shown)


whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
```diff
@@ -2289,20 +2297,20 @@ class MarcFieldHandler extends BaseMarcFieldHandler {
...
-            newEnt = newEntity(state, linkRule.resourceType, linkRule.groupId, linkRule.embedded)
+            newEnt = newEntity(state, (String) linkRule.resourceType, (String) linkRule.groupId, (boolean) linkRule.embedded)
 
             // TODO: use @id (existing or added bnode-id) instead of duplicating newEnt
-            def entRef = newEnt
+            var entRef = newEnt
             if (useLinks && linkRule.link) {
                 if (!newEnt['@id'])
-                    newEnt['@id'] = "_:b-${state.bnodeCounter++}" as String
+                    newEnt['@id'] = "_:b-${((int) state.bnodeCounter)++}" as String
```
andersju: Small regression here. Previously `state.bnodeCounter` was post-incremented, but now due to the casting it is the (temporary) result of the local expression that's post-incremented instead. So maybe something like this:
```groovy
if (!newEnt['@id']) {
    int c = (int) state.bnodeCounter
    newEnt['@id'] = "_:b-${c}" as String
    state.bnodeCounter = c + 1
}
```

andersju (changes requested):
Great! Quick testing of some exports with the marc_export API shows a nice speed boost. Found one small issue (see comment).

whelk-core/src/main/groovy/whelk/converter/marc/MarcFrameConverter.groovy
niklasl: Good catch; will address!

niklasl: Did that fix (verbatrim) in 1ee3dc25c64cebec5abeb6c196ba3c67a08df326

andersju (approved):

PR #1810, opened 2026-07-10
Extend whelk to RDF dump feature with n-quads support

niklasl:
Example usage:
```sh
cd importers
../gradlew jar
mkdir -p /tmp/libris-dumps
java -Dxl.secret.properties=../secret.properties.dev -jar build/libs/xlimporter.jar \
        lddbToNquads /tmp/libris-dumps/dev-bib-nq 4 --gzip bib
```

This is still WIP for various reasons:
- [ ] Compliation of importers failed on develop (possibly due to gradle bump). This is adressed in 5502f9e72f69d77193281a1f2beb6c047720ffc8 which we may want to cherry-pick if that's the way to go.
- [x] We _must_ to use a unique bnodes minter (e.g. doc-and-batch-count as prefix, or uuid:s) per conversion (a.k.a. "standardize apart"); the current nquads writer here will result in horrendous bnode smushing!
- [ ] I prefer to isolate trld-java API dependency a bit more (any may improve its public API further to do so).
- [x] TRLD 0.3.0 includes proper n-quads support (and c14n, though we don't need it here); so we should either just bump trld-java, or backport that part (crucially fixes newlines in strings).
- [ ] We do some intricate cleanup in `JsonLdToTrigSerializer` which we may want to go about differently to reuse here (ideally of course fix the data in XL to prevent e.g. broken IRIs from ever entering).

### DIFF
--- a/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
+++ b/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
@@ -2,7 +2,6 @@ package whelk.importer
 
 import java.lang.annotation.*
 import java.util.concurrent.ExecutorService
-import java.util.zip.GZIPOutputStream
 import groovy.cli.picocli.CliBuilder
 import groovy.util.logging.Slf4j as Log
 import org.apache.commons.io.output.CountingOutputStream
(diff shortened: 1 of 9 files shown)


niklasl:
(Rebased on develop to account for switch to slf4j.)

PR #1813, opened 2026-07-14
feat(rest): Port SearchFeed to Java

andersju:
Last year we did a push to make the rest component purely Java (except the tests - and DigitalReproductionAPI (might be next...)), so let's try to keep it that way (knowing SearchFeed was initially created long before the Java push :sweat_smile: ).

This is a fairly mechanical port that gives identical results to the Groovy version in various tests I did.

(I renamed the file in a separate commit so as to preserve history/git blame, but GitHub doesn't handle it in the diff here.)


kaipoykio (approved):
Looks reasonable ;)

PR #1814, opened 2026-07-15
refactor(rest): port DigitalReproductionAPI to Java

andersju:
Clanker-assisted port, but I've gone over every single line and compared with the original.

Split into multiple files since there were several separate classes in one single Groovy file.

Also added some (generated) tests.

Also changed `build.gradle` since after this change there's no more Groovy code in `rest/` (apart from the tests), so we don't have to pass the code through the Groovy compiler anymore. Instead it goes to directly to javac which speeds up compilation a bit.


kaipoykio (approved):
Nice!

PR #1815, opened 2026-07-17
refactor(whelk-core.util): port util stuff from Groovy to Java

andersju:
Another fairly mechanical port because I had tokens to burn, this time of the remaining groovy stuff in whelk-core's util. Before porting I generated more tests.

One thing of note is that a bug in DocumentUtil.linkBlankNode was unearthed. The original had

```groovy
if (isBlank(node) && (newLinked = linker.link(node, existingLinks) ?: linker.link(node, collectIris(disambiguationNodes)))) {
                result.addAll(newLinked.findAll { l ->
                    !existingLinks.contains(l['@id']) && !result.contains { it['@id'] == l['@id'] }
                })
            } else {
                result.add(node)
            }
```
where `!result.contains { it['@id'] == l['@id'] }` didn't actually do anything (always true). I assume `!result.any {...}` was intended. This should probably be fixed but I didn't want to change behavior in a port, so I just added a TODO comment.

Also removed the Tools class since nothing in it had been used for 10+ years.

As always, GitHub doesn't show the proper diffs on this overview page, but I renamed the files in separate commits and if you check e.g. the "port DocumentUtil to Java" you'll see the real diff.

Net line growth compared to the original Groovy code, not counting the tests or the removed Tools.java, is only a few hundred lines.


kaipoykio (approved):
On a roll! Super!

PR #1817, opened 2026-07-21
feat(importers): add copyRecords tool for importing individual records

andersju:
Something I've wanted for a _long_ time but couldn't be bothered with until now.

To import sample records to a dev db we've only had the importer tool's `copywhelk` command, which takes a file with a list of IDs (and optionally a list of types to copy). This is mainly intended for the initial run when (re)creating a dev db. `copywhelk` redenormalizes the entire db and doesn't reindex on its own.

The use case this PR solves is: "I have an existing dev database and I just want to quickly import one or a few specific records and their dependencies/dependers from the prod db (and redenormalize+index only what's necessary)". Et voilà:

```bash
cd importers/
../gradlew jar
```
To import a couple of records by short ID, including their version history and holdings (if any):
`java -Dxl.secret.properties=../secret.properties -jar build/libs/xlimporter.jar copyRecords --copy-versions --include-items ../temp_source_secret.properties 4ngh2qhg49jhnp4 vhsxx8rks82lgtn5`

By IRI:
`java -Dxl.secret.properties=../secret.properties -jar build/libs/xlimporter.jar copyRecords --copy-versions --include-items ../temp_source_secret.properties https://libris.kb.se/4ngh2qhg49jhnp4`

Where `temp_source_secret.properties` contains the details for the source database (e.g. qa/stg/prod readonly...). Just like with `copywhelk`.

`--overwrite-existing` can also be specified. In that case the specified record will be _updated_ with the contents of the source version (but then version history is not copied).

### DIFF
--- a/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
+++ b/importers/src/main/groovy/whelk/importer/ImporterMain.groovy
@@ -172,6 +172,40 @@ class ImporterMain {
         copier.run()
     }
 
+    /**
+     * Copy one or a few records, specified by XL id (or IRI) on the command line, from
+     * another whelk, and index them. Like copywhelk, records that the specified records
+     * link to, and records that link to them, are copied as well.
+     *
+     * Unlike copywhelk's --copy-versions, this one is boolean: if set, all
+     * historical versions of the *specified records* (not of their deps) are copied.
+     *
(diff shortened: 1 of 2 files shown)


kaipoykio (commented):
Looks fine!

PR #1820, opened 2026-07-24
refactor(housekeeping): Port HouseKeeping from Groovy to Java

andersju:
One more clanker-assisted port that should be faithful to the Groovy version.

Now that everything is Java (except tests), we compile with javac instead of going through the Groovy compiler, reducing compilation time a bit.

Also added a bunch of tests generated against the original Groovy code.


jannistsiroyannis (approved):
🤯 
Not gonna pretend i read this line by line, but [if it works], great!

PR #1823, opened 2026-08-07
chore: migrate logging to slf4j + logback

olovy:
log4j 1.x is EOL since 2015...
There exists a [binary compatible fork](https://reload4j.qos.ch/) of log4j 1.2 with security fixes.
Let's switch to the more or less standard slf4j instead. And use logback as logging backend.

Verified log output on dev2.

There is some additional noise when running unit tests. But I think we can deal with that separately.


andersju (approved):
Nice! LGTM!

PR #1824, opened 2026-08-10
chore: upgrade prometheus client to 1.8.0

olovy:
Upgrade prometheus client from 0.15.0 to 1.8.0. 

* `_total` is now a reserved prefix for counter names and cannot be used by `Gauge`. The following metric names have been changed
    * `api_ongoing_requests_total` --> `api_ongoing_requests`
    * `oaipmh_ongoing_requests_total` --> `oaipmh_ongoing_requests`
    * `marc_export_api_ongoing_requests_total` --> `marc_export_api_ongoing_requests`
* `startTimer()` now returns an auto-closable Timer. No more `observeDuration()`
* resilience4j, HikariCP and guava cache metrics are stuck on on prometheus simpleclient 0.x. They are bridged to 1.x using prometheus-metrics-simpleclient-bridge
    * https://prometheus.github.io/client_java/migration/simpleclient/


whelk-core/src/main/groovy/whelk/component/PostgreSQLComponent.groovy
```diff
@@ -2963,10 +2964,11 @@ class PostgreSQLComponent {
 
     class NotificationListener extends Thread {
         private static final String NAME = 'pg_listener'
-        private static final Counter counter = Counter.build()
+        private static final Counter counter = (Counter) Counter.builder()
```
olovy, line 2967: This cast shouldn't be necessary but somehow the Groovy compiler fails here otherwise?

Same for  CounterDataPoint below

jannistsiroyannis (approved):

PR #1833, opened 2026-09-04
fix(search2/lex): treat non-breaking space as whitespace

olovy:
lxl-web / codemirror sometimes inserts U+00A0 in queries.
Breaks for example: " +\U+00A0NOT" becoming the string " NOT" instead of the keyword NOT.
It makes sense treating nbsp as whitespace anyway.


kaipoykio (approved):
Nice!

PR #1836, opened 2026-09-15
chore(whelk-core): CompileStatic more things

andersju:
For the ongoing quest to convert XL to Java, it's probably best to start with adding `@CompileStatic` to existing Groovy files so we can more easily find/fix type errors (and then the eventual conversion to Java will also be easier). So here's a quick start, just straightforward cases. More to come in other PRs.


kaipoykio (approved):
LG

olovy (approved):
LGTM

